PR #5, opened 2018-05-19
Fix a couple typos on FAQ page

chellbell:
Came across this while browsing the v2 site and figured I'd try to leave it a little better than I found it :)


iHiD:
Thanks, @chellbell :) 

So, I never knew you could spell "fulfil" as "fulfill" :) Turns out the former is English/Australian and the later is American.

@kytrinyx Did you/we ever discuss or decide whether to use English or American spellings? I feel like we should standardise on one or the other. My preference would be whichever is more "international" outside of our respective countries. I'm guessing in development worlds that's probably American (thinking HTML spellings) so happy to standardise on that. Would appreciate your opinion.

kytrinyx:
I don't have strong preferences or opinions in terms of spelling, only that we pick one and stick with it. I think standardizing on American would be perfectly reasonable (HTML spellings is as good a reason as any). In which case, `fulfill` it is, according to grammarist: http://grammarist.com/spelling/fulfil-fulfill/

kytrinyx:
Thanks @chellbell!

PR #8, opened 2018-06-17
Add mentor bio for kytrinyx

kytrinyx:


iHiD:
Remind me: Do we default to Github avatar and link if it's `null`?

kytrinyx:
We do for maintainers.

PR #74, opened 2018-06-20
Add mentor bio (rust) for mairandomness

mairandomness:


mentors/rust/m.json
```diff
@@ -1 +1,11 @@
-[]
+[
+  {
+    "github_username": "mairandomness",
+    "name": "Maira Kodama",
+    "link_text": null,
+    "link_url": null,
+    "avatar_url": null,
+    "bio": "Math teacher turned programmer. Fell in love with Rust while listening to Steve Klabnik talk about its design choices. Eternally grateful for helpful compiler messages. Not tall enough to write multi-threaded code.
+"
```
rpottsoh: @mairandomness I think you just need to move the `"` from the start of line 9 to then end of line 8.  Then this JSON will be valid.

PR #82, opened 2018-06-20
added bio of rahmonov for python track

rahmonov:


amaliacardenas:
Added rahmonov to the mentor bios in the python track. All tests passed, valid JSON was verified, and bio checked.

PR #86, opened 2018-06-21
n8chz, mentor for C track

n8chz:


iHiD:
@n8chz I think an inline image URL will probably not fit inside the database column. Any chance you could either point to a URL or if its not online, comment with the inline image and I'll host it somewhere for you?

mentors/c/n.json
```diff
@@ -1 +1,8 @@
...
+  "bio","Just an enthusiast programmer"
```
rpottsoh: `,` should be a `:`.

PR #91, opened 2018-06-22
Add profile for Thales/Vitruvia

Vitruvia:
I have made this branch to add my profile as a Python mentor.

### DIFF
--- a/mentors/python/t.json
+++ b/mentors/python/t.json
@@ -1,1 +1,8 @@
-[]
+[{
+  "github_username": "Vitruvia",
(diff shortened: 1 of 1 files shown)


amaliacardenas:
Hi @Vitruvia,

Thanks for your contribution. 
Unfortunately we cannot merge your bio because the track/filename is not in the right format. 

It needs to be (track/#{github_username_initial}.json. Currently you have your first name as the initial, but we need the initial of your github account. Please fix and resubmit.

PR #103, opened 2018-07-01
Add alebaffa to Java and Go tracks

alebaffa:


mentors/go/a.json
```diff
@@ -1 +1,8 @@
-[]
+{
```
junedev: Looks like you lost the square brackets around your objects. See https://github.com/exercism/website-copy/pull/8/commits/469c4e2242d320928162d12bc83efea799a1c2fa for reference.

alebaffa: My apologies. I fixed it. Thanks!

PR #120, opened 2018-07-09
Add Terms and Privacy Policy for the updated Exercism site

ccare:


iHiD:
@ccare @kytrinyx Can you agree when you want to discuss this, and then ping me on Slack when with a time and we can go through this together.

PR #128, opened 2018-07-12
Add nathanchere

nathanchere:


amaliacardenas:
Hi @nathanchere Thanks for submitting your pull request. Before we can merge your bio to the elixir track, the JSON file needs to be in the correct format.

I see that the line ```"link_text": "NathanChere.com.au,``` needs a closing comma at the end.
It should read like this: ```"link_text": "NathanChere.com.au",``` Can you fix and resubmit?

PR #178, opened 2018-07-14
Update report_abuse.md

guygastineau:
This sets the policy on the preferred way of referring users to the Code of Conduct.

I tried to put the sentence starting with "As such," on a new line but when I view it it is one the same line as the previous sentence.  Help formatting that would be appreciated.

Hopefully I worded things in a suitable manner.  I just want people to know of the accepted method for referring others to the CoC so it resolves issues instead of making them.


iHiD:
This is great. Thank you.

PR #248, opened 2018-07-24
Fix incorrect commands in CLI page

kytrinyx:
The page was referencing v2 commands.

FYI @iHiD


iHiD:
Great. Thanks. It'll be so good getting all this stuff in this one repo. Makes find/replace so much saner :)

PR #254, opened 2018-07-24
Split out submit problems in FAQ

kytrinyx:
People are missing the bit about how to submit.
This might help reduce the number of support requests we get about it.


kytrinyx:
Maybe instead of this we could say:

> please ensure that you have the most recent version of the CLI.
> If that doesn't help, then open an issue on exercism/exercism.io and we will help you get it sorted out.

iHiD:
I added some text based on what you said, and merged it to get **something** there for now.

PR #259, opened 2018-07-25
updated two links and fixed a small typo

grantwinney:
Fixed two links pointing to the deprecated read-only repo. Also fixed a small typo - "if" vs "it".


iHiD:
Brilliant. Thanks! :)

PR #263, opened 2018-07-26
Update code_of_conduct.md

iHiD:
I wanted to do two things here:
1) Give people confidence to report things to us
2) Emphasise that we take reports seriously, and will ban people if we want to.

As such, I've:
- Added a little bit of structure
- Added sections that make it clear what to do and that we take things seriously.
- Moved the history to the bottom because while it's important to recognise our sources, it isn't the most important thing.

### DIFF
--- a/pages/code_of_conduct.md
+++ b/pages/code_of_conduct.md
@@ -1,15 +1,27 @@
-This policy was adopted from the Front-end London Slack community.
+### Introduction
 
-_This policy is a "living" document, and subject to refinement and expansion in the future. This policy applies to every Exercism entity or event, including but not limited to the Slack community, Twitter chats and GitHub posts._
+Exercism is a platform centered around empathetic conversation. We have a low tolerance for communication that makes anyone feel unwelcome, unsupported, insulted or disciminated against.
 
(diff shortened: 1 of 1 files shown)


nicolechalmers:
Those changes look good. 

I have a couple of suggestions:
- do you think we should mention the age restriction? 
- how about adapting the heading "What to do if you see something wrong" to add more than **seeing** something wrong. Something like "What to do if you see something wrong or **feel uncomfortable** " or "What to do if you see something wrong or **someone has made you feel uncomfortable**"

iHiD:
@nicolechalmers I played around with a couple of headings but they felt too long. I agree with "What to do if you see something wrong or someone has made you feel uncomfortable" as the best one, but I wonder if there's a shorter one we could use instead?

nicolechalmers:
@iHiD Some more options:
> Seen or experienced something you find uncomfortable?
> Seen or experienced something uncomfortable?
> Something on Exercism made you feel uncomfortable?
> Want to check if something you've seen/experienced is against the CoC? (appreciate this isn't exactly shorter)

PR #267, opened 2018-07-28
Add my bio

janczer:


amaliacardenas:
Hi @janczer,
Thanks for submitting your bio. Before we can merge your bio, the filename name needs to be in the right format. Can you rename your file, and resubmit? It needs to be in the following format (track/#{github_username_initial}.json, For example, in your case, it would need to be renamed to mentors/php/j.json).

janczer:
Hi @amaliacardenas,

Thank you for your comment. I change it.

Issue #265, opened 2018-07-27
Rake test denies empty link_url in mentors bio, which is allowed by README.


NobbZ:
There is this test in the `rake test` task:

https://github.com/exercism/website-copy/blob/master/Rakefile#L23-L26

It does check if the `link_url` of a mentor is the empty string.

But the README states, that one should put an empty string for "no entry". Instead the test wants to see `null` here. Which according to the README means to use whatever GitHub has as `link_url`.

kephas:
Should be fixed in #269 if the rakefile is supposed to be correct.

kephas:
I would actually advocate for the rakefile to be changed to be synced with the documentation. I like the idea to be able to opt-out.

NobbZ:
@kytrinyx @iHiD or anyone else who is in the product team (why dont we have @exercism/product?), can we expect a decission on this?

iHiD:
It's assigned to @kytrinyx who's been off for a few days. I'm sure it's up there on her priority list now she's back, but there are lots of things around the CLI that are also high on that list :)

NobbZ:
Ah, sorry haven't seen that it is already assigned, as I'm interacting with tickets through email and mobile interface most of the time.

iHiD:
That's ok. I appreciate the nudge :)

kytrinyx:
Sorry about this--this is my bad. I forgot that `null` means "use what's on GitHub" and `""` means "Don't do anything".

The correct fix here is to allow empty strings (but not allow whitespace).

kytrinyx:
We'll also need to go back and fix people's empty links, which I changed in

06e455f (Fix invalid mentor bios)

PR #269, opened 2018-07-28
Sync documentation with rakefile

kephas:


NobbZ:
I'm not sure if this is the correct way to solve the issue, please see also https://github.com/exercism/website-copy/issues/265#issuecomment-408598059

kytrinyx:
Thanks for this. Having read through the original issue, this is a mistake on my part. I forgot that we were allowing both versions.

The fix needs to be made to the Rake task.

PR #270, opened 2018-07-28
Add stevejb71 to OCaml mentoring

stevejb71:


sshine:
@stevejb71: Perhaps, in case you have become inactive as a mentor, it is better to leave the impression that there are no active mentors on [this page](https://exercism.io/tracks/ocaml/mentors)? (I am no longer active as a mentor, but I continue as a passive maintainer.)

rpottsoh:
If you are no longer able mentor exercises then it is probably best to remove yourself.  I understand your point about the track then appearing to have no mentors....
cc: @iHiD

iHiD:
I guess if the track has no mentors it has no mentors. And it will be in oversubscribed mode.

stevejb71:
I'm mentoring very infrequently now, so agree it's reasonable I should be removed.

rpottsoh:
Open a PR that removes you from the json file.

PR #276, opened 2018-07-30
add Jacqueline Outka bio to Scala track

outkaj:


amaliacardenas:
Hi @outkaj 
Thanks for submitting your bio. Before we can merge your bio, the filename name needs to be in the right format. Can you rename your file, and resubmit? It needs to be in the following format (track/#{github_username_initial}.json, For example, in your case, it would need to be renamed to mentors/scala/o.json).

outkaj:
@amaliacardenas my apologies - should have RT(F)M better :) thanks for the correction. this should be all set now.

PR #291, opened 2018-08-02
Updated "mentor" link

grantwinney:
The other mentor page is still there, but it has no content. I updated the link to point to the new "Become an Exercism mentor" page at http://mentoring.exercism.io/


iHiD:
Thank you!

PR #296, opened 2018-08-03
Add Skade as a Rust mentor

skade:
Hi everyone!


iHiD:
@skade You have your link text and url the wrong way around :)

skade:
Ugh, right :D. Here we go again.

iHiD:
And we're green :)

PR #299, opened 2018-08-03
[Write/Speak/Code 2018] Mentor url fix

hiimbex:
Fixes https://github.com/exercism/website-copy/issues/265 and reverts https://github.com/exercism/website-copy/commit/06e455f4ef40a11596a451b944d531f92d5d8d91.

I wasn't sure what to do for 2 mentors who are currently failing the rakefile test in: [c/k.json](https://github.com/exercism/website-copy/compare/master...hiimbex:mentor-url-fix?expand=1#diff-5b32a6df0eaa38f0b026f297bffa52e8R14) and [rust/k.json](https://github.com/exercism/website-copy/compare/master...hiimbex:mentor-url-fix?expand=1#diff-a309c656815d8fb91ae03fd8b8016862R6), since they're original state was to have their URLs set to `"null"`, I wasn't sure whether they should be switched to `null` or `""`.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -20,16 +20,6 @@ task :test do
 
         username = mentor["github_username"]
 
-        if url.strip.empty?
-          errors << "Link URL should be null for %s in %s" % [username, file]
-          next
-        end
-
-        if url.strip == "null"
-          errors << "Link URL should be null, not the string 'null' for %s in %s." % [username, file]
-          next
-        end
-
         if url.strip != url
           errors << "Link URL has extraneous whitespace for %s in %s." % [username, file]
           next
--- a/Rakefile
+++ b/Rakefile
@@ -20,6 +20,11 @@ task :test do
 
         username = mentor["github_username"]
 
+        if url.strip == "null"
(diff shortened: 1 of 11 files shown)


kytrinyx:
Would you switch `"null"` to null? They seem to be indicating nullness...

hiimbex:
@kytrinyx updated!

I also wanted to add that the existing code [here](https://github.com/exercism/website-copy/blob/master/Rakefile#L33-L36) verifies that the empty strings are whitespace-free. I additionally verified this by watching the tests fail locally white a whitespace string. 🙂

Rakefile
```diff
@@ -16,20 +16,10 @@ task :test do
...
+        next if url.nil? || url.empty?
 
         username = mentor["github_username"]
 
-        if url.strip.empty?
-          errors << "Link URL should be null for %s in %s" % [username, file]
-          next
-        end
-
-        if url.strip == "null"
```
kytrinyx: We need to keep this behavior.

kytrinyx:
This is excellent. Yes--definitely want that extra check for extraneous whitespace. ❤️ 💛 💚 💙 💜 🖤

PR #307, opened 2018-08-07
added missing teams to faq

barnzdan:
https://github.com/exercism/exercism.io/issues/3815


cowlike:
I'm really not trying to be an ass, and I know how delivery can be, but ***coming in the next few days*** was an issue comment from 25 days ago (actually that comment said in a day or so). It probably should just be something like ***as soon as we can finish it*** or something similar.

barnzdan:
@cowlike completely fair, I just grabbed what was replied in that comment. I've pushed better wording on that.

PR #312, opened 2018-08-09
Added 'Citizen Code of Conduct' link

ZapAnton:
Should be useful, since it ensures that both the Exercism maintainers and people, who have just learned that `Citizen Code of Conduct` exists, are referring to the same document.


amaliacardenas:
@ZapAnton Thank you very much. :)

PR #416, opened 2018-09-05
Update `~/bin/exercism` output copy

serixscorpio:
It seems the current website copy may be outdated since the v3 cli output has
changed slightly.  Update the copy to match actual v3 cli output.

actual v3 cli output:

current website copy:


kntsoriano:
Thank you so much!

PR #434, opened 2018-09-14
mentor bio for juhlig on the Erlang track

juhlig:


mentors/erlang/j.json
```diff
@@ -2,8 +2,8 @@
   {
     "github_username": "juhlig",
     "name": "Jan Uhlig",
-    "link_text": "",
-    "link_url": "",
+    "link_text": null,
+    "link_url": null,
```
NobbZ: Have you considered @juhlig have set it to `""` on purpose? `null` means to infer those values from github, while `""` means "leave them blank", at least this is how I understand the README.

@juhlig can you please clarify your intention?

PR #548, opened 2018-11-05
Add mentor bio for ruby track

dipto0321:
Hi,
Add mentor bio for dipto47 on ruby track


dipto0321:
Forget to add labels and other necessary things.

PR #549, opened 2018-11-05
Add mentor bio for ruby track

dipto0321:


mentors/ruby/d.json
```diff
@@ -14,5 +13,13 @@
...
+  {
+    "github_username": "dipto0321",
+    "name": "Dipto Karmakar",
+    "link_text": null,
+    "link_url": null,
+    "avatar_url": null,
+    "bio": "I am passionate about programing. Love to build new things with Ruby, python and Javascript. I am enjoying too much by sharing my knowledge and learning from others. Love to contribute on open source projects"
   }
-]
+]
```
rpottsoh: Try to avoid making changes that are outside of the scope of the PR.

mentors/ruby/d.json
```diff
@@ -1,5 +1,4 @@
-[
-  {
+[{
```
rpottsoh: Try to avoid making changes that are outside of the scope of the PR.  The two unnecessary changes I have flagged are minor so I will go ahead and provide approval for this PR.

dipto0321: Thanks. In the future, I will be aware of these.

PR #553, opened 2018-11-06
Kindergarten garden mentoring notes for Python

bwbeach:


iHiD:
Looks good. Could whoever merges please squash these.

tracks/python/exercises/kindergarten-garden/mentoring.md
```diff
@@ -0,0 +1,36 @@
...
+    ]
+
+    PLANT_NAMES = {
+        'C' : 'Clover',
+        'G' : 'Grass',
+        'R' : 'Radishes',
+        'V' : 'Violets'
+    }
+
+    def __init__(self, diagram, students=DEFAULT_STUDENTS):
```
yawpitch: I would not accept a solution that used a _mutable_ default value; any mutation that happens at any point to the default will affect all future calls that use the default.

The usual idiom would be to do:

```python
def __init__(self, diagram, students=None):
    self.students = sorted(students if students is not None else self.DEFAULT_STUDENTS)
```

tracks/python/exercises/kindergarten-garden/mentoring.md
```diff
@@ -0,0 +1,36 @@
...
+        self.rows = diagram.split('\n')
```
yawpitch: Also, we should probably encourage the use of `diagram.splitlines()`, as that would more portably handle **diagram** values created on OS that do not use the single char `\n` newline convention.

tracks/python/exercises/kindergarten-garden/mentoring.md
```diff
@@ -0,0 +1,36 @@
...
+        return [self.PLANT_NAMES[letter] for letter in letters]
+```
+
+
+### Common suggestions
+- Using a constant dict to map from letters to plant names, like `PLANT_NAMES` above.
+
+### Talking points
+- List comprehensions for lists of two things aren't that much shorter that just two copies of the code, but they work if the problem expands to more rows.
+- You can build the assignments of plants to students once, in the constructor, on on-demand in the `plants()` method.
```
yawpitch: L36 should probably read "or on-demand in the `Garden.plants` method" ... that said, I'm not sure we should encourage that as the default approach. I'd personally always rather see a solution that uses a **dict** to map each student to their list of plant names; it avoids the costly **list.index** call for every student, and it means the parsing happens only once for all students and never twice (or more) for the same student.

tracks/python/exercises/kindergarten-garden/mentoring.md
```diff
@@ -16,21 +16,25 @@ class Garden(object):
         'V' : 'Violets'
     }
 
-    def __init__(self, diagram, students=DEFAULT_STUDENTS):
-        self.rows = diagram.split('\n')
-        self.students = sorted(students)
+    def __init__(self, diagram, students=None):
+        self.rows = diagram.splitlines()
+        self.students = sorted(students if students is not None else self.DEFAULT_STUDENTS)
+        self.student_index = {name:index for (index, name) in enumerate(self.students)}
```
yawpitch: An index lookup is an interesting approach, but `Garden.plants` will die with a somewhat confusing IndexError if a student who _is_ a member of the class but wasn't assigned a garden is requested (ie **diagram** shorter than **students**). In my mind requesting a valid student who didn't get a garden assigned should return an empty list, while a student who isn't valid should either do the same or potentially throw a KeyError. I'll admit though that what the class is actually _supposed_ to do in this case isn't well defined in the exercise. As is what should be done if the diagram exceeds the number of students.

PR #555, opened 2018-11-06
Remove '* `scan` is faster than `split`'

weilandia:
Wondering if we should remove this line as they are actually very close and split is faster when you don't have to split on a regex.

```ruby
input = "Rolling On The Floor Laughing So Hard That My Dogs Came Over And Licked Me"
Benchmark.bm do |x|
  x.report { 300000.times { input.split.map(&:chr).join.upcase } }
  x.report { 300000.times { input.scan(/\b[a-zA-Z]/).join.upcase } }
  x.report { 300000.times { input.scan(/\b\w/).join.upcase } }
  x.report { 300000.times { input.split(/[ -]/).map(&:chr).join.upcase } }
end

user     system      total        real
   1.284399   0.004155   1.288554 (  1.291912)
   2.382366   0.007289   2.389655 (  2.392864)
   2.397778   0.009972   2.407750 (  2.415239)
   2.416601   0.011869   2.428470 (  2.438584)
```

Split without a regex is significantly faster, but fails the tests. The rest are comparable.

I only noticed this because I was pointing people towards scan as a faster solution and I got called out that `split` was actually faster.  This turned out to be wrong, because their solution only split on spaces (which is faster), but might be good to avoid mentions of speed here.

### DIFF
--- a/tracks/ruby/exercises/acronym/mentoring.md
+++ b/tracks/ruby/exercises/acronym/mentoring.md
@@ -28,7 +28,6 @@ give the regex away for free, especially if they already found `scan(/\w+/)`
 
 ## Talking points
 * If the solution uses Class instead of Module: that's totally acceptable at this stage. No reason to discuss.
-* `scan` is faster than `split`
 * It can be hard to catch all the possible delimiters. 
 * `upcasing` the whole string is (10% ?) slower than `upcasing` the acronym only. 
 * Store the regex in a constant or a variable, mostly so it can be named. 
--- a/tracks/ruby/exercises/acronym/mentoring.md
+++ b/tracks/ruby/exercises/acronym/mentoring.md
@@ -28,6 +28,7 @@ give the regex away for free, especially if they already found `scan(/\w+/)`
 
 ## Talking points
 * If the solution uses Class instead of Module: that's totally acceptable at this stage. No reason to discuss.
+* Because of the regex needed for `split`, `scan` is in this exercise (marginally) faster.
 * It can be hard to catch all the possible delimiters. 
 * `upcasing` the whole string is (10% ?) slower than `upcasing` the acronym only. 
 * Store the regex in a constant or a variable, mostly so it can be named. 


iHiD:
I had a similar conversation with @kotp earlier on a different solution. He might have value to add to this conversation.

iHiD:
So this is very interesting, because the whole point of this exercise is to introduce `scan`. @F3PiX can give info on why `scan` should be introduced at this point in the track.

So I totally agree with your analysis that we shouldn't be saying `scan` is faster than `split` here. I wonder if we should change the tests to enforce using `scan` as the correct solution (ie introduce the *need* for a regexp).

Either way, this is a really great PR - thank you.

weilandia:
Works for me -- pr updated.  Thanks for the thoughtful comments!

iHiD:
> IIRC the current policy is to not enforce implementation details in the tests, and I like that.

My point was that if the point of the exercise is to introduce `scan` then `scan` needs to be the best way to solve the exercise.  It turns out that the tests **do** enforce this by having the dash example in them. If they didn't then I was suggesting we should add a test that did. It wasn't clear from the original PR that we're doing a test that stops `split` being faster :)

kotp:
An updated benchmark...

```ruby
require 'benchmark/ips'

input = "Rolling On The Floor Laughing So Hard That My Dogs Came Over And Licked Me"

Benchmark.ips do |x|
  x.report ('split with map') { input.split.map(&:chr).join.upcase }
  x.report ('scan with a-zA-Z ') { input.scan(/\b[a-zA-Z]/).join.upcase }
  x.report ('scan with \w') { input.scan(/\b\w/).join.upcase }
  x.report ('split with regexp') { input.split(/[ -]/).map(&:chr).join.upcase }
  x.compare!
end
```

Report looks like:

```
Warming up --------------------------------------
      split with map    18.340k i/100ms
   scan with a-zA-Z     11.708k i/100ms
        scan with \w    11.662k i/100ms
   split with regexp    11.717k i/100ms
Calculating -------------------------------------
      split with map    198.336k (± 2.4%) i/s -      1.009M in   5.088826s
   scan with a-zA-Z     116.773k (±12.2%) i/s -    573.692k in   5.020656s
        scan with \w    122.347k (± 5.0%) i/s -    618.086k in   5.070323s
   split with regexp    122.543k (± 2.6%) i/s -    621.001k in   5.071095s

Comparison:
      split with map:   198336.0 i/s
   split with regexp:   122543.2 i/s - 1.62x  slower
        scan with \w:   122346.8 i/s - 1.62x  slower
   scan with a-zA-Z :   116772.6 i/s - 1.70x  slower

```

PR #558, opened 2018-11-07
Add mentor notes for elixir/binary-search

angelikatyborska:


iHiD:
Looks great! Thanks :) If possible could you keep the whitespace trim at the bottom, just so it doesn't get parsed into extra empty paragraphs on website. But that's not a blocker for me.

Someone else should review the code though as I don't speak Elixir (cc @exercism/elixir).

angelikatyborska:
Sure thing, done. It skipped my mind that now we need to pay attention how the notes will look on the website.

PR #563, opened 2018-11-08
Add alexZielonko to JavaScript mentors

alexZielonko:


mentors/javascript/a.json
```diff
@@ -15,20 +15,28 @@
...
-  "bio": "Javascript developer, father, and amateur everything else."
- },
+  {
+    "github_username": "andrew-prior",
+    "name": "Andrew Prior",
+    "link_text": null,
+    "link_url": null,
+    "avatar_url": null,
+    "bio": "Javascript developer, father, and amateur everything else."
+  },
```
kotp: Normally you would not want to make changes other than that which is related to the pull request.  Alignment isn't.  If this were in a separate commit, it would be an improvement.

I think it is fine to accept in this case though.

PR #569, opened 2018-11-10
Improve solutions for Ruby's Isogram

serg-kovalev:


tracks/ruby/exercises/isogram/mentoring.md
```diff
@@ -1,12 +1,27 @@
 ### Reasonable solutions
 
+Using `uniq`
 ```ruby
 def isogram?(string)
-  letters = string.downcase.scan(/[a-z]/)
+  letters = string.downcase.scan(/[[:alpha:]]/)
```
kotp: Since it is downcased, perhaps the more explicit search for lowercase letters makes sense.  This expands the search to twice as many characters without benefit.

tracks/ruby/exercises/isogram/mentoring.md
```diff
@@ -1,12 +1,27 @@
...
+class Isogram
+  def self.isogram?(text)
+    seen_before = Set.new
+    text.downcase.
+      scan(/[[:alpha:]]/).
+      all? { |l| seen_before.add?(l) }
+  end
+end
+```
+
```
kotp: Missing the justification for bringing in an external library (even though it is standard library).

tracks/ruby/exercises/isogram/mentoring.md
serg-kovalev: True, but `[a-z]` cover letters only from English alphabet comparing to `[[:alpha:]]`
http://www.rubular.com/r/x6xKwSkTSl
It's out of description, but why not to have it in place?

tracks/ruby/exercises/isogram/mentoring.md
serg-kovalev: I did benchmarks on this. It's about 1.5x slower comparing to native C++ implementation of `uniq`.
```
2.2.3 :003 > strings = ["isogram", "eleven", "Alphabet", "six-year-old", "Emily Jung Schwartzkopf", "accentor"]
 => ["isogram", "eleven", "Alphabet", "six-year-old", "Emily Jung Schwartzkopf", "accentor"] 
2.2.3 :004 > Benchmark.bm do |x|
2.2.3 :005 >     x.report { 1_000.times { strings.each { |e| Isogram.isogram?(e) } } }
2.2.3 :006?>   x.report { 1_000.times { strings.each { |e| Isogram.isogram_w_uniq?(e) } } }
2.2.3 :007?>   end
       user     system      total        real
   0.070000   0.000000   0.070000 (  0.068080)
   0.050000   0.000000   0.050000 (  0.049794)
```
But if there lots of repeating chars in the beginning of a string. Execution timings are almost the same:
```
2.2.3 :009 > strings = %w(isogram eleven Alphabet six-year-old aaaaaaaaaabbbbbbbbbcccccc)
 => ["isogram", "eleven", "Alphabet", "six-year-old", "aaaaaaaaaabbbbbbbbbcccccc"] 
2.2.3 :010 > Benchmark.bm do |x|
2.2.3 :011 >     x.report { 1_000.times { strings.each { |e| Isogram.isogram?(e) } } }
2.2.3 :012?>   x.report { 1_000.times { strings.each { |e| Isogram.isogram_w_uniq?(e) } } }
2.2.3 :013?>   end
       user     system      total        real
   0.040000   0.000000   0.040000 (  0.037315)
   0.030000   0.000000   0.030000 (  0.038422)
```

When I solved this exercise myself, I used the first 'reasonable solution'. But I was impressed with implementation by a student. What a nice usage of `Set`'s `add?` in coupe with `Enumerable`'s `all?`
I do agree that it loses a performance game. But it's beautiful solution, isn't it?

Would be glad to hear your thoughts on this
Thanks!

tracks/ruby/exercises/isogram/mentoring.md
kotp: In that case, it may be worthwhile contributing to the specifications repository to get the exercise changed.

tracks/ruby/exercises/isogram/mentoring.md
serg-kovalev: Agreed. In the same time, we don't talk about implementation difficulty of `Array`'s `uniq`. It's a top hierarchy for us in Ruby (internals are written in C).

If we want a student to write their own algorithm of searching unique chars, we would want preventing usage of `uniq` as well. I'm just thinking about what the goal of this exercise is.

kotp: Right, and we likely don't because `uniq` is a core part of Ruby, we don't have to bring in an external (even if it is standard) library.

I don't think anyone has stated about writing their own implementation of `uniq` but when they go out to grab something that may be available, looking at external things, perhaps they have not explored what is available in core Ruby.

serg-kovalev: I will move this new 'reasonable' solution to talking points. I'm sure if a student found `Set` standard library and used its `add?` method. We should appreciate their diligence.

Thanks for such a good discussion on this!

pgaspar: I went through an interesting exploration of these things with a student once. I think it's relevant to this conversation: https://exercism.io/mentor/solutions/dec89cad6c3241c4a6c3c7319f395c06

NobbZ: Sadly that discussion can only been read by mentors of that track and the student. Could you anonymize and quote the relevant parts of the discussion?

pgaspar: I'll try!

The student submitted two solutions: one using a Set and another using `Array#uniq`.

**Student:**

> I came up with 2 approaches (second one is commented out) and would love guidance on which is better/more ruby-like. My understanding (correct me if I'm wrong) is that by iterating through and adding each character to a set it will be slightly better performing than using the builtin Array.uniq since that will iterate through each character

**My response**

> Hello!
> 
> First of all, thank you for your questions - they motivated me to explore the performance implications of both approaches more thoroughly than I had before 👍
> 
> **Using Array#uniq**
> 
> To find out what exactly Ruby is doing when we call `uniq` I looked at the Ruby source code. Ruby documentation typically allows you to see the method source, and [`Array#uniq`](https://ruby-doc.org/core-2.5.0/Array.html#method-i-uniq)'s implementation can be seen [here](https://github.com/ruby/ruby/blob/22d8ff8c96206314c0f55814a0ae7056adb5a187/array.c#L4643).
> 
> The most relevant part is:
> 
> ```c
> hash = ary_make_hash(ary);
> uniq = rb_hash_values(hash);
> ```
> 
> What Ruby does is convert the array to a Hash and then get the Hash's values (they could use the Hash's keys just as well, as we'll see below).
> 
> If you look for `ary_make_hash` in the [Array source code](https://github.com/ruby/ruby/blob/22d8ff8c96206314c0f55814a0ae7056adb5a187/array.c#L4140-L4144) you find that it creates an hash and then calls `ary_add_hash`, which is defined in the same file [here](https://github.com/ruby/ruby/blob/22d8ff8c96206314c0f55814a0ae7056adb5a187/array.c#L4118-L4127).
> 
> This code does what you assumed it would do! It iterates over the array and adds each element to the hash. We can mimic this in Ruby ourselves:
> 
> ```ruby
> ary = [4,1,2,4,3,2,5]
> hsh = {}
> 
> # Iterate over the array, adding each element to the hash
> ary.each { |el| hsh[el] = el }
> 
> hsh
> # => {4=>4, 1=>1, 2=>2, 3=>3, 5=>5}
> 
> hsh.values
> # => [4,1,2,3,5]
> 
> hsh.keys
> # => [4,1,2,3,5]
> 
> ary.uniq
> # => [4,1,2,3,5]
> ```
> 
> We did it! Our own implementation of `uniq` 🙌
> 
> `==` may also be doing some iteration. The documentation seems to point that way:
> 
> > Two arrays are equal if they contain the same number of elements and if each element is equal to (according to Object#==) the corresponding element in other_ary.
> 
> One way to avoid that would be to compare sizes:
> 
> ```ruby
> letters.uniq.size == letters.size
> ```
> 
> **Using a Set**
> 
> Now, how about using Set? First notable thing is that to use Set you need to require it. That can be a performance and maintainability consideration by itself - am I gaining enough value that makes requiring these [817 extra lines](https://github.com/ruby/ruby/blob/trunk/lib/set.rb) worth it?
> 
> So, I wanted to understand what happens when you add something to a Set. It must have an underlying data structure supporting it - somewhere data is stored.
> 
> The documentation says:
> 
> > It is a hybrid of Array's intuitive inter-operation facilities and Hash's fast lookup.
> 
> And further down:
> 
> > Set uses Hash as storage
> 
> Interesting! Looking up `add?` you find it [here](https://github.com/ruby/ruby/blob/trunk/lib/set.rb#L360-L362):
> 
> ```ruby
> def add?(o)
>   add(o) unless include?(o)
> end
> ```
> 
> `add` does `@hash[o] = true`. `include?` is defined [here](https://github.com/ruby/ruby/blob/trunk/lib/set.rb#L253-L255) and is simply:
> 
> ```ruby
> def include?(o)
>   @hash[o]
> end
> ```
> 
> This makes sense! Hash lookups are much faster than iterating over a collection and there's the added benefit of being able to return `false` earlier, which `uniq` does't give you.
> 
> This also tells you it's very much possible to ditch Set and do this yourself:
> 
> ```ruby
> seen_characters = {}
> letters.each do |character|
>   return false if seen_characters[character]
>   seen_characters[character] = true
> end
> true
> 
> # Or the possibly more idiomatic,
> # but maybe a bit confusing:
> 
> seen_characters_count = Hash.new(0)
> letters.none? do |character|
>   seen_characters_count[character] += 1
>   seen_characters_count[character] > 1
> end
> ```
> 
> 
> 
> This way you can make things even faster, without requiring Set.
> 
> **Final thoughts**
> 
> From our exploration your assumption was right! Of course, to test this more thoroughly you could do some benchmarking with really big input strings to see how much faster using Set or our own hash is, compared with the `uniq` approach.
> 
> > I came up with 2 approaches (second one is commented out) and would love guidance on which is better/more ruby-like
> 
> Which approach is better depends on your focus. If you expect performance will be an essential concern for your problem, one of the Set-like implementations would be better.
> 
> On the other hand, if maintainability is more important for you than a (possibly marginal) performance gain, as is the case in many (the majority?) of modern applications, then use the `uniq` approach as it's shorter and easier to grasp. (This is typically the approach we aim for when mentoring this exercise.)
> 
> I'm going to approve your submission because your assessment about the trade-offs in this exercise was right on the money 👍 Now it's up to you to pick the approach you think is better for your circumstances. Well, imagined circumstances, since this is a practice exercise 😄.

tracks/ruby/exercises/isogram/mentoring.md
NobbZ: All exercises in exercism always ever cover only 7 bit ASCII.

Unless the track specifies extra unicode tests, I do not think that any unicode specialties need to be covered.

Using `[[:alpha:]]` will bloat the resulting DFA without any need.

But as we are not necissarily optimising for speed here, I have to admit, I consider `[[:alpha:]]` more readable than `[a-z]` or `\pLl` (or whatever lowercase letter group was).

rpottsoh:
Been over a month since there has been any activity here....  Can this be placed in the category of "something is better than nothing" so it may be merged?

F3PiX:
Hmmm. I feel most mentors think that introducing Set does not qualify as a 'reasonable' solution. 

I'd vote for not implementing it as is, but closing.

kotp:
I vote for a close as well, though there are improvements that don't include the `Set` library.  We can always reopen.

PR #570, opened 2018-11-10
Add mentoring notes for Python binary-search

bwbeach:


tracks/python/exercises/binary-search/mentoring.md
```diff
@@ -0,0 +1,72 @@
...
+        else:
+            return (mid_index + 1) + binary_search(list_of_numbers[mid_index+1:], number)
+```
+
+### Common suggestions
+- When the code gets the wrong answer, find a small test case that fails and suggest thinking through the logic to see where it goes wrong.
+
+### Talking points
+- Using a single variable to track where the midpoint is on each iteration is not enough, because you also need to track how long the current sub-list is.
+- You can represent the high end of the range as the last index where the value could be (inclusive), or one past that (exclusive).  Both work.  The exclusive approach seems more natural in Python, because that's the way slicing works.
```
rpottsoh:
```suggestion
- You can represent the high end of the range as the last index where the value could be (inclusive), or one past that (exclusive), both work.  The exclusive approach seems more natural in Python, because that's the way slicing works.
```

rpottsoh: `Both work` doesn't make sense as part of the next sentence.  The next sentence doesn't discuss two different things like the previous sentence does.  The next sentence only mentions the `exclusive approach`.

rpottsoh:
```suggestion
- You can represent the high end of the range as the last index where the value could be (inclusive), or one past that (exclusive).  The exclusive approach seems more natural in Python, because that's the way slicing works.
```

Another suggested way to change this.

rpottsoh:
@bwbeach do you have any questions or need some help with making any of the suggested changes?

bwbeach:
Yes.  I have a question...

I had thought this was merged, so I deleted my repo last weekend.  Does this commit exist somewhere I can pull from and then update the pull request?

rpottsoh:
I think I can merge this.  Afterwards you are welcome to fork this repo and open another PR to make further changes to the file.

PR #576, opened 2018-11-12
go/raindrops: mention exalysis mentoring tool

tehsphinx:
As discussed with @iHiD we want to mention github.com/tehsphinx/exalysis to help mentors start mentoring.


iHiD:
@tehsphinx I merged this then realised it should be `###` not `##` - mind changing pls?

PR #581, opened 2018-11-15
Add mentor bio for Makan

MakanTaghizadeh:


MakanTaghizadeh:
@rpottsoh Oops, my bad! That's absolutely true. Yeah. I'll surely take care of it in the future. Thank you so much 👍

PR #582, opened 2018-11-15
rust/leap: Fixed operator typo

ZapAnton:
Small typo fix: `$` -> `%`


kotp:
Thank you @ZapAnton Good catch!

PR #605, opened 2018-11-21
javascript/simple-cipher: Add basic notes including benchmark

SleeplessByte:
Add basic notes with variations and documentation links. Also has a benchmark for those students/mentors who care.


serixscorpio:
I found this note super useful (even though not yet merged), especially the breakdown into different parts like 'Key Validation', 'Key Generation', 'Encoding/Decoding', (and to a certain extent 'Transformation' might also be considered another part).  It helps me zoom in to focus on a particular area of a solution.  Thanks @SleeplessByte !

PR #595, opened 2018-11-19
javascript/leap: Add mentor notes

SleeplessByte:
Basic mentor notes with links to documentation about logical operators, precedence and talking points.

PR #607, opened 2018-11-22
javascript/leap: Improve mentor notes

SleeplessByte:
This PR tries to improve on the current mentor notes for `javascript/leap`, following up on #595.

- Add a summary of goals, similar to the other JavaScript mentor notes
- Fixes the reasonable solution
  The first reasonable solution is missing parentheses, which means that when the left hand of the logical AND short-cirtcuits, the right hand of the logical OR is always executed. The left hand short-circuits if a year is not divisible by 4, so the right hand of the logical OR will ALWAYS be false (as it can't be divisible by 400 then).
- Add reasonable solution using a function generator
  Uncommon but not rare are functional solutions and solutions using anonymous functions. Most of the time they get the order of operations and order of evaluations wrong. This adds notes how to deal with that.
- Add suggestion about order of evaluation
  Operator precedence and short-circuiting is important in this exercise. Not just order of operations.

### DIFF
--- a/tracks/javascript/exercises/leap/mentoring.md
+++ b/tracks/javascript/exercises/leap/mentoring.md
@@ -17,14 +17,11 @@ Variations check for truthy `year % m` and falsy `(!year % m)`.
 - If there are more than two "returns" or more than three "tests", suggest that there are just two cases that return `true`:
   - A year is a multiple of 4 *and not* 100
   - A year is a multiple of 4, 100, and 400
-- Order of operations matter:
-  - 75% of all years *cannot* be leap years because they are not mulitples of 4; test `year % 4 === 0` first.
+- If the order of the tests is not `disivible by 4`, `divisible by 100`, `divisible by 100`, explain that the _order of operations_ matter:
+  - 75% of all years *cannot* be leap years because they are not multiples of 4; test the common path `year % 4 === 0` first.
   - 98.97% of all years that are multiples of 4 are not multiples of 100; test `year % 100 !== 0` second.
   - 1.03% of all years that are multiples of 4 are also multiples of 100 and 400; test `year % 400 === 0` third.
-- Eliminate duplicate work:
-  -  Some submissions may repeat the same condition multiple times (`year % 4 === 0` is common); encourage students to resolve this by consolidating and/or reordering conditions.
-- Submissions often explicitly return `true` or `false`; encourage the use of an implicit return.
-- Submissions may include multiple `if/else` statements to satisfy the test suite; encourage users to consider how [logical operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators) can make their solution more concise.
+- Eliminate duplicate work: no year should ever have to be checked multiple times for the same condition; encourage students to resolve this by consolidating and/or reordering conditions.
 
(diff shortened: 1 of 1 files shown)


tracks/javascript/exercises/leap/mentoring.md
```diff
@@ -1,27 +1,67 @@
...
 ### Reasonable solutions
 
 ```javascript
-export const isLeap = (year) => year % 4 === 0 && year % 100 !== 0 || year % 400 === 0;
+export const isLeap = (year) => year % 4 === 0 && (year % 100 !== 0 || year % 400 === 0);
 ```
+Variations check for truthy `year % m` and falsy `(!year % m)`.
 
 ```javascript
-export const isLeap = (year) => !(year % 4) && year % 100 || !(year % 400);
```
alexZielonko: The statement added in this pull request,

> Variations check for truthy `year % m` and falsy `(!year % m)`.

is less ambiguous without the removal of this snippet. Consider including both lines for greater clarity.

tracks/javascript/exercises/leap/mentoring.md
```diff
@@ -1,27 +1,67 @@
...
-```
+  function isDivisibleByFn(mod) {
+    return number => number % mod === 0;
+  }
+
+  const by4 = isDivisibleByFn(4);
+  const by100 = isDivisibleByFn(100);
+  const by400 = isDivisibleByFn(400);
+
+  export const isLeap = year => by4(year) && (!by10(year) || by400(year))
```
alexZielonko: Should students walk away from the Leap exercise with an understanding of currying or partial application? 

Consider the following solution as it seems more applicable to the majority of students at this stage in the track,

```JavaScript
function isDivisibleBy (year, mod) {
  return year % mod === 0;
}

export const isLeap = (year) => {
  const by4 = isDivisibleBy(year, 4);
  const by100 = isDivisibleBy(year, 100);
  const by400 = isDivisibleBy(year, 400);

  return by4 && (!by100 || by400);
}
```

alexZielonko: The "reasonable solution" introduced in this pull request fails the exercise's test suite.

tracks/javascript/exercises/leap/mentoring.md
SleeplessByte: Agreed. Added it back in!

tracks/javascript/exercises/leap/mentoring.md
SleeplessByte: I simply copied what about 1/10 students had. I agree that your solution is easier to read, but it is not identical to the non-functional solution. You make all comparisons and the whole "order of operations and evaluations" is sorta pointless. Instead I combined what was there with what you wrote!

SleeplessByte: Fixed. Don't know what happened there.

SleeplessByte:
@alexZielonko Thank you for reviewing! They are good comments. As per my response: I don't want to teach students currying at this point. I am basing the notes on the solutions + what I intuitively think is the goal of the exercise.

I've amended the functional solution using your suggestion plus my original code. It now is super clear (to me) and is lazily executed, as is the inline solution.

PR #608, opened 2018-11-22
javascript/hamming: Add mentor notes

SleeplessByte:
Add mentor notes and benchmark on various implementations.

Unlike the Ruby variant, JavaScript doesn't have the same niceties when it comes to enumerables and special reducers that allow programmers to reduce noise without sacrificing speed.

### DIFF
--- a/tracks/javascript/exercises/hamming/mentoring.md
+++ b/tracks/javascript/exercises/hamming/mentoring.md
@@ -3,17 +3,17 @@ This exercise focuses on:
 
 ## Reasonable solutions
 ```javascript
-function compute(leftStrand, rightStrand) {
-    const length = leftStrand.length;
-    if (length != rightStrand.length) {
-        throw Error('left and right strands must be of equal length');
-    }
+export function compute(leftStrand, rightStrand) {
+  const { length } = leftStrand;
+  if (length !== rightStrand.length) {
+    throw Error('left and right strands must be of equal length');
+  }
 
-    let distance = 0;
-    for (let i = 0; i < length; i++) {
-        distance += leftStrand.charCodeAt(i) !== rightStrand.charCodeAt(i)
-    }
-    return distance;
+  let distance = 0;
+  for (let i = 0; i < length; i++) {
+    distance += leftStrand.charCodeAt(i) !== rightStrand.charCodeAt(i);
+  }
+  return distance;
 }
 ```
 Variations include `String#charAt(i)` and `String#[i]` (similar speed)
--- a/tracks/javascript/exercises/hamming/mentoring.md
+++ b/tracks/javascript/exercises/hamming/mentoring.md
@@ -1,0 +1,62 @@
+This exercise focuses on:
+- control flow loops: how to _iterate_ through two collections and how to _count_ based on a condition
+
+## Reasonable solutions
+```javascript
+function compute(leftStrand, rightStrand) {
+    const length = leftStrand.length;
+    if (length != rightStrand.length) {
+        throw Error('left and right strands must be of equal length');
+    }
+
+    let distance = 0;
+    for (let i = 0; i < length; i++) {
+        distance += leftStrand.charCodeAt(i) !== rightStrand.charCodeAt(i)
+    }
+    return distance;
+}
+```
+Variations include `String#charAt(i)` and `String#[i]` (similar speed)
+
+Using `continue` (similar speed):
+```javascript
+if (leftStrand.charCodeAt(i) === rightStrand.charCodeAt(i)) {
+  continue;
+}
+distance += 1;
+```
+
+Explicit addition with ternary (similar speed):
+```javascript
+distance += leftStrand[i] === rightStrand[i] ? 0 : 1
+```
+Variations include re-assignment using `? distance : distance + 1`
+
+### Using streams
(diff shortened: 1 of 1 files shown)


tracks/javascript/exercises/hamming/mentoring.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+Explicit addition with ternary (similar speed):
+```javascript
+distance += leftStrand[i] === rightStrand[i] ? 0 : 1
+```
+Variations include re-assignment using `? distance : distance + 1`
+
+### Using streams
+Vanilla JavaScript doesn't have a `Array#zip(array)` or `Enumerable#count(expression)` function, yet, but it can still be attempted by using `split` with `reduce` or `replace`. These solutions work, but are a factor 5-10 slower. You might want to challenge a student to use a for loop. **Note** that performance is heavily impacted by the JavaScript engine and things like bable/bubble implementation/version.
```
NobbZ: If those functions do not exist in vanilla JavaScript, were do they exist then?

tracks/javascript/exercises/hamming/mentoring.md
```diff
@@ -0,0 +1,62 @@
...
+## Common suggestions
+- When going for performance, at time of writing these notes, a `for` loop, counting each difference in a mutable variable, _always_ trumps any form of `String#split` + `String#replace` or `Array#reduce`: [Benchmarks](https://run.perf.zone/view/Count-differences-between-strings-1542904507660).
+
+
+## Talking points
+- If a student uses streams:
+  - `reduce` and `filter` are better candidates than `forEach` (use a for loop)
+  - These functions have second arguments `index`. Don't do manual bookkeeping of an index
+- If a student uses a for loop:
+  - `continue` is nice to mention
```
NobbZ: I don't think so.

The use of `continue` in the loop on equal chars encourages inverted logic.

This can cause confusion.

NobbZ: In common CS hamming distance is the count of different symbols in a ordered sequences of the same length. So by this definition incrementing on inequality seems to be much more natural than skipping on equality.

Of course, both views are semantically equivalent in this case, and in many others as well.

But I often seen projects fail because someone inverted löogic for some speedgain in certain edge cases. Somewhere inbetween they lost track of the fact that it was inverted logic and assumed it were regular logic and that caused bugs.

The other thing is, that I generally consider control flow using `continue`, `break` and (even worse) `reapeat` (for languages where its available) as bad style. It interupts the regular flow and costs extra brain capacity to resort where to continue reading the code.

Of course, in some circumstances the really make the code easier to understand or massively reduce complexity of the loop condition (in C style for loops), but in this case I do not think that a `continue` is reasonable.

tracks/javascript/exercises/hamming/mentoring.md
NobbZ: I think it is easier for the reader if you either mention those (or even link to them) in a side note or don't mention it at all.

Also raw speed is often not the (first) thing to optimize for. Until a bottleneck has been identified one should write code that is easily understood by the team. (Of course FP style code is not considered readable by most people ;) )

tracks/javascript/exercises/hamming/mentoring.md
SleeplessByte: I think this is just where we disagree. Using control flow is perfectly fine and can be taught to a student, which is why it's mention and not encourage. Often I see students do convoluted things with if else branched or inline functions in order to achieve "continue".

NobbZ: Then lets agree to disagree.

tracks/javascript/exercises/hamming/mentoring.md
SleeplessByte: (p.s. the ruby notes I use and mentoring I do is exactly what you suggest: use stdlib functional style and try to remove all bookkeeping variables.

I still add the streams solution to these notes because they are valid and readable and I prefer them :p. "Might want to challenge" for me is literally that.)

NobbZ: > Pulling in libraries to fulfill tasks you can do with a simple for loop is not something we want to encourage. Dependency hell is a common problem in JavaScript.

Yupp, I understand. But you assume that every mentor who shall work with your text knows about the modules/classes/methods/whatever. This might not be true at all, therefore you should provide those explanation for the other mentors.

> Optimising for raw speed on the other hand ís what has been done (over the past years) in most mentor sessions on the JavaScript track.

Mentoring is only available for ~6 months now.

And if your current mentors are all arguing for speed, then you should propably encourage a shift of mind.

Of course I can understand, that especially for vanilla javascript, the most performant reason is also one of the most performant, but that is just a lucky coincident.

> Personally I think intuitively knowing that using a simple for loops is almost always faster is a good thing and should be taught rather sooner than later.

For most JS developers I had contact with so far I think that also a straight `for`ward loop (pun intended) is also the most readable way considering the more FP style approaches.

So I do not see anything bad in a `for` loop, but I don't like the priority that is given to raw execution speed in the mentor notes.

SleeplessByte: I can definitely link to some of the fp style extensions. That is a good idea.

Mentoring has only been available for that long, but before that there was community commentary and as you can see I've been on here for four years. For some reason speed is always a thing. Then again, outside of exercism, premature optimisation with JavaScript had always been a thing.

When I'm mentoring I do approve readable answers over for loops -- I dont agree with you that _mentioning_ other solutions are slower is something we shouldn't have in the mentor notes. 

The issue with JavaScript is that on low end devices, a 2x or in this case up to 10x can actually be so significant it is visible :cry:. Of course not what these simple exercises (unless they make a O(2**n) implementation))…

SleeplessByte: Funny thing though, myself I prefer the functional and streams approach almost always over a for loop when it comes to readability :flushed:

PR #615, opened 2018-11-24
Bash: Armstrong Numbers: Create mentoring notes

rpalo:
I noticed that the Bash track is sadly lacking in Mentoring notes and I thought it might help us get more mentors active.  So here's the first one 😄


NobbZ:
@kotp would you mind reviewing this?

rpalo:
Woohoo!  Thanks!

PR #621, opened 2018-11-29
Rust: add mentoring notes for sublist

yaahallo:


tracks/rust/exercises/sublist/mentoring.md
```diff
@@ -0,0 +1,45 @@
...
+```rust
+#[derive(Debug, PartialEq)]
+pub enum Comparison {
+    Equal,
+    Sublist,
+    Superlist,
+    Unequal,
+}
+
+pub fn sublist<T: PartialEq>(_first_list: &[T], _second_list: &[T]) -> Comparison {
```
NobbZ: Shouldn't these be named without the underscore, as the arguments are used in the functions body?

PR #622, opened 2018-11-29
Rust: add mentoring notes for luhn

yaahallo:


tracks/rust/exercises/luhn/mentoring.md
```diff
@@ -0,0 +1,28 @@
...
+- utilize higher order functions on iterators and on option types
+
+### Examples
+
+```rust
+pub fn is_valid(code: &str) -> bool {
+    code.chars()
+        .rev()
+        .filter(|c| !c.is_whitespace())
+        .try_fold((0, 0), |(sum, c), d| {
```
NobbZ: I have to be honest, I haven't solved this one in rust yet, but shouldn't we use `enumerate()` instead of manually bookkeeping the current index?

yaahallo:
Manually doing the book keeping let's you do the length check in the same pass. And it's not very hard to prove correctness so I tend to think either is fine.

PR #625, opened 2018-11-29
Rust: add mentoring notes for clock

yaahallo:


tracks/rust/exercises/clock/mentoring.md
```diff
@@ -0,0 +1,45 @@
+# Clock
```
iHiD: Please don't have exercise names as a heading.

tracks/rust/exercises/clock/mentoring.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Concepts
```
iHiD: Please use `###` for headings, not `##`

tracks/rust/exercises/clock/mentoring.md
```diff
@@ -0,0 +1,45 @@
...
+}
+
+impl fmt::Display for Clock {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:02}:{:02}", self.minutes / 60, self.minutes % 60)
+    }
+}
+```
+
+## Example Comments
```
iHiD: Please don't have titles without content below.

tracks/rust/exercises/clock/mentoring.md
```diff
@@ -0,0 +1,41 @@
+### Concepts
+
+- derive
+- struct
+- traits
+
+### Reasonable solutions
+
+A reasonable solution should do the following:
```
kotp: This indicates that we are awaiting a lit of behaviors that a reasonable solution should do.   Seems to be missing some information.

rpottsoh: @kotp I think you have a valid point here.  This PR appears to be stale.  Should we go ahead and merge as is and let someone else pick this up later?

kotp: Yeah, it will definitely be noticed, and if it isn't updated in the next couple of days (I will let the mentors know in the Rust room) we should merge it, it will draw the needed attention in that way.

rpottsoh: At that time perhaps we should add a blurb to the file that states `information needed here` or something along those lines.

rpottsoh:
```suggestion
A reasonable solution should do the following:
`items needed here`
```

PR #633, opened 2018-11-30
Ruby/HighScores: New Notes Format

F3PiX:
Here's Mentor Notes with a different format.

It's meant to make mentoring easier for mentors, and also improve consistency of mentoring. 
- List the new concepts covered by the exercise 
- Help mentors decide to approve a solution or not
- Help mentors that are fairly new to the language to mentor to idiomatic solutions
- Help mentors to keep track of possible approaches and the discussions on those
- List the collected wisdom of research notes
and in general improve mentor happiness. 

Note on the new **'Minimal Solution for Approval**' : this is meant for the easiest exercises only, where there's only one, maybe two idiomatic solutions. That goes for only a few exercises. In Ruby: TwoFer, Acronym, HighScores.

### DIFF
--- a/tracks/ruby/exercises/high-scores/mentoring.md
+++ b/tracks/ruby/exercises/high-scores/mentoring.md
@@ -1,0 +1,71 @@
+New concepts: Array query methods (+ chaining methods), instantiating a class, getter method or `attr_reader`.
+
+### Minimal solution for approval
+
+```ruby
+
+# version 2+
+
+class HighScores
+  def initialize(scores)
+    @scores = scores
+  end
+
+  def scores  # no need to change getter method to `attr_reader`
+    @scores
+  end
+
+  def personal_best
+    scores.max
+  end
+
+  def latest
+    scores.last
+  end
+
+  def personal_top
+    scores.sort.reverse.take(3) 
+  end
+
+  # composing a string either with string interpolation or with an if.. else.. statement
+  def report
+    difference = "#{personal_best - latest} short of" unless personal_best == latest
+
+    "Your latest score was #{latest}. That's #{difference} your personal best!".squeeze
+  end
+end
+
+```
+
+### Reasonable approaches
+
+- See the one above
+- Defining `@scores`, `@personal_best` and `personal_top` in the constructor.  
+
+### General notes
+- The Instructions point to a beginner friendly explanation of instantiating a class: 
+http://ruby-for-beginners.rubymonstas.org/writing_classes/initializers.html
+and to the rubydocs Array. 
+- Most submissions use an getter method `scores`, not `attr_reader`. 
+That's okay at this point in the track, no need to suggest `attr_reader`.
+- Beginners create the attribute reader, but may call the instance variable instead of the getter. 
+This link gives a nice introduction of getters and setters, and the naming convention in Ruby:
+https://launchschool.com/books/oo_ruby/read/classes_and_objects_part1#accessormethods.
+It's recommended to link to this or your favorite explanation, instead of typing the solution for them.
+- `sort.reverse` is not very performant, but at this point in the track it's more important that students explore the 
+standard lib (and work with methods chaining) than that they know how to optimize.
+
+### Talking points
+- `Array#max` takes an argument; `scores.max(3)` works.
+- Most submissions retrieve the `personal_top` values with the `Array#[]` method. 
+`Array#take` wins in readability. And `Array#max()` wins them all. 
+- The `report` method asks for a conditionally composed string. It's meant to think about the trade-off 'duplication' vs 'readability'. Recommendation: approve any solution unless it's a supercomplicated `if/else statement`. 
+- Consider `String#squeeze` to avoid hard-to-spot spacing in the string itself. 
+
+### Mentor Research
+- `String#<< / String.concat` vs. `String#+`? 
+Concatenation mutates the string instance in-place, String#+, creates new string objects.
+Ruby 2.3 will freeze string literals by default.
+
+### Changelog
+- Version 2 changed method names "highest" -> "personal_best", "top" -> "personal_top"
--- a/tracks/ruby/exercises/high-scores/mentoring.md
+++ b/tracks/ruby/exercises/high-scores/mentoring.md
@@ -4,14 +4,12 @@ New concepts: Array query methods (+ chaining methods), instantiating a class, g
 
 ```ruby
 
-# version 2+
-
 class HighScores
   def initialize(scores)
     @scores = scores
   end
 
-  def scores  # no need to change getter method to `attr_reader`
+  def scores
     @scores
   end
 
@@ -27,44 +25,38 @@ class HighScores
     scores.sort.reverse.take(3) 
   end
 
-  # composing a string either with string interpolation or with an if.. else.. statement
   def report
     difference = "#{personal_best - latest} short of" unless personal_best == latest
 
-    "Your latest score was #{latest}. That's #{difference} your personal best!".squeeze
+    "Your latest score was #{latest}. That's #{difference} your personal best!".squeeze(' ')
   end
 end
 
 ```
-
-### Reasonable approaches
-
-- See the one above
-- Defining `@scores`, `@personal_best` and `personal_top` in the constructor.  
+### Reasonable variants
+ - Using a getter method `scores`, not `attr_reader`: That's okay at this point in the track, no need to suggest `attr_reader`.
+ - Variants in composing the string in `report`, as long as it's not a complicated if/else statement (see Talking Points).
 
 ### General notes
 - The Instructions point to a beginner friendly explanation of instantiating a class: 
 http://ruby-for-beginners.rubymonstas.org/writing_classes/initializers.html
 and to the rubydocs Array. 
-- Most submissions use an getter method `scores`, not `attr_reader`. 
-That's okay at this point in the track, no need to suggest `attr_reader`.
 - Beginners create the attribute reader, but may call the instance variable instead of the getter. 
 This link gives a nice introduction of getters and setters, and the naming convention in Ruby:
-https://launchschool.com/books/oo_ruby/read/classes_and_objects_part1#accessormethods.
+[launchschool: accessor_methods](https://launchschool.com/books/oo_ruby/read/classes_and_objects_part1#accessormethods).
 It's recommended to link to this or your favorite explanation, instead of typing the solution for them.
 - `sort.reverse` is not very performant, but at this point in the track it's more important that students explore the 
-standard lib (and work with methods chaining) than that they know how to optimize.
+standard libraries (and work with method chaining) than that they know how to optimize.
 
 ### Talking points
-- `Array#max` takes an argument; `scores.max(3)` works.
-- Most submissions retrieve the `personal_top` values with the `Array#[]` method. 
-`Array#take` wins in readability. And `Array#max()` wins them all. 
-- The `report` method asks for a conditionally composed string. It's meant to think about the trade-off 'duplication' vs 'readability'. Recommendation: approve any solution unless it's a supercomplicated `if/else statement`. 
-- Consider `String#squeeze` to avoid hard-to-spot spacing in the string itself. 
+- [`Array#max`](https://ruby-doc.org/core/Array.html#method-i-max) takes an argument; `scores.max(3)` works.
+- Most submissions retrieve the `personal_top` values with the `Array#[]` method, which is harder to read than the convenience methods that Ruby offers: `Array#take(3)` or `Array#first(3)` win in readability. And `Array#max(3)` doesn't even need the sorting-and-reverting steps. 
+- The `report` method asks for a conditionally composed string. It's meant to think about the trade-off 'duplication' vs 'readability'. Recommendation: approve any solution unless it's a supercomplicated if/else statement. 
+- Consider `String#squeeze(' ')` to avoid hard-to-spot spacing with string interpolation. 
 
 ### Mentor Research
-- `String#<< / String.concat` vs. `String#+`? 
-Concatenation mutates the string instance in-place, String#+, creates new string objects.
+- [String](https://ruby-doc.org/core-2.5.3/String.html): `String#<< / String.concat` vs. `String#+`? 
+Concatenation mutates the string instance in-place, `String#+` creates new string objects.
 Ruby 2.3 will freeze string literals by default.
 
 ### Changelog


tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
...
+    "Your latest score was #{latest}. That's #{difference} your personal best!".squeeze
+  end
+end
+
+```
+
+### Reasonable approaches
+
+- See the one above
+- Defining `@scores`, `@personal_best` and `personal_top` in the constructor.
```
pgaspar: Not sure I love this approach being here. It's very brittle and breaks as soon as the scores array is modified after instantiation (even though we don't test for this scenario). Would love to hear your thoughts!

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
...
+- `Array#max` takes an argument; `scores.max(3)` works.
+- Most submissions retrieve the `personal_top` values with the `Array#[]` method. 
+`Array#take` wins in readability. And `Array#max()` wins them all. 
+- The `report` method asks for a conditionally composed string. It's meant to think about the trade-off 'duplication' vs 'readability'. Recommendation: approve any solution unless it's a supercomplicated `if/else statement`. 
+- Consider `String#squeeze` to avoid hard-to-spot spacing in the string itself. 
+
+### Mentor Research
+- `String#<< / String.concat` vs. `String#+`? 
+Concatenation mutates the string instance in-place, String#+, creates new string objects.
+Ruby 2.3 will freeze string literals by default.
```
pgaspar: Should we add String interpolation to this list?

tracks/ruby/exercises/high-scores/mentoring.md
F3PiX: I just went with the fact that one of the other mentors approved this approach, to be honest. 
You're totally right that it would break after the scores are modified, but the thing is: we don't test that, and that's for a reason. And I don't like to reject solutions based on reasons that we don't test for. In pure TDD mode, this would suffice to get the tests to pass.
I would be very happy if there's another argument why this is not a good approach.

tracks/ruby/exercises/high-scores/mentoring.md
F3PiX: We could. In general, this is a place where mentors can share their own research notes. So, it's a growing living thing. Let's see if someone wants to add it.

pgaspar: Oh it wasn't 100% clear what this section was about for me and how it would change. Thanks for clarifying :+1:.

tracks/ruby/exercises/high-scores/mentoring.md
pgaspar: > And I don't like to reject solutions based on reasons that we don't test for.

🤔 so do you always accept as soon as the tests pass? There's a lot of idiomatic nuance we can't realistically test for. This is a broader discussion, but from what I've read before I think a lot of mentors require more from a solution than just passing the tests, no?

For example, in Acronym I typically point to using `scan` when people use `split` as suggested in the mentor notes. Do you usually accept `split` solutions and then suggest `scan` or only accept when `scan` (or something in that spirit) is used?

That's why I'm mentioning this - I'd be ok with someone with more experience using this approach (after making sure they understand what it means),  but I'm afraid it can lead to confusion for novice programmers. So I would always follow an approach like this with a discussion on if and why it may be troublesome for you in the future, and I wouldn't recommend it for this exercise.

SleeplessByte:
FWIW, `two-fer` actually has at least 4 solutions that are idiomatic. I think the vagueness of "reasonable solutions and variations" is better than "this is probably what it should be".

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
+New concepts: Array query methods (+ chaining methods), instantiating a class, getter method or `attr_reader`.
```
SleeplessByte: I've actually been adding "this exercise focusses on x y and z" to most of my note submission, so I really like that you're adding that as well. In general I tend to reference to the official ruby documentation, or javascript/typescript and I would encourage you linking the word `Array`, `class` and `attr_reader` here as well.

It helps the mentor because then they are encouraged to use the "official" docs as opposed to random articles, and can serve as a reminder / quick link, reducing the work for the mentor.

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Minimal solution for approval
+
+```ruby
+
+# version 2+
```
SleeplessByte: I would suggest removing `# version 2+`, the version should be bound either by the `Gemfile`, a `.ruby-version` or whatever -- I don'think you want to deal with n versions of solutions in the end in the mentor notes, which is what this implies to me.

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
...
+
+class HighScores
+  def initialize(scores)
+    @scores = scores
+  end
+
+  def scores  # no need to change getter method to `attr_reader`
```
SleeplessByte: Should this comment really be in the _solution_? IMO it's better as a block of text below this with an explanation _why_ it should not need to change to `attr_reader`. I think in this case you meant something along the lines of "it's fine to use a method instead of `attr_reader`, but it's not clear to me at all.

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
...
+
+  def latest
+    scores.last
+  end
+
+  def personal_top
+    scores.sort.reverse.take(3) 
+  end
+
+  # composing a string either with string interpolation or with an if.. else.. statement
```
SleeplessByte: Again I would use variations below this, and not inline this in the "minimal" solution.

tracks/ruby/exercises/high-scores/mentoring.md
SleeplessByte: I agree with @pgaspar , even though it's not tested for, a brittle solution is not a good guide for a mentor. If your example here is that "ok but TDD, this is barely enough", then why don't we accept shameless green solutions (solutions that basically have hardcoded solutions to the test cases)? **However**: I completely agree that solutions should _not_ be rejected if not covered by the instructions or test cases. Note here that IMO the instructions are part of the challenge and should not be disregarded. 

Example: A mentor on a different track mentored my `leap` exercise, suggesting I remove the parentheses, because their minimal solution did not have it in the guide. I had to explain that logically the solution to leap without parentheses is _not equivalent_.

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+### Reasonable approaches
+
+- See the one above
+- Defining `@scores`, `@personal_best` and `personal_top` in the constructor.  
+
+### General notes
+- The Instructions point to a beginner friendly explanation of instantiating a class: 
+http://ruby-for-beginners.rubymonstas.org/writing_classes/initializers.html
```
SleeplessByte: Please link, this does _not_ show up as a link on exercism.

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
...
+and to the rubydocs Array.
```
SleeplessByte: Actually link `Array` to rubydocs.

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
...
+### Reasonable approaches
+
+- See the one above
+- Defining `@scores`, `@personal_best` and `personal_top` in the constructor.  
+
+### General notes
+- The Instructions point to a beginner friendly explanation of instantiating a class: 
+http://ruby-for-beginners.rubymonstas.org/writing_classes/initializers.html
+and to the rubydocs Array. 
+- Most submissions use an getter method `scores`, not `attr_reader`.
```
SleeplessByte: I would make this part of "accepted variations".

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
...
+
+- See the one above
+- Defining `@scores`, `@personal_best` and `personal_top` in the constructor.  
+
+### General notes
+- The Instructions point to a beginner friendly explanation of instantiating a class: 
+http://ruby-for-beginners.rubymonstas.org/writing_classes/initializers.html
+and to the rubydocs Array. 
+- Most submissions use an getter method `scores`, not `attr_reader`. 
+That's okay at this point in the track, no need to suggest `attr_reader`.
```
SleeplessByte: I would make this part of "accepted variations".

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
...
+- Beginners create the attribute reader, but may call the instance variable instead of the getter. 
+This link gives a nice introduction of getters and setters, and the naming convention in Ruby:
+https://launchschool.com/books/oo_ruby/read/classes_and_objects_part1#accessormethods.
+It's recommended to link to this or your favorite explanation, instead of typing the solution for them.
+- `sort.reverse` is not very performant, but at this point in the track it's more important that students explore the
```
SleeplessByte: I would make this part of "accepted variations".

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
...
+That's okay at this point in the track, no need to suggest `attr_reader`.
+- Beginners create the attribute reader, but may call the instance variable instead of the getter. 
+This link gives a nice introduction of getters and setters, and the naming convention in Ruby:
+https://launchschool.com/books/oo_ruby/read/classes_and_objects_part1#accessormethods.
+It's recommended to link to this or your favorite explanation, instead of typing the solution for them.
+- `sort.reverse` is not very performant, but at this point in the track it's more important that students explore the 
+standard lib (and work with methods chaining) than that they know how to optimize.
+
+### Talking points
+- `Array#max` takes an argument; `scores.max(3)` works.
```
SleeplessByte: I would link this to the `max` function in the `Array` docs

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
...
+- Most submissions retrieve the `personal_top` values with the `Array#[]` method.
```
SleeplessByte: And therefore that's the right approach? I think this talking point is not clear on "why" it's a talking point.

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
...
+This link gives a nice introduction of getters and setters, and the naming convention in Ruby:
+https://launchschool.com/books/oo_ruby/read/classes_and_objects_part1#accessormethods.
+It's recommended to link to this or your favorite explanation, instead of typing the solution for them.
+- `sort.reverse` is not very performant, but at this point in the track it's more important that students explore the 
+standard lib (and work with methods chaining) than that they know how to optimize.
+
+### Talking points
+- `Array#max` takes an argument; `scores.max(3)` works.
+- Most submissions retrieve the `personal_top` values with the `Array#[]` method. 
+`Array#take` wins in readability. And `Array#max()` wins them all.
```
SleeplessByte: "wins them all" why? I think this talking point is not clear on "why" it's a talking point.

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Talking points
+- `Array#max` takes an argument; `scores.max(3)` works.
+- Most submissions retrieve the `personal_top` values with the `Array#[]` method. 
+`Array#take` wins in readability. And `Array#max()` wins them all. 
+- The `report` method asks for a conditionally composed string. It's meant to think about the trade-off 'duplication' vs 'readability'. Recommendation: approve any solution unless it's a supercomplicated `if/else statement`. 
+- Consider `String#squeeze` to avoid hard-to-spot spacing in the string itself. 
+
+### Mentor Research
+- `String#<< / String.concat` vs. `String#+`?
```
SleeplessByte: These should be links to the docs. And these notes are pretty meaningless if they don't include what the resolution of the research was.

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
...
+- The `report` method asks for a conditionally composed string. It's meant to think about the trade-off 'duplication' vs 'readability'. Recommendation: approve any solution unless it's a supercomplicated `if/else statement`. 
+- Consider `String#squeeze` to avoid hard-to-spot spacing in the string itself. 
+
+### Mentor Research
+- `String#<< / String.concat` vs. `String#+`? 
+Concatenation mutates the string instance in-place, String#+, creates new string objects.
+Ruby 2.3 will freeze string literals by default.
+
+### Changelog
+- Version 2 changed method names "highest" -> "personal_best", "top" -> "personal_top"
```
SleeplessByte: This is GREAT. I would love if we had this everywhere.

tracks/ruby/exercises/high-scores/mentoring.md
F3PiX: There's two questions in here:
1) Do we want brittle/non-idiomatic solutions to be approved, just because they pass the tests?
2) Is the 'everything in the initializer' an approvable approach?

1) No. 
2) I think not, I wouldn't do it myself, but why? If the _only_ argument against it is that it will break _if_ we'd add new _functionality_, I don't find that very convincing. 
If we think it's an approach that shouldn't be approved, or should be approved with food for thought, I'd like to add the arguments here that can help new mentors.
 
Four possible solutions in this matter, in the order of my liking:
1) Add a test that _proves_ this to be a brittle/bad solution.
2) An argument other than 'it will break if we add new functionality'. 
3) Add some explanation in the Mentor Notes here, that we can not _not_ approve this, due to the limitations of the exercise, so do `approve with food for thought` , including a recommendation to extract methods.
4) Add text to the instructions that implies that putting everything in the initializer is not an acceptable solution here.  

What do you all think?

tracks/ruby/exercises/high-scores/mentoring.md
F3PiX: Hmmm. I feel strongly that this is not the right place for links. I'll add it to the evaluation for this experiment.

tracks/ruby/exercises/high-scores/mentoring.md
F3PiX: Actually, I meant the version of the exercise. But as it's confusing, obviously, and there's a changelog, I'll remove it.

tracks/ruby/exercises/high-scores/mentoring.md
SleeplessByte: I meant it in a similar way like this: https://github.com/exercism/website-copy/pull/605/files#diff-832d9576c0235e6c66cf08e2eb0e0bdfR2, but perhaps you are right 🗡

tracks/ruby/exercises/high-scores/mentoring.md
SleeplessByte: Hi Maud,

I think you're asking the right questions!! (sidenote: in GFMD, you have to add an extra newline if you want a new list, now it show's up like 1 2 3 4. Typing 1. 1. 1. 1. would have resulted in the same).

I actually think that the FOUR solutions you provide should _all_ / or at least the first three be implemented

> Add a test that proves this to be a brittle/bad solution.
If this is a possibility then yes. However, forcing `high scores` to be immutable, would also fix this issue. Testing that could be:
- init
- modify scores array outside of the class
- check if the outcome reflects the initial values

> An argument other than 'it will break if we add new functionality'.
My argument is the _Example_ I listed above. A "bad" minimal solution will have mentors use that as baseline.

> Is the 'everything in the initializer' an approvable approach?
Idiomatic ruby likes shallow but narrow classes, and plenty of methods, so I think you're right that it's not; however, calling methods to calculate the outcomes and memoize e.g. in instance variables from the initializer IMO would be perfectly fine.

tracks/ruby/exercises/high-scores/mentoring.md
F3PiX: Aha, yes, links like to the Modulo arithmetic are marvellous!

tracks/ruby/exercises/high-scores/mentoring.md
pgaspar: Hey Maud,

> I think not, I wouldn't do it myself, but why? If the only argument against it is that it will break if we'd add new functionality, I don't find that very convincing.
If we think it's an approach that shouldn't be approved, or should be approved with food for thought, I'd like to add the arguments here that can help new mentors.

My point is that we probably shouldn't even mention it.

Let's think of a different case: in the Minimal solution for approval we have in these notes replace `@scores` with `$scores`. It's totally valid Ruby and will pass all our tests. However, as people experienced with Ruby will notice this is not a Reasonable approach at all - yes it works for this specific case with the current tests but it will lead to problems as soon as you continue expanding the class or using it in slightly different ways, not described in the tests (just as the all in initializer approach does). However, probably most if not all attentive mentors would comment on the fact that `$scores` is a global variable and how that's a bad practice **in certain situations** not described in the specs (just as the all in initializer approach is).

Here's how it would break:

```ruby
a = HighScores.new([1, 2, 3])
b = HighScores.new([4, 5, 6])

p a.scores
# => [4,5,6]
p b.scores
# => [4,5,6]
```

Do I think we should add a note to these mentor notes saying that global variables are bad? No. Should we add a spec showing how this is a bad practice? Also no - this is just one of a long list of bad practices we would need to test against. I expect the mentors to point this out themselves. And if we added a note and/or a spec we'd have to add it to pretty much all exercises, because this is an issue in pretty much all of them.

The same thing happens for the all in initializer approach - I expect mentors to naturally comment on it, but I don't think we should call it a "reasonable" solution, nor do I think that we should add a note or a test in every exercise where using this approach passes the tests (there are a bunch of them!).

We may be running in circles a bit here. Maybe a comment from @kotp or @Insti could help? Thanks for your comments too @SleeplessByte :+1:

F3PiX: > My point is that we probably shouldn't even mention it. <- Pedro

That makes perfect sense, actually! 

Let's go with that. 👍

tracks/ruby/exercises/high-scores/mentoring.md
kotp: I have not checked all submissions, and "most" could be an incorrect statement at any point in time.  Would "Many" be more suitable here?  Or "Common"?  Or even "Some" which acknowledges that it is done, without a moving target.

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
...
+http://ruby-for-beginners.rubymonstas.org/writing_classes/initializers.html
+and to the rubydocs Array. 
+- Most submissions use an getter method `scores`, not `attr_reader`. 
+That's okay at this point in the track, no need to suggest `attr_reader`.
+- Beginners create the attribute reader, but may call the instance variable instead of the getter. 
+This link gives a nice introduction of getters and setters, and the naming convention in Ruby:
+https://launchschool.com/books/oo_ruby/read/classes_and_objects_part1#accessormethods.
+It's recommended to link to this or your favorite explanation, instead of typing the solution for them.
+- `sort.reverse` is not very performant, but at this point in the track it's more important that students explore the 
+standard lib (and work with methods chaining) than that they know how to optimize.
```
kotp: "standard lib" should be "core" and "library".  No need to abbreviate a word, and especially not one that comes up with definitions that may not be clear in the English dictionary.  Since this is Ruby, there is a difference between "core" and "standard" and often nothing more than core is necessary.  Though I agree it is important that students explore both collections.

tracks/ruby/exercises/high-scores/mentoring.md
kotp: Unless the links are dynamic to the version that is promoted with this language, I would avoid the links.  _It is important that students know where and how to access the documentation_.   By this exercise, I would say it should not be something to need to link to. And general references, rather than links, will reduce the areas that we need to update because of version changes.

tracks/ruby/exercises/high-scores/mentoring.md
kotp: "most" is quantitative and may change at any point in time.   I would encourage why, without mentioning quantity of submissions, which is a moving target.

tracks/ruby/exercises/high-scores/mentoring.md
kotp: I don't agree about the links to the documentation.  For example, I know that I can do `ri 'String#+'` to get to the documentation, and a link is not going to change this.  Others may not.  Plus the links may decay over time, due to our supported language, or their dynamic nature depending on where the links direct.  I am also aware of students using different versions of Ruby, and so they would have to adjust for their own use anyway.

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -0,0 +1,71 @@
...
+
+  def personal_top
+    scores.sort.reverse.take(3) 
+  end
+
+  # composing a string either with string interpolation or with an if.. else.. statement
+  def report
+    difference = "#{personal_best - latest} short of" unless personal_best == latest
+
+    "Your latest score was #{latest}. That's #{difference} your personal best!".squeeze
```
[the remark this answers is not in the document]
kotp: Never mind, using squeeze corrects for that. :blush:

tracks/ruby/exercises/high-scores/mentoring.md
SleeplessByte: When I do `ri String#+` it gives me

```
$ ri String#+
Nothing known about String
```,

[String#+](https://ruby-doc.org/core-2.5.0/String.html#method-i-2B) however works, regardless the version. In general ruby doesn't _remove_ methods, _ever_, so usually links to older versions are not a problem.

Might just be me though :)

File: tracks/ruby/exercises/high-scores/mentoring.md (hunk 18 above)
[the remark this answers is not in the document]
Comment (SleeplessByte) [reply tracks/ruby/exercises/high-scores/mentoring.md]: Should use `squeeze(' ')`, because `"squeeze".squeeze` returns `squeze`.

File: tracks/ruby/exercises/high-scores/mentoring.md (hunk 12 above)
Comment (SleeplessByte) [reply tracks/ruby/exercises/high-scores/mentoring.md]: I agree with your reasoning.

File: tracks/ruby/exercises/high-scores/mentoring.md (hunk 9 above)
Comment (SleeplessByte) [reply tracks/ruby/exercises/high-scores/mentoring.md]: I think _some_ would be more appropiate. Removing the quantifier altogether would be best I think:

> There are submissions that use a getter method `scores`, instead of `attr_reader :scores`.

That way it doesn't matter what happens in the future!

File: tracks/ruby/exercises/high-scores/mentoring.md (hunk 12 above)
Comment (SleeplessByte) [reply tracks/ruby/exercises/high-scores/mentoring.md]: FWIW:

[https://ruby-doc.org/core/Array.html#method-i-max](https://ruby-doc.org/core/Array.html#method-i-max) always links to the latest _stable_ Ruby core docs.

File: tracks/ruby/exercises/high-scores/mentoring.md (hunk 15 above)
Comment (kotp) [reply tracks/ruby/exercises/high-scores/mentoring.md]: You probably haven't generated the ri documentation for core.  But your gems will show.  Once you generate the core documentation, it will show, if the documentation for gems are showing.

**Comment by rpottsoh** [comment]
Is there anything keeping this from being merged?

**Comment by kotp** [comment]
Strictly speaking not as far as I can see. Personally would prefer not to abbreviate words because of the multiple reasons that abbreviations end up being a little harder for non-native speakers.  The words being used in that way are likely known from a "I am programmer" land though.

A little bit stronger objection would be "standard library" vs "core library" since for Ruby there is a difference, and core is more the focus than the standard library.  ([see this review comment](https://github.com/exercism/website-copy/pull/633#pullrequestreview-181547125)).

File: tracks/ruby/exercises/high-scores/mentoring.md (hunk 17 above)
Comment (F3PiX) [reply tracks/ruby/exercises/high-scores/mentoring.md]: I use 'Core' and 'Standard' when referring to specific Ruby libraries. With lower case 'standard' I meant it in a more general way.

**Comment by F3PiX** [comment]
I mainly fixed the errors. There's also some matters of personal taste; feel free to open a new PR to discuss details.

@rpottsoh Sorry for the delay, this is ready to merge now.

**Comment by rpottsoh** [comment]
Thanks @F3PiX, there is no rush.  It just was not clear to me what the hold up was since there were two approvals.  I'm happy to merge this if you like, but I tend not to merge if the issuer has write access (or an excessive amount of time has transpired and the PR appears to be OK to merge).

**Comment by F3PiX** [comment]
Could you approve it @rpottsoh ? Then I can merge. It says Review required - probably for the last commit.

**Comment by rpottsoh** [comment]
You got it!  :smile:

**Comment by rpottsoh** [comment]
> probably for the last commit.

That commit appears to have dismissed the two approvals as stale.

File: tracks/ruby/exercises/high-scores/mentoring.md [hunk 19]
```diff
@@ -27,44 +25,38 @@ class HighScores
...
-- Most submissions use an getter method `scores`, not `attr_reader`. 
-That's okay at this point in the track, no need to suggest `attr_reader`.
 - Beginners create the attribute reader, but may call the instance variable instead of the getter. 
 This link gives a nice introduction of getters and setters, and the naming convention in Ruby:
-https://launchschool.com/books/oo_ruby/read/classes_and_objects_part1#accessormethods.
+[launchschool: accessor_methods](https://launchschool.com/books/oo_ruby/read/classes_and_objects_part1#accessormethods).
 It's recommended to link to this or your favorite explanation, instead of typing the solution for them.
 - `sort.reverse` is not very performant, but at this point in the track it's more important that students explore the 
-standard lib (and work with methods chaining) than that they know how to optimize.
+standard libraries (and work with method chaining) than that they know how to optimize.
```
kotp: _core library_ rather than the standard library.  The standard library is probably not where we want them to go.

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -27,44 +25,38 @@ class HighScores
...
 
 ### Talking points
-- `Array#max` takes an argument; `scores.max(3)` works.
-- Most submissions retrieve the `personal_top` values with the `Array#[]` method. 
-`Array#take` wins in readability. And `Array#max()` wins them all. 
-- The `report` method asks for a conditionally composed string. It's meant to think about the trade-off 'duplication' vs 'readability'. Recommendation: approve any solution unless it's a supercomplicated `if/else statement`. 
-- Consider `String#squeeze` to avoid hard-to-spot spacing in the string itself. 
+- [`Array#max`](https://ruby-doc.org/core/Array.html#method-i-max) takes an argument; `scores.max(3)` works.
+- Most submissions retrieve the `personal_top` values with the `Array#[]` method, which is harder to read than the convenience methods that Ruby offers: `Array#take(3)` or `Array#first(3)` win in readability. And `Array#max(3)` doesn't even need the sorting-and-reverting steps.
```
kotp: I would remove "most" in preference to "some".  No need to tempt students in a quantity argument here.  Many or Some, but as soon as you say "most" it may become untruthful at a point when we aren't counting.

tracks/ruby/exercises/high-scores/mentoring.md
iHiD: @kotp See this comment here in case you missed it: https://github.com/exercism/website-copy/pull/633#discussion_r240696982

I think "general" is a better word when not refering to "core" or "standard" specifically.

tracks/ruby/exercises/high-scores/mentoring.md
kotp: Easy to [stroop](https://faculty.washington.edu/chudler/words.html) though.

tracks/ruby/exercises/high-scores/mentoring.md
kotp: I agree, it removes that stroop effect completely.  Still better to get the changes in, I think, than be stopped by that.  Not a strong objection, I just see it causing (me) cognitive issues. :stuck_out_tongue_winking_eye:

PR #631, opened 2018-11-29
Add Leo Benkel bio

leobenkel:

PR #634, opened 2018-11-30
remove wrong bio entry for Scala

rpottsoh:
Persons name starts with letter L and not B.  This was accidentally missed during review.


rpottsoh:
If you are able to merge then please do so when approved.

rpottsoh:
reversing #631.

PR #640, opened 2018-12-02
csharp/leap: Improve mentor notes

SleeplessByte:
Add suggestions and talking points around the order of operation and evaluation. The old reasonable solution is logically incorrect!

### DIFF
--- a/tracks/csharp/exercises/leap/mentoring.md
+++ b/tracks/csharp/exercises/leap/mentoring.md
@@ -7,7 +7,7 @@ public static class Leap
 {
     public static bool IsLeapYear(int year)
     {
-        return year % 4 == 0 && year % 100 != 0 || year % 400 == 0;
+        return year % 4 == 0 && (year % 100 != 0 || year % 400 == 0);
     }
 }
 ```
@@ -20,7 +20,7 @@ using System;
 public static class Leap
 {
     public static bool IsLeapYear(int year)
-        => year % 4 == 0 && year % 100 != 0 || year % 400 == 0;
+        => year % 4 == 0 && (year % 100 != 0 || year % 400 == 0);
 }
 ```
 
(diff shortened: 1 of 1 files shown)


tracks/csharp/exercises/leap/mentoring.md
```diff
@@ -20,17 +20,33 @@ using System;
...
 }
 ```
 
 ### Common suggestions
 
 - Instead of using several, potentially nested if-statement, suggest to remove the if-statements by returning a single expression using the boolean `&&` and `||` operators.
 - Instead of explicitly returning `true` or `false` (e.g. `if (x) return true else return false`), suggest to simply return the boolean condition (`return x`).
-- If students are doing more than three checks in their if-statement, ask if they could get it down to just three checks. You could give them a hint to following the instructions in the README.
+- If students are doing more than three checks in their if-statement, ask if they could get it down to just three checks. You could give them a hint to following the instructions in the README. There are only two cases that return true:
+  - a year MUST be a multiple of four AND NOT be a multiple of 100
```
ErikSchierboom: I like that you make this explicit.

tracks/csharp/exercises/leap/mentoring.md
```diff
@@ -20,17 +20,33 @@ using System;
...
+  - a year MUST be a multiple of 400 (implies multiple of 4 and 100)
+- The _order of operation_ and _evaluation_ matter, see talking points:
```
ErikSchierboom: Should this really be a common suggestion? To me, optimizing which checks are executed is nice, but not required at this point in the track (leap being the second exercise). Maybe this section could be moved as a whole to the talking points section?

@robkeim @jpreese what do you think?

tracks/csharp/exercises/leap/mentoring.md
```diff
@@ -20,17 +20,33 @@ using System;
...
+  - A student might be missing parentheses:
+    ```csharp
+    return year % 4 == 0 && year % 100 != 0 || year % 400 == 0;
+    ```
+    Albeit correct in output, the check `year % 400` is executed for all years that don't match `year % 4 == 0`, and for these (75% of all years), always returns `false`, because if something is not divisible by 4 it _can not_ be divisible by 400. This is making a redundant check.
+  - A student might have inversed the test order:
+    ```csharp
+    return year % 400 == 0 || (year % 4 == 0 && year % 100 != 0);
+    ```
+    Albeit correct in output, the check `year % 400` is _only_ true for `1.03%` years that are divisible by 4. Always try to put the common path first. Logically, following the instructions putting this check first is **incorrect**.
```
ErikSchierboom: Why is inversing the test order logically incorrect? It is still the same logic, albeit with redundant checks.

SleeplessByte:
This is matching the other tracks where exercises are being disapproved if
they don't match the order of the instructions due to the redundancy of
tests.

I've been mentored here that I could leave out parentheses which I didn't
like as it is logically not the same solution; other students following
multiple tracks might actually get confused if one track enforces this (JS
and Ruby do) and some tracks don't (csharp and TypeScript at the moment,
PRs are open).

Personally I think the exercise was chosen originally for the ruby track
exactly because it's simple to implement but has a lot of _haken en ogen_.

I can't use GitHub formatting in email response, but the second comment
about why it's not logically correct is because it's not:

The instructions are as follows +paraphrased+

A year is a leap year if it is divisible by 4 AND
- it is not divisible by 100
- OR it is divisible by 400

The inverse test or lacking parentheses yield different "resulting
instructions". This sets up the student to read the instructions carefully.
None of the tests of none of the languages test for implementation details
(yay) so this is meant to set them up to receive comments about skimming
instructions instead of carefully implementing them.

SleeplessByte:
This is matching the other tracks where exercises are being disapproved if
they don't match the order of the instructions due to the redundancy of
tests.


I've been mentored here that I could leave out parentheses which I didn't
like as it is logically not the same solution; other students following
multiple tracks might actually get confused if one track enforces this (JS
and Ruby do) and some tracks don't (csharp and TypeScript at the moment,
PRs are open).


Personally I think the exercise was chosen originally for the ruby track
exactly because it's simple to implement but has a lot of _haken en ogen_.


I can't use GitHub formatting in email response, but the second comment
about why it's not logically correct is because it's not:

The instructions are as follows +paraphrased+

A year is a leap year if it is divisible by 4 AND
- it is not divisible by 100
- OR it is divisible by 400

The inverse test or lacking parentheses yield different "resulting
instructions". This sets up the student to read the instructions carefully.
None of the tests of none of the languages test for implementation details
(yay) so this is meant to set them up to receive comments about skimming
instructions instead of carefully implementing them.

tracks/csharp/exercises/leap/mentoring.md
SleeplessByte: Substitute this simple exercise with very expensive function calls and suddenly we care. Teach the student to care now so they don't mess up later.

I think that was the original thought behind the note in the ruby and later JavaScript tracks.

robkeim: I think whether or not this should be discussed depends on the student's previous experience.

For someone that is not new to programming but only new to the language it's a good thing to discuss when reviewing the solution.

However, for someone that is just getting started with programming I think that we need to be careful not to overwhelm them. Agreed that order of execution is an important thing to consider, but if you've just completed HelloWorld and this is your second exercise it's not the most important thing for you to be focusing on at that stage IMO.

SleeplessByte:
I definitely don't disagree with that. I think two-fer could be between
this and hello world.

The issue still is that the experience is different between tracks in a,
imo, destructive way.

tracks/csharp/exercises/leap/mentoring.md
SleeplessByte: I definitely don't disagree with that. I think two-fer could be between this and hello world.
The issue still is that the experience is different between tracks in a, imo, destructive way.

Would it work if I made these their own topic, but not "forced suggestions"? Instead something the mentor can mention when they approve a solution _without_ parentheses or "wrong" order? I understand that not all tracks need be the same, even though it feels wrong to me.

ErikSchierboom: > For someone that is not new to programming but only new to the language it's a good thing to discuss when reviewing the solution.

Well, this is the target audience of Exercism: people that already know how to program but want to become fluent in a language.

SleeplessByte:
Thanks! 

FWIW I am happy to restructure this after a while if some things prove not to be as effective as I hope them to be.

ErikSchierboom:
> FWIW I am happy to restructure this after a while if some things prove not to be as effective as I hope them to be.

Let's keep it pragmatic and do just that! Thanks for the PR!

PR #641, opened 2018-12-02
ruby/two-fer: Improve mentor notes

SleeplessByte:
After having an interaction with a mentor on a solution without any good explanation why the interaction was what it was, I've sifted through other solutions and mentors comments. These mentor-note changes reflect my findings and what I believe to be better in terms of the exercism learning experience.

### Group style suggestions 
Groups the suggestions about style together, so it's clear which suggestions are functional and which are not. Add links to the specific community style guide rules.

### Remove explicit module notion and make it a first class citizen 
At this point in the track it's fine to use ANYTHING, but the notion that "class" is somehow more correct than module is off-putting.

A class should only be used if its functionality requires instantation. Utility functions and the sorts should either be on a named constant which is an Object, or a module so it can be included/extended and shared in other modules and classes.

### Add more reasonable solutions including string templates
The point of this exercise is to prepare a student for the track, not force one usage over another.

String interpolation is simpler, but template strings are much more maintainable when used with named arguments as it can be subtituted easily for internationalisation and/or extended.

This adds all the variations as reasonable so mentors no longer penalise students for using it.

### DIFF
--- a/tracks/ruby/exercises/two-fer/mentoring.md
+++ b/tracks/ruby/exercises/two-fer/mentoring.md
@@ -34,9 +34,9 @@ end
 ```
 
 ### Common suggestions
-- Suggest using a default value instead of any form of conditionals. 
-- People often set the default of `name` to `nil`. Ask if there is something they could do with the default value to avoid the conditional and make their code simpler.
-- Some people need help with running the tests. 
+- Use a default value instead of any form of conditionals.
+- When a student set the default of `name` to `nil`: Ask if there is something they could do with the default value to avoid the conditional and make their code simpler.
+- Some people need help with running the tests.
 - Style suggestions based on community conventions:
   - Remove explicit `return` ([no-explicit-return](https://github.com/rubocop-hq/ruby-style-guide#no-explicit-return))
   - Use `def` with parentheses, because of the [parameter (method-parens)](https://github.com/rubocop-hq/ruby-style-guide#method-parens).
--- a/tracks/ruby/exercises/two-fer/mentoring.md
+++ b/tracks/ruby/exercises/two-fer/mentoring.md
@@ -13,19 +13,17 @@ end
 ### Common suggestions
 - Suggest using a default value instead of any form of conditionals. 
 - People often set the default of `name` to `nil`. Ask if there is something they could do with the default value to avoid the conditional and make their code simpler.
-- Suggest to remove `return`
-- Use `def` with parentheses, because of the parameter.
 - Check if the BookKeeper module is present and the comment removed.
 - Some people need help with running the tests. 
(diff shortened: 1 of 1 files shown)


tracks/ruby/exercises/two-fer/mentoring.md
```diff
@@ -1,34 +1,53 @@
...
-- Style preferences. This exercise is a good opportunity to talk about style conventions like indentation, parameter parenthesis in method declarations and removing redundant comments. 
-https://github.com/rubocop-hq/ruby-style-guide
-Given the place in the curriculum, it may be worth to not going to deep, and to not addressing points that are controversial or personal preference. 
+- String interpolation (http://ruby-for-beginners.rubymonstas.org/bonus/string_interpolation.html), but a student may also use string templates via `format` or `%`. You may mention the alternative technique (templates when interpolation is used, and vice versa), always prefer either over string concatination.
+- Style preferences based on [community style guide via rubocop](https://github.com/rubocop-hq/ruby-style-guide): This exercise is a good opportunity to talk about style conventions like indentation, parameter parenthesis in method declarations and removing redundant comments. Given the place in the curriculum, it may be worth to not going to deep, and to not addressing points that are controversial or personal preference.
```
kotp: "may be worth not going too deep"  Remove the first "to" and correct the second *to* spelling to _too_.

SleeplessByte: Copied from a different note; will PR the fix there too!

iHiD:
As the three approaches could all be mixed up, I'm not sure separating them into three makes sense.
I wonder if rather than having 3 solutions, it would be clearer to have one solution with comments. e.g.

```ruby
# May use class instead of module
module TwoFer

  # May use module_function instead of self.
  def self.two_fer(name = 'you')

    # Or "format('One for %<name>s, one for me.', name: name)'"
    # Or "One for %s, one for me.' % name"
    "One for #{name}, one for me."
  end
end
```

SleeplessByte:
@iHiD I personally really don't like these types of comments in solutions, as they tend to clutter up, it's not what comments are meant for and only works okay for `two-fer`, but not for more complex exercises. In this case the comments are also missing the valid `metaclass` and `sprintf`-almost-but-not-quite-alias.

In general in the JS and TS tracks I've started writing "variations include" without examples. Other mentors requested me to write out those examples as they felt that to be more clear. If this were up to me I'd only put down one solutions and a list of "acceptable variations" and leave it up to mentor to fill in the blanks.

Ideally we'd have one of those "samples" boxes like this:
image

But then with multiple approaches as tabs or something like that :P

rpottsoh:
@kotp is this OK to merge?

PR #642, opened 2018-12-02
typescript/leap: Add mentor notes

SleeplessByte:
Adds mentor notes based on the JavaScript notes, but greatly extended for typescript:
- `tslint` mentions,
- links to typescriptlang.org's handbook,
- TypeScript specifics such as typing functions, type inference and boxed types.


sshine:
Very nice!

I think the last example has "by10" when "by100" was meant.

PR #646, opened 2018-12-03
Rakefile: Add "github_username" tests

### LINKED ISSUE #636: propose new test for Rakefile: First letter of name matches filename.
It is easy to miss this check when manually checking mentor bio JSON updates.  Sometimes people grab the wrong file and sometimes those reviewing forget to check that the first letter of their name agrees with the name of the JSON file.

I would like to see a test added to Rakefile that will confirm that first letter and filename agree.

rpottsoh:
-  Verify that "github_username" is not null
-  Verify that "github_username" starts with same letter as filename

fixes #636

My first time tinkering with a Rakefile...  Let me have it!

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -20,13 +20,18 @@ task :test do
         username = mentor["github_username"]
         name = mentor["name"]
 
-        if name == "" || name.nil?
-          errors << "Name can't be blank for %s in %s." % [username, file]
+        if username == "" || username.nil?
+          errors << "Github username can't be blank for %s." % [file]
           next
         end
 
         unless username.downcase.start_with?(filename)
           errors << "First letter of github_username \"%s\" must equal JSON filename %s" % [username, file]
+          next
+        end
+
+        if name == "" || name.nil?
+          errors << "Name can't be blank for %s in %s." % [username, file]
           next
         end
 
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -23,6 +25,11 @@ task :test do
           next
         end
 
+        if username[0].upcase != filename.upcase
```
iHiD: Maybe `unless username[0].downcase == filename`

iHiD:
This looks good to me. Assigning to @kytrinyx :)

Rakefile
rpottsoh: will `filename` always be lower case?  If that is a certainty then I am fine with this suggestion.  Otherwise wouldn't it be safer to force both sides to be the same case?

rpottsoh:
I plan to work on this more during my evening.

rpottsoh:
> String#start_with? for it to read better :) Subsequently, since we are now sending username a message, maybe we should check if it exists, too.

I am planning to work this in too.

Rakefile
rpottsoh:
```suggestion
        unless username.start_with?(filename, filename.upcase)
```

rpottsoh:
Thanks for the suggestions @kntsoriano and @iHiD.  Let me know if there are any other changes you think I should make.

iHiD:
>  pretty soon...

This could be days or months. Definitely worth doing.

iHiD:
I'm going to merge this - thanks for doing it :)

PR #648, opened 2018-12-04
Update mentoring for swift two-fer

onevcat:


NobbZ:
@exercism/swift could you please review? I'm not proficient enough in swift to make any decissions on this, but shadowing names seems as if we want to avoid it.

sendtobo:
I agree with @masters3d . Since we are just asking for a function its not necessary to create the overhead of a class to store state or initialize. A single function should be the encouraged solution 👍

PR #653, opened 2018-12-06
C/Hamming mentor notes

elyashiv:


elyashiv:
Good point on changing the first solution. I just copied it from the example... 

Right now I don't have more common comments on this exercise, but if I will have more, I will open another PR.

sshine:
Excellent job, thanks!

PR #654, opened 2018-12-07
haskell/nucleotide-count: Add mentoring notes

sshine:


sshine:
Very nice, @frerich. Thanks! Since exercism/haskell#714 we're using a custom data type

    data Nucleotide = A | C | G | T deriving (Eq, Ord, Show)

I'll try and adapt your solution to this data type and add this.

frerich:
I imagine replacing `"ACGT"` with `[A,C,G,T]` (or maybe deriving `Enum` and then using `[A .. T]` if you fancy) ought to be enough. :-)

sshine:
@frerich: I like the use of `M.fromListWith (+)`, so I merged it with my previous favorite solution and added the use of `readEither` and `<>` for error conversion, so it now relies entirely on the `Read` instance (and there are no separate validation functions that can desync from the data type). I also added `Enum` and `Bounded` so that the zero map can be generated as

    M.fromListWith (+) [minBound..] [0,0..]

sshine:
@frerich: Could you re-approve my changes?

tracks/haskell/exercises/nucleotide-count/mentoring.md
```diff
@@ -0,0 +1,163 @@
...
+  where count c = length . filter (== c) $ s
+
+isNucleotide :: Char -> Bool
+isNucleotide = (`elem` "ACGT")
+```
+
+This solution also handles errors in an initial guard.
+
+Because GHC can fuse list combinators, four separate `length . filter (== c)`
+do not necessarily perform bad. Zeroes are not a special case here.
```
frerich: This sentence seems a little fishy to me -- four separate `length . filter (== c)` calls "do not necessarily" perform bad?

It's true that GHC can fuse list combinators, but to the best of my knowledge this only works if you compose certain functions (at least catamorphisms). I'm not sure this claim is accurate when applied to four independent invocations of `count`.

sshine: I'll have to run a benchmark then.

Sounds like you're right.

I recall this one not running nearly like four times for large input.

rpottsoh:
Is this ready to be merged?  The approvals to merge do not read like actual approvals....

sshine:
This was still in my queue of things to address once I got home from vacation (yesterday), but you've only bumped it up the queue, since the mentor notes are now somehow incorrect. :-D

I'll fix this Monday.

PR #637, opened 2018-12-01
Haskell: Leap mentoring notes

a59:
Adding new Mentoring notes for Haskell leap.

PR #655, opened 2018-12-08
add a59 mentor profile. add mentoring notes for Haskell Leap.

a59:


mentors/haskell/a.json
```diff
@@ -23,4 +23,12 @@
     "avatar_url": null,
     "bio": "A Haskell fꜷn."
   }
```
rpottsoh:
```suggestion
  },
```

rpottsoh: You need to add a comma here....

sshine:
Yes, one subject per PR, please. :)

a59:
I am sorry for the stupid question, but how do I address the requested changes?

sshine:
Since your feature branch of #637 is still available, I suggest you do the following:

1. Create a new feature branch with `git checkout -b leap-mentoring`, copy `leap/mentoring.md` to that branch, commit it, push it to your GitHub copy of the repo, and open a pull request with a title like "Haskell: Leap mentoring notes"

2. Remove `leap/mentoring.md` from this branch, commit the removal and push it again. It doesn't matter that there's first a commit that adds it and later one that removes it, because we'll squash the commits together into one anyway, so anything related to `leap/mentoring.md` will disappear.

Now, I think maybe you've complicated this workflow a little by creating a pull request from your master branch. It means you can't create the feature branch based directly on master, since it contains stuff you don't want to be in the `leap-mentoring` feature branch.

You can do a few things to overcome that: For example, `git checkout` something other than `master`, and then `git checkout -b` from there. That "something other" could be "`upstream`":

    git remote add upstream https://github.com/exercism/website-copy
    git fetch upstream
    git checkout upstream/master
    git checkout -b leap-mentoring

Then you're in a branch based on the latest remote master, rather than on your local master, which is currently in use as another feature branch.

PR #674, opened 2018-12-23
add instruction for [ and ] chars in JSON bio files

rpottsoh:
There have been a few occasions where the first mentor to a JSON file has removed the square brackets that are intended to encapsulate the mentors that may be added to a particular JSON file.  The blurb I have added is intended to indicate that the square brackets should not be removed.

### DIFF
--- a/mentors/README.md
+++ b/mentors/README.md
@@ -49,6 +49,8 @@ The `link_url` and `link_text` are for linking to a personal site, if you want t
 Note that if you leave a field `null`, we will use the content from your public GitHub profile.
 If you want to leave a field blank, set it to an empty string.
 
+The file should consist of a JSON array of mentor bios.  The first line of the file should contain only a single `[` character and the last line should be only a single `]` character.  Each bio should be separated by a single comma.
+
 Please create a Pull Request in this repository, which we will review and merge.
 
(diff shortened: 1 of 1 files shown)


mentors/README.md
```diff
@@ -49,6 +49,8 @@ The `link_url` and `link_text` are for linking to a personal site, if you want t
 Note that if you leave a field `null`, we will use the content from your public GitHub profile.
 If you want to leave a field blank, set it to an empty string.
 
+The first line of the file should consist of only a single `[` character and the last line should consist of only a single `]` character.
+
```
kytrinyx: I think this is a good idea. How about specifying first that the file should contain a JSON array, and then follow that up with the clarification that you have? I think that extra context will make it a bit easier to understand.

rpottsoh: Thanks for the suggestion.  I agree.

kytrinyx: There's a small tyypo:

> bio's

This should be bios.

rpottsoh: Thanks!  Got it.  This is ready to be merged.

kytrinyx:
Thanks so much, @rpottsoh !

PR #681, opened 2019-01-03
Create Mentoring Copy for Difference of Squares (Swift)

sendtobo:


sendtobo:
Is there someone I should tag to get this reviewed?

tracks/swift/exercises/difference-of-squares/mentoring.md
```diff
@@ -0,0 +1,54 @@
+# Difference of Squares
+
+_Differences of Squares_ is a core exercise, unlocked after _Gigasecond_.
```
sshine: This section is not necessary.

tracks/swift/exercises/difference-of-squares/mentoring.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Reasonable Solutions
+
+```swift
+class Squares {
+    let maxNum: Int
+    
+    lazy var squareOfSum: Int = {
+        let sum = (0...self.maxNum).reduce(0, { $0 + $1 })
```
sshine: I don't know Swift, so I take it there's no standard library `sum` method, since the "next level" solution defines it via `reduce`.

Is there a difference between generating a list and an iterable range between 0 and `self.maxNum`, and if so, do students sometimes pick the less optimal?

tracks/swift/exercises/difference-of-squares/mentoring.md
```diff
@@ -0,0 +1,54 @@
...
+    
+    lazy var sumOfSquares: Int = {
+        return (0...self.maxNum).reduce(0, { $0 + $1.squared })
+    }()
+    
+    lazy var differenceOfSquares: Int = {
+        return self.squareOfSum - self.sumOfSquares
+    }()
+    
+    init(_ maxNum: Int) {
```
sshine: It looks like the exercise requires the creation of an object that stores `maxNum`; I've seen similar use of OO in the Perl track.

So without modifying the exercise, this appears to be the right thing to do. But I don't think that makes it "reasonable".

Have you considered changing the exercise so it doesn't wrap integers in classes just to perform method calls on them?

(This comment is not a request to change anything in this PR.)

tracks/swift/exercises/difference-of-squares/mentoring.md
```diff
@@ -0,0 +1,54 @@
...
+}
+```
+
+## Common suggestions
+
+- Sometimes students will try to do all of the computations within the init method. This is not optimal when numbers get very large; the cost of computation starts to rise. So the use of lazy variables is strongly encouraged so that computation is only done if needed.
+- Encourage the use of reduce here. It's a perfect place to understand exactly what this function does and how powerful it can be.
+- In addition it is a great place to teach students about the power of ranges. You can assemble a range in a variety of ways, however the syntactial sugar of `...` is the simpliest and most straight forward.
+
+## Next Level Additions
```
sshine: Have you considered the format of mentor notes described [here](https://github.com/exercism/website-copy/blob/master/README.md#mentor-notes)?

Reasonable solutions, common suggestions, talking points.

This looks like it could be another reasonable solution, since it's so complete. But you could have advanced alternatives as a talking point, too.

tracks/swift/exercises/difference-of-squares/mentoring.md
sendtobo: removed

tracks/swift/exercises/difference-of-squares/mentoring.md
sendtobo: There is no sum function within swift. The common practice is to do with with the reduce function.

Ranges and Lists are different types within swift, but each is essentially the same here because the methods used are defined on Collections which Ranges and Lists both are. I'll call that out here as well

tracks/swift/exercises/difference-of-squares/mentoring.md
sendtobo: Thanks for pointing me there, I think I cleaned it up now to the correct structure.

tracks/swift/exercises/difference-of-squares/mentoring.md
```diff
@@ -0,0 +1,50 @@
...
+}
+```
+
+### Common suggestions
+
+- Sometimes students will try to do all of the computations within the init method. This is not optimal when numbers get very large; the cost of computation starts to rise. So the use of lazy variables is strongly encouraged so that computation is only done if needed.
+- Encourage the use of reduce here. It's a perfect place to understand exactly what this function does and how powerful it can be.
+- In addition it is a great place to teach students about the power of ranges. You can assemble a range in a variety of ways, however the syntactial sugar of `...` is the simpliest and most straight forward.
+
+### Talking Points
```
sshine:
```suggestion
### Talking Points

```

tracks/swift/exercises/difference-of-squares/mentoring.md
sshine: Excellent.

tracks/swift/exercises/difference-of-squares/mentoring.md
sshine: Yes, but in https://github.com/exercism/swift

The reviewer there will be a track maintainer.

Tracks can have conventions of their own.

sendtobo:
👍🏼

PR #682, opened 2019-01-03
Add mentoring notes for leap in R

jonmcalder:


tracks/r/exercises/leap/mentoring.md
```diff
@@ -0,0 +1,30 @@
...
+leap <- function(year) {
+  year %% 4 == 0 & (year %% 100 != 0 | year %% 400 == 0)
+}
+```
+
+
+### Common suggestions
+- there are just two cases that return True:
+- a year is a multiple of 4 *and not* 100
+- a year is a multiple of 4, 100, and 400
```
tqa236: I think the second condition should be "a year is a multiple of 400". Clearly it will be a multiple of 4 and 100. If we express the second in the current way, we should check the divisibility for 4 and 100 (again) and then 400 to match the description, which is not the case here, even in the sample solution.

jonmcalder: @tqa236 I see what you mean - thanks for the suggestion. Could you open a new PR with that change?

PR #683, opened 2019-01-05
Update code_of_conduct.md

Jarak-Jakar:
Modify a couple of lines in the 'We condemn' section, in order to make the style consistent (those lines didn't follow the "we condemn..." style).

Compress two separate 'disrespect' lines into one, as it seems a bit pointless to have the two of them apart.

Capitalise the start of each entry in the 'The Simple Version' and 'The Details' lists, in order to make them consistent with the other lists on the page.

Write out Code of Conduct in full the first time it is used on the page with the contraction in parentheses following it.  I know this is arguably a bit pointless when the page is titled Code of Conduct, but it's considered best-practice style in pretty much every field and venue I have written for in the past.

### DIFF
--- a/pages/code_of_conduct.md
+++ b/pages/code_of_conduct.md
@@ -8,7 +8,7 @@ If you see or experience abuse, harassment, discrimination, or feel unsafe or up
 
 ### Enforcement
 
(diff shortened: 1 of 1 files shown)


Jarak-Jakar:
@rpottsoh : Thanks to you and the rest of the Exercism team for the absolutely excellent site and system :)  I'm just glad to have been actually able to help!

PR #691, opened 2019-01-08
Acronym mentoring: add another (better?) regexp

link2xt:


link2xt:
See https://github.com/exercism/problem-specifications/pull/1436

PR #694, opened 2019-01-13
ruby/series Add Mentor Notes

F3PiX:
New to the track: Mentor Notes for Series. 
To prepare for the upcoming promotion to a core exercise. https://github.com/exercism/ruby/pull/914
It's using the experimental format first introduced in [High Scores notes](https://github.com/exercism/website-copy/blob/master/tracks/ruby/exercises/high-scores/mentoring.md).

Very happy with comments that correct mistakes.
There's also an issue (https://github.com/exercism/ruby/issues/912) to discuss the Minimal Solution for Approval.

### DIFF
--- a/tracks/ruby/exercises/series/mentoring.md
+++ b/tracks/ruby/exercises/series/mentoring.md
@@ -1,4 +1,4 @@
-New concepts: Iterating over an array (preferably with `each_cons`, avoid `each_with_index`); `.chars` to split a string; `attr_reader`; `private` (for the attr_reader); raising an Argument Error; inline if-statement 
+New concepts: Iterating over an array (preferably with `each_cons`, avoid `each_with_index`); `.chars` to split a string; (private) `attr_reader`; raising an Argument Error; inline if-statement 
 
 ### Minimal solution for approval
 
@@ -32,7 +32,7 @@ and to the rubydocs [Enumerable](https://ruby-doc.org/core/Enumerable.html) with
 
 
 ### Talking points
-- `each_cons` instead of an iterator `with_index`: In Ruby, you rarely have to write iterators that need to keep track of the index. Enumerable has powerful methods that does that for us.
+- `each_cons` instead of an iterator `with_index`: In Ruby, you rarely have to write iterators that need to keep track of the index. Enumerable has powerful methods that do that for us.
 - `chars`: Use Ruby's built-in methods instead of implementing it yourself: `chars` instead of `split('')`   
 - `attr_reader`: instead of the instance variable (Explained here:)[https://ivoanjo.me/blog/2017/09/20/why-i-always-use-attr_reader-to-access-instance-variables]
 - `private` attr_reader: Following the 'rule' for encapsulation: if it doesn't need to be public, make it private. [This link](http://ruby-for-beginners.rubymonstas.org/writing_classes/state_and_behaviour.html) may come in handy for a first introduction. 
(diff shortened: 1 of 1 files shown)


tracks/ruby/exercises/series/mentoring.md
```diff
@@ -0,0 +1,49 @@
+New concepts: Iterating over an array (preferably with `each_cons`, avoid `each_with_index`); `.chars` to split a string; (private) `attr_reader`; raising an Argument Error; inline if-statement 
+
+### Minimal solution for approval
+
+```ruby
+
+class Series
+
+  def initialize(serie)
```
kotp: If we are working with a series and we ask for a serie, since they have the same meaning, why do we need the class?  It is self referential.  It says to me "Hey, I will build a series for you if you give me a series."  Instead, we want a collection of numbers as a string.  I think we can ask for that.

tracks/ruby/exercises/series/mentoring.md
```diff
@@ -0,0 +1,49 @@
...
+    @serie = serie
+  end
+
+  def slices(span)
+    raise ArgumentError if span > serie.size
```
kotp: Consider using `unless` and showing what "good" looks like for the conditional statement.

tracks/ruby/exercises/series/mentoring.md
```diff
@@ -0,0 +1,49 @@
...
+  ```
+
+### General 
+- The Instructions point to a beginner friendly [explanation of iterating in Ruby:](http://jeromedalbert.com/ruby-how-to-iterate-the-right-way/)
+and to the rubydocs [Enumerable](https://ruby-doc.org/core/Enumerable.html) with a hint that should make finding `each_cons` easier. 
+
+
+### Talking points
+- `each_cons` instead of an iterator `with_index`: In Ruby, you rarely have to write iterators that need to keep track of the index. Enumerable has powerful methods that do that for us.
+- `chars`: Use Ruby's built-in methods instead of implementing it yourself: `chars` instead of `split('')`
```
kotp: The comment about using the built in methods doesn't seem to apply here.  That seems to be an aside from what the lead is indicating,

`chars`: Use instead of `split('')`

tracks/ruby/exercises/series/mentoring.md
kotp: I don't know.  I only knew what was being asked for after reading the README for the exercise.  The name 'serie' was confusing me because of that self reference, so I had to reach for external documentation instead of the source code.  A 'collection of digits' is what is important to receive, the tests will encourage how that is implemented from there.

kotp: `digits.chars` maybe?  Yes, it is a method name, but  perhaps `numbers` would be better.  I think `numerals` would work, but generally these represent the symbols used, rather than numbers,  Then again, it is what we want, a collection of those symbols, to break out into the series, not necessarily treating the collection of the digits as a single number, and not at this point.

F3PiX:
@kotp I propose to continue the discussion on the naming after the PR is merged. If you can approve, can you also merge it please?

F3PiX:
Thanks a lot!

tracks/ruby/exercises/series/mentoring.md
```diff
@@ -0,0 +1,49 @@
...
+
+class Series
+
+  def initialize(numerals)
+    @numerals = numerals.chars
+  end
+
+  def slices(span)
+    raise ArgumentError unless span <= numerals.size
+    numeral.each_cons(span).map(|slice| slice.join )
```
kotp: I believe this should have an `s` after `numeral`

F3PiX: Thanks! I opened a new PR to fix it. #698

PR #697, opened 2019-01-15
Haskell/Pangram: Update mentoring notes

sshine:
Add more examples, common suggestions and talking points

### DIFF
--- a/tracks/haskell/exercises/pangram/mentoring.md
+++ b/tracks/haskell/exercises/pangram/mentoring.md
@@ -4,19 +4,64 @@
 import Data.Char (toLower)
 
 isPangram :: String -> Bool
-isPangram xs = all (`elem` map toLower xs) ['a'..'z']
+isPangram s = all (`elem` map toLower s) ['a'..'z']
 ```
+
+This solution traverses the input once for each letter, but terminates once each was found.
+
+It has the benefit of being declarative and to some extent short-circuiting.
 
 ```haskell
 import Data.Char (toLower)
 
 isPangram :: String -> Bool
-isPangram xs = null (['a'..'z'] \\ map toLower xs)
+isPangram s = null (['a'..'z'] \\ map toLower s)
 ```
 
+This solution traverses the input at most once and terminates once each is found.
+
+It has the benefit of being efficient, declarative and set-based (although not `Data.Set`-based).
+
+```haskell
+import Data.Char (toLower)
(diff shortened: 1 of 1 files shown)


tracks/haskell/exercises/pangram/mentoring.md
```diff
@@ -4,19 +4,64 @@
 import Data.Char (toLower)
 
 isPangram :: String -> Bool
-isPangram xs = all (`elem` map toLower xs) ['a'..'z']
+isPangram s = all (`elem` map toLower s) ['a'..'z']
 ```
 
+This solution traverses the input once for each letter, but terminates once each was found.
```
ErikSchierboom: You could consider creating separate headings for each of the solutions, which can help in quickly scanning the reasonable solutions. I've done this in the C# and F# tracks, see for example [this file](https://github.com/exercism/website-copy/blob/b78072b386b9792b3d43cf4292d8d1dd1b4b6712/tracks/csharp/exercises/bob/mentoring.md).

Note: this is merely a suggestion

ErikSchierboom: > Thanks, I'll do that! Perhaps this suggestion should be added here.

This might be a good idea. Maybe @iHiD has any thoughts on whether these sub-headings for different reasonable solutions is a good idea?

iHiD: Yeah - I'm good with that as long as they're `####` and the general headings are respected. A PR to the readme would be very welcome :)

tracks/haskell/exercises/pangram/mentoring.md
```diff
@@ -4,19 +4,64 @@
...
 import Data.Char (toLower)
+import Data.Set (fromList, delete)
+
+alphabet :: Set Char
+alphabet = fromList ['a'..'z']
 
 isPangram :: String -> Bool
-isPangram xs = null (['a'..'z'] \\ map toLower xs)
+isPangram = any null . scanl deleteLower alphabet
+  where deleteLower cs c = delete (toLower c) cs
```
MatthijsBlom, line 34: @sshine A memory leak in this function was discovered by @mx-ws. The problem is that `delete`s bunch up behind the first `(:)` because `null` does not chase them to the end. A workaround is to use `(0 ==) . length` or `Set.delete` instead, but maybe you have other ideas?

PR #699, opened 2019-01-16
Update a.json

astoehr:
Added bio info for astoehr


astoehr:
Thank you!  Happy to be here!

PR #704, opened 2019-01-17
[Kotlin]: Add another hamming solution that uses deconstruction

mikegehard:
A bit more typing but makes the comparison a bit more readable.


mikegehard:
That's good feedback. I'll look into running KLint on future submissions.

SleeplessByte:
Ping @exercism/kotlin -- if there are no objections let's get this merged and iterate later 👍

PR #708, opened 2019-01-18
Added Quentin Laillé to the list of C mentors

QLaille:


QLaille:
Added indentation as requested.
I also squash'd my commits into one.

PR #710, opened 2019-01-19
Add rotational cipher solution for Kotlin

mikegehard:
Add something that we can iterate on based on a solution I was mentoring.

### DIFF
--- a/tracks/kotlin/exercises/rotational-cipher/mentoring.md
+++ b/tracks/kotlin/exercises/rotational-cipher/mentoring.md
@@ -1,0 +1,42 @@
+### Reasonable Solutions
+
+```kotlin
+class RotationalCipher(private val key: Int) {
+    init {
+        require(key in 0..26) {
+            "Key should be integer between 0 and 26"
+        }
+    }
+
+    fun encode(input: String): String =
+            input.map(shiftBy(key)).joinToString("")
+
+    // Do not allow private members to access object state.
+    // This allows for easier refactoring to another object
+    // because their state is self contained.
+    private fun shiftBy(key: Int): (Char) -> Char = {
+        if (shouldNotShift(it)) it
+        else doShift(key, it)
+    }
+
+    private fun doShift(key: Int, c: Char): Char {
+        val mod = 26
+        val shift = if (c.isUpperCase()) {
+            65
+        } else {
+            97
+        }
+
+        return ((c.toInt() + key - shift) % mod + shift).toChar()
+    }
+
+    private fun shouldNotShift(c: Char): Boolean {
+        val symbolRange = 32..64
+        return c.toInt() in symbolRange
+    }
+}
+```
+
+Thanks go out to (Xifax)[https://exercism.io/profiles/Xifax] for the inspiration.
+
+### Common suggestions
--- a/tracks/kotlin/exercises/rotational-cipher/mentoring.md
+++ b/tracks/kotlin/exercises/rotational-cipher/mentoring.md
@@ -24,7 +24,5 @@ class RotationalCipher(private val key: Int) {
 }
 ```
 
-Thanks go out to (Xifax)[https://exercism.io/profiles/Xifax] for the inspiration.
-
 ### Common suggestions
 * Do not allow private members to access object state. This allows for easier refactoring to another object because their state is self contained.
--- a/tracks/kotlin/exercises/rotational-cipher/mentoring.md
+++ b/tracks/kotlin/exercises/rotational-cipher/mentoring.md
@@ -15,11 +15,11 @@ class RotationalCipher(private val key: Int) {
         if (it.isLetter()) doShift(key, it) else it
     }
 
-    private fun doShift(key: Int, c: Char): Char {
+    private fun doShift(key: Int, char: Char): Char {
         val mod = 26
-        val shift = if (c.isUpperCase()) 65 else 97
+        val shift = if (char.isUpperCase()) 65 else 97
 
-        return ((c.toInt() + key - shift) % mod + shift).toChar()
+        return ((char.toInt() + key - shift) % mod + shift).toChar()
     }
 }
 ```
--- a/tracks/kotlin/exercises/rotational-cipher/mentoring.md
+++ b/tracks/kotlin/exercises/rotational-cipher/mentoring.md
@@ -11,28 +11,15 @@ class RotationalCipher(private val key: Int) {
     fun encode(input: String): String =
             input.map(shiftBy(key)).joinToString("")
 
-    // Do not allow private members to access object state.
-    // This allows for easier refactoring to another object
-    // because their state is self contained.
     private fun shiftBy(key: Int): (Char) -> Char = {
-        if (shouldNotShift(it)) it
-        else doShift(key, it)
+        if (it.isLetter()) doShift(key, it) else it
     }
 
     private fun doShift(key: Int, c: Char): Char {
         val mod = 26
-        val shift = if (c.isUpperCase()) {
-            65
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,2 +1,4 @@
 # ignore Mac OS temp file
 .DS_Store
+.idea/
```
SleeplessByte: Can this be a separate pull? It will be tagged weirdly in the git history otherwise.

tracks/kotlin/exercises/rotational-cipher/mentoring.md
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    fun encode(input: String): String =
+            input.map(shiftBy(key)).joinToString("")
+
+    // Do not allow private members to access object state.
+    // This allows for easier refactoring to another object
+    // because their state is self contained.
+    private fun shiftBy(key: Int): (Char) -> Char = {
+        if (shouldNotShift(it)) it
```
SleeplessByte: Kotlin's official style guide suggests inline for this:

```kotlin
if (shouldNotShift(it)) it else doShift(key, it)
```

tracks/kotlin/exercises/rotational-cipher/mentoring.md
```diff
@@ -0,0 +1,42 @@
...
+        else doShift(key, it)
+    }
+
+    private fun doShift(key: Int, c: Char): Char {
+        val mod = 26
+        val shift = if (c.isUpperCase()) {
```
SleeplessByte: Kotlin's official style guide suggests no brackets and inline for this:

```kotlin
val shift = if (c.isUpperCase()) 65 else 97
```

tracks/kotlin/exercises/rotational-cipher/mentoring.md
```diff
@@ -0,0 +1,42 @@
...
+
+    // Do not allow private members to access object state.
+    // This allows for easier refactoring to another object
+    // because their state is self contained.
+    private fun shiftBy(key: Int): (Char) -> Char = {
+        if (shouldNotShift(it)) it
+        else doShift(key, it)
+    }
+
+    private fun doShift(key: Int, c: Char): Char {
```
SleeplessByte: Use descriptive names such as `char` or `character`.

tracks/kotlin/exercises/rotational-cipher/mentoring.md
```diff
@@ -0,0 +1,42 @@
...
+        }
+    }
+
+    fun encode(input: String): String =
+            input.map(shiftBy(key)).joinToString("")
+
+    // Do not allow private members to access object state.
+    // This allows for easier refactoring to another object
+    // because their state is self contained.
+    private fun shiftBy(key: Int): (Char) -> Char = {
```
SleeplessByte: Use '[`Char#isLetter`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/is-letter.html) instead, which allows you to drop `shouldNotShift` completely.

tracks/kotlin/exercises/rotational-cipher/mentoring.md
```diff
@@ -0,0 +1,42 @@
...
+    init {
+        require(key in 0..26) {
+            "Key should be integer between 0 and 26"
+        }
+    }
+
+    fun encode(input: String): String =
+            input.map(shiftBy(key)).joinToString("")
+
+    // Do not allow private members to access object state.
```
SleeplessByte: I don't think this commentary makes sense here. Sure, it's good advice, but doesn't add to this solution. Instead add a talking point 💯

tracks/kotlin/exercises/rotational-cipher/mentoring.md
SleeplessByte: I prefer these magic numbers (`mod`, `65` and `97`) to be private constants on the `companion object`.

tracks/kotlin/exercises/rotational-cipher/mentoring.md
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    private fun shouldNotShift(c: Char): Boolean {
+        val symbolRange = 32..64
+        return c.toInt() in symbolRange
+    }
+}
+```
+
+Thanks go out to (Xifax)[https://exercism.io/profiles/Xifax] for the inspiration.
```
SleeplessByte: This might make sense now, but not if this is iterated on over and over. If they are on Github, shout them out and praise in the PR instead 🎉

tracks/kotlin/exercises/rotational-cipher/mentoring.md
mikegehard: Would you still accept the pull without the private constants for now to get more input?

I prefer them here because the context within which they sit explain their use clearly enough without another abstraction. Also, they are only used here so adding them as consts on the companion object increase the scope in which they are visible, adding to the mental overhead of where I care about them.

tracks/kotlin/exercises/rotational-cipher/mentoring.md
mikegehard: This seems like another place where we disagree. Would you accept this pull request without it?

I actually prefer the shorter names in this case because the type signature tells me what this thing is...making a longer name doesn't add anything to that understanding but just adds more characters.

tracks/kotlin/exercises/rotational-cipher/mentoring.md
mikegehard: 👍

SleeplessByte:
Yeah, this is why I said "I prefer" hehe. I'll approve it either way :)

mikegehard:
Thanks for the feedback...some good refinements in there.

I did have some push back on a couple of the "preference" items that I wanted your feedback on.

tracks/kotlin/exercises/rotational-cipher/mentoring.md
SleeplessByte: I replied by email, but yes. It's fine for now. The biggest issue is that there is no context. You and I know these are A and Z. But that's not apparent enough. 

However you are right that having something that's not the best but still good is better than having nothing at all!

tracks/kotlin/exercises/rotational-cipher/mentoring.md
mikegehard: If this method were longer I would definitely want something more descriptive...I guess this comes from my days of writing Go. :-)

SleeplessByte: This isn't code golf and the Kotlin style guide specifically says to use descriptive names.

Further down the type annotation is not there unless you're in an IDE. I had to scroll up to figure out what c was. I don't want me tors suggesting using non descriptive single character arguments, even though they're not supposed to comment on style in the first place :)

Don't be afraid of characters! :D

mikegehard:
Thanks goes out to @Xifax for the inspiration on this solution.

SleeplessByte:
Just commented on the inlines :) if the character name is the only thing left I'll approve it, but I do encourage you to write them out even if you use it once. :D

A student may forego it but this is for the highest atainable clarity for the mentor.

tracks/kotlin/exercises/rotational-cipher/mentoring.md
mikegehard: Good point on this being for mentors and a way to create consistency...I'll make the change. :-)

SleeplessByte:
@exercism/kotlin -- if there are no objections let's get this merged and iterate later 👍

PR #711, opened 2019-01-19
Add a possible solution for kotlin secret-handshake

mikegehard:
This is probably close but may require some tweaking.


iHiD:
I don't know Kotlin but the notes look sane so I'm merging.

iHiD:
Thanks 💙

PR #716, opened 2019-01-22
Fix link to become a maintainer

SleeplessByte:
This fixes the link in this document. It was pointing to a non-existing path.

### DIFF
--- a/pages/contribute.md
+++ b/pages/contribute.md
@@ -2,7 +2,7 @@ We're really excited and grateful that you want to help with Exercism.
 
 There are three main ways to contribute to Exercism:
 - [Become a mentor](http://mentoring.exercism.io/) and help others learn a language.
-- Get involved in creating and managing a language track and eventually [become a maintainer](/become-a-maintainer).
+- Get involved in creating and managing a language track and eventually [become a maintainer](/pages/become-a-maintainer.md).
 - Report issues on [GitHub](https://github.com/exercism/exercism.io/issues).
 
 Thanks! :)


pages/contribute.md
```diff
@@ -2,7 +2,7 @@ We're really excited and grateful that you want to help with Exercism.
 
 There are three main ways to contribute to Exercism:
 - [Become a mentor](http://mentoring.exercism.io/) and help others learn a language.
-- Get involved in creating and managing a language track and eventually [become a maintainer](/become-a-maintainer).
```
tejasbubane: Hmm. This link points to the correct page - https://exercism.io/become-a-maintainer. I guess this change is not required?

tejasbubane: Or am I missing something? I am navigating from this page - https://exercism.io/contribute

SleeplessByte:
Ah. I was navigating from GitHub. I'll make it an absolute link, that fixes
it across the board :D

tejasbubane:
@SleeplessByte There are a lot of links like that. eg: All the links [on this page](https://github.com/exercism/website-copy/blob/master/pages/about.md). 

I suggest you open an issue to discuss with others before making the change. There might be a reason why the links are relative - my guess is to keep them working even if the base URL changes later.

SleeplessByte:
@tejasbubane yep! My issue is that some are not hehe. Either everything should be or nothing should be, not half of them. Will open a discussion and propose both fixes.

PR #727, opened 2019-01-22
typescript/matrix: minimal mentor notes

SleeplessByte:
Add minimal mentor notes for `matrix`.

**Note**: This is a bit ahead of the track. For round 1, https://github.com/exercism/javascript/pull/590 is most likely coming for the typescript track as well. As `Simple Cipher` will move down considerably, this is actually the first time they'll really _need_ `split` -- even though they could have used it in `RNA-transcription` and `pangram`. This is also the first time they can see the "function as first class citizen", as they'll need to use `Number(other)` or `parseInt(string)` with `map`.

@exercism/typescript

### DIFF
--- a/tracks/typescript/exercises/matrix/mentoring.md
+++ b/tracks/typescript/exercises/matrix/mentoring.md
@@ -1,7 +1,8 @@
 This exercise introduces:
-- String reader via [`String#split`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split) and [`Array#map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map)
+- String reader via [`String#split`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split)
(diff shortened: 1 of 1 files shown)


tracks/typescript/exercises/matrix/mentoring.md
```diff
@@ -0,0 +1,78 @@
+This exercise introduces:
```
iHiD: Can you merge this into one line rather than having the bullet point pls?

tracks/typescript/exercises/matrix/mentoring.md
```diff
@@ -0,0 +1,78 @@
...
+- String reader via [`String#split`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split) and [`Array#map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map)
+
+### Readonable solutions
```
iHiD: Typo

tracks/typescript/exercises/matrix/mentoring.md
iHiD: Actually, split it into two bullets?

PR #730, opened 2019-01-23
Rename method name to be consistent with test suite

michael-berger-FR:
Current "reasonable solution" does not pass tests. I propose to rename method name to be consistent with test suite.


SleeplessByte:
@michael-berger-FR thank you!

Java team: merging this in, as it's only a `mentoring.md` minor change that seems right 💯

PR #740, opened 2019-01-23
javascript/pascals-triangle: minimal mentor notes

SleeplessByte:


tracks/javascript/exercises/pascals-triangle/mentorings.md
```diff
@@ -0,0 +1,52 @@
...
+            }
+            this.rows.push(this.lastRow)
+        }
+    }
+}
+```
+There are many valid ways implementing building the triangle, which all reduce
+to the same and as long as the logic is sound, you should approve it.
+
+### Common suggestions
```
iHiD: Remove both the unused headers here pls?

PR #753, opened 2019-01-28
Add mentor bio for Samusan

SamuSan:


mentors/ruby/s.json
```diff
@@ -59,5 +59,13 @@
     "link_url": null,
     "avatar_url": null,
     "bio": "When I started dabbling with programming in Ruby over 15 years ago, I did not expect it to pay my bills ánd make me happy!"
+  },
+  {
+    "github_username": "Samusan",
```
tejasbubane: Shouldn't it be `SamuSan`. Although github seems to recognize this as well, I suggest using correct one to avoid any error.

rpottsoh: I agree with @tejasbubane.  `Rakefile` is not checking for this type of spelling mistake.  The only check related to spelling of the name that is taking place is simply to make sure that the first letter of the name agrees with the name of the bio JSON file being updated.

I have not researched this, but I am assuming the [Rakefile](https://github.com/exercism/website-copy/blob/master/Rakefile) could be modified to perform a case-sensitive spelling check of the name...

PR #757, opened 2019-01-29
Update s.json

sbromberger:
Added myself to mentor list for Julia.

### DIFF
--- a/mentors/julia/s.json
+++ b/mentors/julia/s.json
@@ -1,1 +1,10 @@
-[]
+[
+{
+  "github_username": "sbromberger",
+  "name": "Seth Bromberger",
+  "link_text": "Homepage",
+  "link_url": "https://www.bromberger.com",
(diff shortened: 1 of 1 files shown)


sbromberger:
I'm assuming that markdown is available in the text. If not, let me know and I'll change my bio accordingly.

rpottsoh:
Good question.  I don't know.  Probably no harm in leaving it in.  Once merged you should appear as a mentor on the track page fairly quickly and we'll know if the MD worked. If it doesn't just file another PR to remove the md.

Once minor change that would be appreciated would be to indent everything between `[` and `]` by two spaces.  So it looks like:
```
[
  {
    "something": "and something else"
  }
]
```

sbromberger:
sorry about the indentation issue. Should be fixed now.

sbromberger:
#759

PR #758, opened 2019-01-30
kotlin/Add a possible solution for bank account

mikegehard:


SleeplessByte:
Shouldn't `get` and ` close` also be `@synchronized`?

mikegehard:
There but there could be a race condition if one thread was changing the state and another reading for `get` but no tests enforce it. :-(

`close` could probably stand to be synchronized because it changes state but there aren't any tests for it.

This is why I dislike mutating state. :-(

SleeplessByte:
Yes, but that was not my point ;) 

You've already pointed out that synchronised only works on the method level. IMO it doesn't make sense to make one of these methods synchronised at all. If you'd like to make this thread-safe, implement a proper lock (or semaphore). No race-conditions.

mikegehard:
Added `@Synchronized` to the other methods based on your feedback.

Not sure having the student implement a full locking mechanism in a medium difficulty problem aligns with the difficulty level. :-)

SleeplessByte:
I would almost say: No synchronisation is better than some, as this does _not_ make the class thread-safe.

@exercism/kotlin-contributors please check.

mikegehard:
@SleeplessByte I'm wondering if there is more work here since it got assigned to me.

You mentioned that the changes I made adding synchronized to the getter and to close() don't make the class thread safe. Maybe I'm misunderstanding how that works. Would it be possible for you to write some tests that would assure that the implementation would be thread safe?

Are you thinking that we'd remove the part of the problem that reads?

> Create an account that can be accessed from multiple threads/processes (terminology depends on your programming language).

SleeplessByte:
> Would it be possible for you to write some tests that would assure that the implementation would be thread safe?

Yes sir! I'll do this.

Are you thinking that we'd remove the part of the problem that reads?

> Create an account that can be accessed from multiple threads/processes (terminology depends on your programming language).

Nah, that's fine. I think this is actually an exercise that is _meant_ to be concurrent!

Your biggest issue is that `@Synchronised` only makes it lock the _method_, not the _instance_. What I mean is that. If all the methods are  `@Synchronised`, this is the behaviour:

```
account(balance: 100)
Thread 1: adjustBalance(10) // => ok
Thread 1: balance  // => 110 OR 112
Thread 2: adjustBalance(2)   // => ok
Thread 2: balance  // => 110 OR 112

// This is fine. AFTER both threads are done, the balance is _always_ 112
```

```
account(balance: 100)
Thread 1: close // => ok
Thread 2: balance // => 100 OR boom

// Sometimes this will _boom_ on `check`, sometimes this will work
```

mikegehard:
Gothca...that all makes sense. 

Yeah my understanding of sychronized was a bit off.  It's been a while since I've written code that mutates object (been using immutable objects). :-)

Let's see where we end up after you get the tests done and see how much work making it fully synchronized is. :-)

Thanks!

rpottsoh:
Pinging @mikegehard and @SleeplessByte.  Where does this stand?

SleeplessByte:
Flew under the radar. This is me!

sshine:
@SleeplessByte: What is the least amount of work you can do to close this?

SleeplessByte:
It's on the backlog because I haven't had the time for it. Let me see what we can do.

SleeplessByte:
(close-reopening to make CI run, and go from there)

SleeplessByte (approved):

PR #763, opened 2019-02-02
Update faqs.md

AntHass:
I guess inexperienced users, like your truly, would not know what to do if the 'dotnet test' command fails. It took me ages to figure out!

### DIFF
--- a/pages/faqs.md
+++ b/pages/faqs.md
@@ -31,6 +31,10 @@ A new language track gets created when a member of the community takes the lead 
 
 ## Exercism Command-line Client FAQs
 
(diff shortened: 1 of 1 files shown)


NobbZ:
This is very specific to the dotnet tracks (C# and F#, IIRC) and would be misleading for those that are unable to test their JavaScript using jasmine.

@erikschierbohm would you take care of this?

robkeim:
It does feel like there's quite a bit of information already available, but @AntHass please let us know if there are certain things that we can change to make this more explicit. We certainly don't want to lose people in the setup.

PR #765, opened 2019-02-02
replace non-English character in mentor bio info

rpottsoh:
replace "ł" with "l"

### DIFF
--- a/mentors/javascript/p.json
+++ b/mentors/javascript/p.json
@@ -9,7 +9,7 @@
   },
   {
     "github_username": "pablo83",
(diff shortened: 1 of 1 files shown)


kotp:
> But that is probably not his name.

:100:

rpottsoh:
It looks fine to me https://exercism.io/tracks/javascript/mentors.  Almost can't tell that it is `ł` instead of `l`.  I'm fine to close this.  @kotp?

kotp:
Definitely a close vote from me.

PR #775, opened 2019-02-06
Create ruby/scrabble-score

iHiD:
@exercism/website-copy-ruby I just want to get a reasonable solution on master as hints for mentors. We can then flesh this out :)


iHiD:
Thanks @rpottsoh

PR #776, opened 2019-02-06
[Java/Hamming] Improve mentor notes

michael-berger-FR:
Offer a reasonable solution with String instead of char[] (both approaches are valid).

### DIFF
--- a/tracks/java/exercises/hamming/mentoring.md
+++ b/tracks/java/exercises/hamming/mentoring.md
@@ -2,19 +2,19 @@
 
 ```java
 class Hamming {
-    private char[] leftStrand, rightStrand;
+    private final String leftStrand, rightStrand;
 
     Hamming(String leftStrand, String rightStrand) {
         if (leftStrand.length() != rightStrand.length())
             throw new IllegalArgumentException("leftStrand and rightStrand must be of equal length.");
-        this.leftStrand = leftStrand.toCharArray();
-        this.rightStrand = rightStrand.toCharArray();
+        this.leftStrand = leftStrand;
+        this.rightStrand = rightStrand;
     }
 
     int getHammingDistance() {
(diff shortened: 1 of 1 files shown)


michael-berger-FR:
Thanks @Smarticles101 for the feedback. I tried to find a reference for "Using `charAt` should be more performant over `toCharArray`" and stumbled on that topic [closed as primarily opinion-based](https://stackoverflow.com/questions/22445104/should-i-use-charat-or-tochararray). What do you think?

SleeplessByte:
It's closed as primarily opinion based because the question isn't "which is faster" or "which is more common", both can be answered using empirical evidence, but "which is better", which requires someone to value based on opinion (e.g faster isn't always better, readability might be).

IMO you should have _both_ in the notes, as well as this version:

```java
for(char leftChar : leftStrand.toCharArray()) {
  ...
}
```

A student who submits the latter did not necessarily miss anything in the exercise (what are we trying to teach here?) and it makes for a nice talking point of speed vs readability / raw performance, and asking if the student knows of `charAt` in the first place.

michael-berger-FR:
I apologize if my answer was unclear... I did not want to emphasize on the "opinion-based" part! @Smarticles101 was saying that "using `charAt` should be more performant over `toCharArray`" and I wasn't sure it was the case. Anyway, we agree that both approaches are valid :-)

@SleeplessByte , I'm a bit confused with your proposal `for(char leftChar : leftStrand.toCharArray())`: how do you compare left and right nucleotides if you have no index?

SleeplessByte:
@michael-berger-FR bookkeeping variables! For example, keeping track of index. 

I wasn't clear though. I did not mean under solution, but under suggestions of what to change! Bookkeeping variables with a `for :` loop are less idiomatic than your actual solutions ❤️

PR #777, opened 2019-02-06
Simplify ruby/scrabble-score

iHiD:
Add a simpler reasonable solution.

### DIFF
--- a/tracks/ruby/exercises/scrabble-score/mentoring.md
+++ b/tracks/ruby/exercises/scrabble-score/mentoring.md
@@ -2,6 +2,39 @@
 
 ```ruby
 class Scrabble
+  LETTERS = [
+    [%w(A E I O U L N R S T), 1],
+    [%w(D G), 2],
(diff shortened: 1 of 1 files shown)


pgaspar:
I like this exercise 🙂 I'm going through it [here](https://exercism.io/my/solutions/2eecc37d4c934cc3a2fec10c93dee4f8?iteration_idx=3) and I also struggle in the trade-offs between solutions like these and "score per letter" solutions like @F3PiX is suggesting.

This is probably one of those exercises where mentoring is more nuanced because, at least to me, both approaches have interesting advantages and it ends up depending on what you value most: performance or a clearer link between exercise instructions and the data representation on the solution.

Anyway, :+1: on these changes!

F3PiX:
Thanks Jeremy!

iHiD:
Thanks both. I really like this exercise too. It's my favourite on the track so far.

iHiD:
(both to do and to mentor)

PR #783, opened 2019-02-12
Ruby/HighScores Update with latest changes

F3PiX:
Update changelog, See PR https://github.com/exercism/ruby/pull/936

I kept the 'report' method in the example for now, because there's a lot of the old version out there

### DIFF
--- a/tracks/ruby/exercises/high-scores/mentoring.md
+++ b/tracks/ruby/exercises/high-scores/mentoring.md
@@ -24,21 +24,13 @@ class HighScores
   def personal_top
     scores.sort.reverse.take(3) 
   end
-
-  # this method is removed in version 3/4
-  def report
-    difference = "#{personal_best - latest} short of" unless personal_best == latest
-
-    "Your latest score was #{latest}. That's #{difference} your personal best!".squeeze(' ')
-  end
 end
 
 ```
 ### Reasonable variants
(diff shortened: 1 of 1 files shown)


F3PiX:
@maintainers/ruby

tracks/ruby/exercises/high-scores/mentoring.md
```diff
@@ -59,4 +60,5 @@ Concatenation mutates the string instance in-place, `String#+` creates new strin
 Ruby 2.3 will freeze string literals by default.
 
 ### Changelog
-- Version 2 changed method names "highest" -> "personal_best", "top" -> "personal_top"
+- Version 3/4 changed method names "personal_top" -> "personal_top_three" and removed the "report" method.
```
pgaspar: I would make this 2 lines, even though version 3 was never released by itself. But I don't feel strongly about it if you'd rather not.

F3PiX: My thinking was: separate lines suggest that there are 2 extra versions around, and I hoped that this implicated there's only 1 newer version.

pgaspar:
@F3PiX let me know if you think my comments regarding removing `report` mentions should be addressed in a separate issue.

F3PiX:
@pgaspar All cleaned up!

PR #785, opened 2019-02-13
[Java/Scrabble-score] Add mentoring notes

michael-berger-FR:
Add mentoring.md file with two reasonable solutions


SleeplessByte:
Thank you @michael-berger-FR ! 🥇

PR #790, opened 2019-02-15
Add my profile

ningke:
Add my mentor profile.

PR #791, opened 2019-02-15
General improvements to Rakefile

rpottsoh:
Tests for "github_username" and "name" simplified

Improved checks of "link_text"
- Only one correct way to spell `GitHub`.
- should be null if same as "link_url"
- should be null if "link_url" is null / ""

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -17,43 +17,54 @@ task :test do
       end
 
       mentors.each do |mentor|
-        username = mentor["github_username"]
-        name = mentor["name"]
+        github_username = mentor["github_username"].to_s
+        name = mentor["name"].to_s
 
-        if username == "" || username.nil?
+        if github_username.empty?
           errors << "Github username can't be blank for %s." % [file]
           next
         end
 
-        unless username.downcase.start_with?(filename)
-          errors << "First letter of github_username \"%s\" must equal JSON filename %s" % [username, file]
+        unless github_username.downcase.start_with?(filename)
+          errors << "First letter of github_username \"%s\" must equal JSON filename %s" % [github_username, file]
           next
         end
 
-        if name == "" || name.nil?
-          errors << "Name can't be blank for %s in %s." % [username, file]
+        if name.empty?
+          errors << "Name can't be blank for %s in %s." % [github_username, file]
           next
         end
 
-        url = mentor["link_url"]
-        next if url.nil? || url.empty?
-
-        if url.strip == "null"
-          errors << "Link URL should be null, not the string 'null' for %s in %s." % [username, file]
+        link_text = mentor["link_text"].to_s
+        link_url = mentor["link_url"].to_s
+        if link_url.empty?
+          if !link_text.empty?
+            errors << "\"link_text\" should be null since \"link_url\" is null/empty for %s in %s." % [github_username, file]
+          end
           next
         end
 
-        if url.strip != url
-          errors << "Link URL has extraneous whitespace for %s in %s." % [username, file]
+        if link_url.strip == "null"
+          errors << "Link URL should be null, not the string 'null' for %s in %s." % [github_username, file]
           next
         end
 
-        if url !~ /^https?/
-          errors << "Link URL must have HTTP protocol for %s in %s." % [username, file]
+        if link_url.strip != link_url
+          errors << "Link URL has extraneous whitespace for %s in %s." % [github_username, file]
+          next
         end
 
-        if mentor["link_text"].to_s.include?("Github")
-          errors << "GitHub should be spelled with an uppercase 'H' for %s in %s." % [username, file]
+        if link_url !~ /^https?/
+          errors << "Link URL must have HTTP protocol for %s in %s." % [github_username, file]
+        end
+
+        if link_text.downcase.gsub(/https?\:\/\//,"") == link_url.downcase.gsub(/https?\:\/\//,"")
+          errors << "Use null for \"link_text\" when equivalent to \"link_url\" for %s in %s." % [github_username, file]
+          next
+        end
+
+        if link_text.downcase.include?("github") && !link_text.include?("GitHub")
+          errors << "Capitalization of GitHub is incorrect in \"link_text\" for %s in %s." % [github_username, file]
         end
       end
     rescue


Rakefile
```diff
@@ -52,8 +52,12 @@ task :test do
           errors << "Link URL must have HTTP protocol for %s in %s." % [username, file]
         end
 
-        if mentor["link_text"].to_s.include?("Github")
-          errors << "GitHub should be spelled with an uppercase 'H' for %s in %s." % [username, file]
+        linkText = mentor{"link_text"].to_s
+        if linkText.downcase.include?("github")
+          unless linkText.include?("GitHub")
+            errors << "GitHub should be spelled with lowercase 'i, t, u, and b' and uppercase 'G' and 'H' for %s in %s." % [username, file]
+            next
```
rpottsoh: I think I probably don't need `next`.
```suggestion
```

rpottsoh:
This change was inspired by #790.

rpottsoh:
Travis CI should stop failing once #792 is merged.

Rakefile
```diff
@@ -52,8 +52,13 @@ task :test do
           errors << "Link URL must have HTTP protocol for %s in %s." % [username, file]
         end
 
-        if mentor["link_text"].to_s.include?("Github")
-          errors << "GitHub should be spelled with an uppercase 'H' for %s in %s." % [username, file]
+        linkText = mentor["link_text"].to_s
+        unless linkText == url
```
kytrinyx: Can we say `if linkText != url` here? I find `unless` really difficult to process in this context.

Rakefile
```diff
@@ -52,8 +52,13 @@ task :test do
...
+          if linkText.downcase.include?("github")
```
kytrinyx: 👍

Rakefile
```diff
@@ -52,8 +52,13 @@ task :test do
...
+            unless linkText.include?("GitHub")
```
kytrinyx: Maybe:

```
if linkText.downcase.include?("github") && !linkText.include?("GitHub")
```

Rakefile
rpottsoh: Certainly.  I don't have a good sense of when to use `unless` or when to use `if !=`.

Rakefile
rpottsoh: I like it this way too.

rpottsoh:
Thanks for taking a look @kytrinyx.  I'll probably have an update posted this evening.

kytrinyx:
When looking at https://github.com/exercism/website-copy/pull/792 it occurred to me:

I wonder if it's worth having linting that ensures that link_text is null if it's essentially the same as link_url (e.g. either identical or just missing the HTTP schema).

rpottsoh:
> I wonder if it's worth having linting that ensures that link_text is null if it's essentially the same as link_url (e.g. either identical or just missing the HTTP schema).

This is what I was essentially doing by hand for my first commit to #792 https://github.com/exercism/website-copy/pull/792/commits/f47407ab74d212751474052137a3653803ddc9df but I think not knowing how the site would cope with "link_text" as null when "link_url" wasn't created some concern.  That brought about the second commit to #792 which left "link_text" alone if it is a match to "link_url".

I am of the opinion that "link_text" should be null if your desire is to use "link_url" as text too.  "link_text" shouldn't be "" if "link_url" is not _blank_.

I'll work this whole idea into my update as well.

rpottsoh:
Sample output:
```
D:\projects.git\Exercism\exercism-website-copy>rake test
- "link_text" should be null instead of same as "link_url" for manuri in mentors/ballerina/m.json.
- "link_text" should be null instead of same as "link_url" for MaryamZi in mentors/ballerina/m.json.
- "link_text" should be null instead of same as "link_url" for fozcodes in mentors/elixir/f.json.
- "link_text" should be null instead of same as "link_url" for sshine in mentors/haskell/s.json.
- GitHub should be spelled with lowercase 'i, t, u, and b' and uppercase 'G' and 'H' for alexZielonko in mentors/javascript/a.json.
- "link_text" should be null instead of same as "link_url" for aibrahim3546 in mentors/javascript/a.json.
- "link_text" should be null instead of same as "link_url" for fozcodes in mentors/javascript/f.json.
- GitHub should be spelled with lowercase 'i, t, u, and b' and uppercase 'G' and 'H' for tekerson in mentors/javascript/t.json.
- GitHub should be spelled with lowercase 'i, t, u, and b' and uppercase 'G' and 'H' for therynamo in mentors/javascript/t.json.
- "link_text" should be null instead of same as "link_url" for sshine in mentors/ocaml/s.json.
- "link_text" should be null instead of same as "link_url" for rishiip in mentors/ruby/r.json.
- GitHub should be spelled with lowercase 'i, t, u, and b' and uppercase 'G' and 'H' for tekerson in mentors/rust/t.json.
- "link_text" should be null instead of same as "link_url" for sshine in mentors/sml/s.json.
- GitHub should be spelled with lowercase 'i, t, u, and b' and uppercase 'G' and 'H' for tekerson in mentors/typescript/t.json.
```

rpottsoh:
#792 will need to be updated once the changes here are agreed upon.

rpottsoh:
@kytrinyx I have made the changes you suggested and added another test.  Opening comment to the PR updated.  Title of PR also changed since focus has shifted from just making sure GitHub is spelled correctly to making sure "link_text" has reasonable information assigned to it.

kytrinyx:
Thanks @rpottsoh !

Looking at the CI failures, I think the logic isn't quite right yet. For tekerson CI says:

```
- GitHub should be spelled with lowercase 'i, t, u, and b' and uppercase 'G' and 'H' for tekerson in mentors/javascript/t.json.
```

And the file says:

https://github.com/exercism/website-copy/blob/5539dc7b409bb3e733ba9d66a02ff78107cb98ed/mentors/javascript/t.json#L10-L17

I believe that should complain that the link url and link text are equivalent rather than complaining about the case.

In terms of the error messages I wonder if we could be more succinct. Maybe, instead of:

> GitHub should be spelled with lowercase 'i, t, u, and b' and uppercase 'G' and 'H' for therynamo in mentors/javascript/t.json.

something like:

```
Capitalization of GitHub is incorrect in `link_text` for therynamo in mentors/javascript/t.json.
```

And for:

> "link_text" should be null instead of same as "link_url" for aibrahim3546 in mentors/javascript/a.json.

Maybe something like:

```
Use null for 'link_text' when equivalent to 'link_url'.
```

rpottsoh:
> I believe that should complain that the link url and link text are equivalent rather than complaining about the case.

This was in the back of my mind.  I need to figure out how to remove substr from a string in Ruby.  I don't Ruby much at all and working on this is stretching me a little. :+1:

I like your suggestions for the error messages.  I'll work them in.

kytrinyx:
@rpottsoh :-) Try opening an IRB console and playing with `"somestring".gsub("some", "")`

rpottsoh:
Will do, thanks for the helping :hand:.

Rakefile
```diff
@@ -52,8 +58,13 @@ task :test do
           errors << "Link URL must have HTTP protocol for %s in %s." % [username, file]
         end
 
-        if mentor["link_text"].to_s.include?("Github")
-          errors << "GitHub should be spelled with an uppercase 'H' for %s in %s." % [username, file]
+        if linkText.downcase.gsub("https://","").gsub("http://","") == url.downcase.gsub("https://","").gsub("http://","")
```
rpottsoh: @kytrinyx if there is a cleaner way to go about this `if` I am all for it.  However, this is working.  All the bios that Travis CI is complaining about based on this latest version of Rakefile have been updated in #792.

Once we are happy with the way Rakefile is operating and all the necessary changes have been made to the bios I plan to remove `WIP` from #792.

Rakefile
```diff
@@ -35,8 +35,14 @@ task :test do
           next
         end
 
+        linkText = mentor["link_text"].to_s
```
kytrinyx: Nitpick: this should be `link_text` for consistency with Ruby style.

Rakefile
```diff
@@ -35,8 +35,14 @@ task :test do
...
         url = mentor["link_url"]
-        next if url.nil? || url.empty?
+        if url.nil? || url.empty?
```
kytrinyx: Let's add the same `to_s` on line 39 that you did for line 38. That way we can check only against `empty?` and not against `nil`.

Rakefile
```diff
@@ -35,8 +35,14 @@ task :test do
...
+          if !linkText.nil? && !linkText.empty?
```
kytrinyx: Link text is never nil, since you stringified it when assigning it.

Rakefile
kytrinyx: `gsub` can take a regex for the first argument:

```
link_text.gsub(/https?\:\/\//, "")
```

rpottsoh: I assumed that would be possible, my RegEx powers are weak.

rpottsoh:
It is tempting to stringify
https://github.com/exercism/website-copy/blob/ca1f6f07b43884c21b6fce374b63d7c28f9e0b06/Rakefile#L20-L21
and remove null checks like I am doing for link_url and link_text.

If you are fine with that I would re-title this PR again as the scope will have expanded to being more of a general overhaul of sorts to the file.

Rakefile
```diff
@@ -35,25 +35,36 @@ task :test do
           next
         end
 
-        url = mentor["link_url"]
-        next if url.nil? || url.empty?
+        link_text = mentor["link_text"].to_s
+        link_url = mentor["link_url"].to_s
+        if link_url.empty?
```
rpottsoh:
```suggestion
        if link_url.empty? && !link_text.empty?
```

Rakefile
```diff
@@ -35,25 +35,36 @@ task :test do
...
+          if !link_text.empty?
```
rpottsoh:
```suggestion
```

Rakefile
```diff
@@ -35,25 +35,36 @@ task :test do
...
+            errors << "\"link_text\" should be null since \"link_url\" is null/empty for %s in %s." % [username, file]
```
rpottsoh:
```suggestion
          errors << "\"link_text\" should be null since \"link_url\" is null/empty for %s in %s." % [username, file]
```

kytrinyx:
> If you are fine with that I would re-title this PR again as the scope will have expanded to being more of a general overhaul of sorts to the file.

Yepp! I like that approach

rpottsoh:
I think this is ready to be merged.  Let me know if there is anything else you think should be changed.  I think I am going to go ahead and remove `WIP` from #792.  If there are no more changes here #792 should be merged.

kytrinyx:
I'll take a (final!) look at this today. Thank you so much for your patience @rpottsoh

rpottsoh:
Yup.  that is what #792 is for.  I just need someone to approve it.

kytrinyx:
Perfect. I've merged #792 — want to rebase this one onto the most recent master?

rpottsoh:
Absolutely.  Thanks.

rpottsoh:
I was just typing to you!  Nice, thanks for merging.  😄

PR #792, opened 2019-02-15
Fixed Bios that did not pass scrutiny of revised Rakefile

rpottsoh:
I either corrected the spelling, in some cases I replaced the text with null because the field contained the same information as field "link_url".

### DIFF
--- a/mentors/javascript/a.json
+++ b/mentors/javascript/a.json
@@ -34,7 +34,7 @@
   {
     "github_username": "alexZielonko",
     "name": "Alex Zielonko",
-    "link_text": "github.com/alexZielonko",
+    "link_text": null,
     "link_url": "https://github.com/alexZielonko",
     "avatar_url": "https://avatars1.githubusercontent.com/u/5274697?s=460&v=4",
     "bio": "I'm one part professional developer and two parts programming enthusiast, shaken with a liberal dash of pragmatism."
@@ -42,7 +42,7 @@
   {
     "github_username": "aibrahim3546",
     "name": "Ahmad Ibrahim",
-    "link_text": "https://aibrahim3546.github.io/",
+    "link_text": null,
     "link_url": "https://aibrahim3546.github.io/",
     "avatar_url": null,
     "bio": "I am a fan of javascript. I have been working with ReactJs for a year and yeah I am really good at front-end development. Also Good at NodeJs and React Native"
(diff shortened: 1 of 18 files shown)


rpottsoh:
These discrepancies were found by running `Rakefile` from #791

mentors/javascript/a.json
```diff
@@ -34,15 +34,15 @@
...
+    "link_text": "GitHub.com/alexZielonko",
     "link_url": "https://github.com/alexZielonko",
     "avatar_url": "https://avatars1.githubusercontent.com/u/5274697?s=460&v=4",
     "bio": "I'm one part professional developer and two parts programming enthusiast, shaken with a liberal dash of pragmatism."
   },
   {
     "github_username": "aibrahim3546",
     "name": "Ahmad Ibrahim",
-    "link_text": "https://aibrahim3546.github.io/",
+    "link_text": null,
```
NobbZ: These are Not equivalent, per readme null means take this fields value from GitHub.

mentors/ruby/r.json
```diff
@@ -11,7 +11,7 @@
   {
     "github_username": "rishiip",
     "name": "Rishi Pithadiya",
-    "link_text": "http://rishiip.github.io/",
+    "link_text": null,
```
NobbZ: See above

mentors/javascript/a.json
rpottsoh: I was working under the assumption that if "link_text" is null that "link_url" would be displayed instead.

rpottsoh: I think I am going to modify my changes to `Rakefile` in #791 to effectively ignore what is in "link_text" if it is identical to what is in "link_url".  This would make some of the changes I have made in this PR unnecessary.

kytrinyx:
I wonder if it's worth having linting that ensures that link_text is null if it's essentially the same as link_url (e.g. either identical or just missing the HTTP schema).

kytrinyx:
Actually, this comment is more about the linting, not the fixes so the linter passes. Moving the comment to the right PR.

rpottsoh:
This PR will be further affected by the final outcome of #791.  This PR should not be merged until after #791 is merged and all the mentor bios that do not pass the scrutiny of `Rakefile` are corrected and are included in this PR.

PR #793, opened 2019-02-20
Go/Hamming: Update mentor notes to reflect new tests

gabeguz:
The tests no longer specify a `-1` return value, update the mentor notes to reflect this.


tracks/go/exercises/hamming/mentoring.md
```diff
@@ -44,8 +44,7 @@ The most common feedback revolves around:
 * Is the error message empty or vague? In Go the returned error can provide context for where the error occurred and why, so it should concisely and specifically describe the problem.
 * Is the error formatted correctly? For example the error [should not be capitalized or have punctuation](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).
 * Did they use a combination of `errors.New` and `fmt.Sprintf`? Hint at `golint` and `fmt.Errorf`.
-* Did they use `fmt.Errorf` for a non-interpolated string? It may be worth suggesting they use `errors.New` in these cases to explicitly indicate that no interpolation will be occurring. Encouraging early exposure to the `errors` package might start benefitting them immediately, e.g., `errors.Wrap`.
-* Did they return `-1` or `0` in the error case? Returning the zero value may be more idiomatic, since the error indicates that the preceding value isn't that useful. There are test cases that specify `-1` should be returned, but because this value isn't checked by the tests, a student doing strict TDD might not discover this; careful readers might.
```
bitfield: I would still keep a mention of the `-1` issue; some people will instinctively use an out-of-band value like this in the error case, because they're used to languages where this is necessary. Even though the tests no longer specify `-1`, I think we'll still see solutions where `-1` is returned, so it would be nice for the mentor notes to pick up on this.

gabeguz: That's a good point, I'll reword it.

tracks/go/exercises/hamming/mentoring.md
```diff
@@ -44,8 +44,8 @@ The most common feedback revolves around:
...
-* Did they use `fmt.Errorf` for a non-interpolated string? It may be worth suggesting they use `errors.New` in these cases to explicitly indicate that no interpolation will be occurring. Encouraging early exposure to the `errors` package might start benefitting them immediately, e.g., `errors.Wrap`.
-* Did they return `-1` or `0` in the error case? Returning the zero value may be more idiomatic, since the error indicates that the preceding value isn't that useful. There are test cases that specify `-1` should be returned, but because this value isn't checked by the tests, a student doing strict TDD might not discover this; careful readers might. 
+* Did they use `fmt.Errorf` for a non-interpolated string? It may be worth suggesting they use `errors.New` in these cases to explicitly indicate that no interpolation will be occurring. Encouraging early exposure to the `errors` package might start benefitting them immediately, e.g., `errors.Wrap`. 
+* Did they return `-1` or `0` in the error case? Point out that returning the zero value is more idiomatic, and that returning `-1` (or any other flag value) isn't necessary in go.
```
bitfield: Nitpick: "Go", not "go". Otherwise this is great!

bitfield: Worth looking at what Exalysis says in this situation:

> - When your function returns an `error`, this tells the caller that the other return value shouldn't be used. 
Since it doesn't actually matter what you return in that case, the convention is to return the [zero value](https://golang.org/ref/spec#The_zero_value). In languages that don't allow multiple return values from functions, you often need to use an invalid 'out-of-band' value to signal an error, such as `-1`. That's not necessary in Go.

gabeguz: Ok, capitalized `Go`, and tried to make it more inline with the wording from exalysis --- I'm a bit worried that it sounds too forced now, what do you think?

SleeplessByte: By forced, do you mean forceful? I prefer the new proposed change over would/could/maybe/etc and I think it's better in mentor notes to be concise and direct.

SleeplessByte:
Given that @bitfield approved and he is an amazing Go mentor,, let's merge this in!

PR #795, opened 2019-02-21
Go: Add mentor bio for serg-kovalev

serg-kovalev:


mentors/go/s.json
```diff
@@ -22,5 +22,13 @@
...
     "avatar_url": null,
     "bio": "Go developer based out of Nashville, TN. I enjoy Go because of it's simplicity and no-frills approach to development."
+  },
+  {
+    "github_username": "serg-kovalev",
+    "name": "Sergey Kovalev",
+    "link_text": null,
+    "link_url": null,
+    "avatar_url": null,
+    "bio": "I have more than 10 years of production experience. I love Golang/Ruby because of its friendly and active community. Nothing is impossible when you do what you love! Happy coding!"
```
bitfield: A small point, perhaps, but the name of the language is "Go", not "Golang". We should probably set the right example here!

serg-kovalev: thanks, John!

PR #801, opened 2019-02-22
[Java/Hamming] Improve mentor notes

loradd:
- null-safe constructor in solution#2


SleeplessByte:
Why does this improve the mentor notes? Are the tests called with `null`? 

If not, this is not a requirement of the exercise and (at least in most tracks) we don't _encourage_ defensive programming. If this should be a requirement, please PR the canonical data and then a test in the Java exercise repo.

Ping @exercism/website-copy-java @exercism/java

loradd:
I agree, this is not required considering the current test cases. Sorry about that, still a fresh mentor ;)

PR #818, opened 2019-02-27
javascript/list-ops: add minimal notes

SleeplessByte:
Add minimal mentor notes for `list-ops`.

### DIFF
--- a/tracks/javascript/exercises/list-ops/mentoring.md
+++ b/tracks/javascript/exercises/list-ops/mentoring.md
@@ -1,0 +1,304 @@
+### Reasonable solutions
+
+This exercise has many _valid_ implementations, but also many suggestions that
+can be applied to implementations. This exercise is meant to be implemented
+with as little of built in functions as possible. This means that a backing
+array is allowed, but _no_ functions from the `Array.prototype` should be used.
+
+Most students will actually still use a backing array for the values but have
+various ways of not using the built in functions to process them. There are (at
+least) two correct ways of implementing this using only array destructuring to
+go from the test array to the internal representation and from the internal
+representation back to `values`.
+
+> *Note*: You don't need **any** function from the `Array.prototype`, and only
+> need the language feature to build construct and desconstruct arrays via
+> destructuring as well as the array literal `[]`.
+>
+> *Note*: most implementations using `for` loops, also use `.length` on the
+> backing array, or the `array[indexer]`. You don't need either!
+
+#### Functional data type without backing array
+
+This is a representation of one of the most fundamental *cons* tructs in most
+functional languages called [`Cons`](https://en.wikipedia.org/wiki/Cons). It
+is actually a way to build a linked-list.
+
+```javascript
+Cons.fromArray([1, 42, 'foo'])
+// => Cons(1, Cons(42, Cons('foo', Null)))
+
+Cons.fromArray()
+// => Null
+
+Cons(1, Cons(42, Cons('foo', Null))).values
+// => [1, 42, 'foo']
+
+Null.values
+// => []
+```
+
+The `Null` object is a convenience object to replace `null` or `undefined` at
+the end of the list and allows for code without a any branches (exception is
+the conversion code).
+
+```javascript
+const Null = {
+  get value() { return undefined },
+  get next() { return this },
+  get values() { return [] },
+
+  get() { return this.value },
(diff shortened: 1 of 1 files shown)


tracks/javascript/exercises/list-ops/mentoring.md
```diff
@@ -0,0 +1,350 @@
...
+  `backingarray.length`):
+  ```javascript
+  // using forEach
+  filter(predicate) {
+    const result = new List()
+    this.forEach(item => predicate(item) && result.push(item))
+    return result
+  }
+
+  // using foldl
```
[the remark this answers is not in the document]
SleeplessByte: woops. Foldl is 📦

tejasbubane:
If I understand recursion correctly, you should be able to define `foldl` and `foldr` using plain recursion without need of `forEach` (problem statement does not require `forEach`):

```js
foldl(callback, acc) {
  val newAcc = callback(acc, this.value);
  this.next.foldl(callback, newAcc);
}

foldr(callback, acc) {
  if (this.next == Null) {
    callback(acc, this.value);
  }
  else {
    callback(this.next.foldr(callback, acc), this.value);
  }
}
```
PS: This is some quick code I put up without any testing.

SleeplessByte:
@tejasbubane Yes, it's like that, but the if statement is unnecessary, as we can define `foldr` on `Null`.

Here are the optimal implementations using `Cons`, and you can instantly see "how" it recurses ;)

```javascript
forEach(callback) {
  this.foldl((_, item) => callback(item))
}

foldr(callback, initial = undefined) {
  return callback(this.next.foldr(callback, initial), this.value);
}

foldl(callback, initial = undefined) {
  return this.next.foldl(callback, callback(initial, this.value))
}
```

SleeplessByte:
@tejasbubane see new implementations, without a single `forEach` 💎 💎. I actually have this as my personal solution haha.

tracks/javascript/exercises/list-ops/mentoring.md
```diff
@@ -0,0 +1,304 @@
...
+    return this.next.foldl(callback, callback(initial, this.value))
+  }
+
+  forEach(callback) {
+    this.foldl((_, item) => callback(item))
+  }
+
+  foldr(callback, initial = undefined) {
+    return callback(this.next.foldr(callback, initial), this.value)
+  }
```
tejasbubane: Yay! `foldl` and `foldr` are super elegant now 👍 I guess we don't need `forEach` then?

tejasbubane:
@SleeplessByte This looks good to me. Just to be sure, can you change [example solution](https://github.com/exercism/javascript/blob/master/exercises/list-ops/example.js) so that we run tests and nothing goes wrong?

tracks/javascript/exercises/list-ops/mentoring.md
SleeplessByte: We don't, but I keep it because it's valid. Using JUST foldl is fine, but using for each is more declarative I think :)

SleeplessByte:
@tejasbubane if you can approve, I'll merge this once the js pr turns grean on ci.

SleeplessByte:
Sync is here: https://github.com/exercism/javascript/pull/624

SleeplessByte:
(Same issue as #821, CI stuck -- rebased force push)

tracks/javascript/exercises/list-ops/mentoring.md
tejasbubane: Yes, but the problem statement doesn't require `forEach`. So should we keep it in expected solution?

SleeplessByte: I don't like mentors suggesting to remove `forEach`, because using `forEach` to `fold` is only minorly less nice than `folding` to `forEach` (if you get what I mean). Yes, the current solution is optimal, but having a `forEach` is okay I guess 👍 

How do you feel about this?

tejasbubane: Right. But I think we should not add `forEach` in the expected solution. We can mention in the notes that someone can optionally choose to use `forEach` instead of fold and that is totally fine.

SleeplessByte: Got it. I'll change it!

SleeplessByte:
@tejasbubane done!

PR #820, opened 2019-02-28
javascript/wordy: add minimal notes

SleeplessByte:
Add minimal mentor notes for `wordy`.


tracks/javascript/wordy/mentoring.md
```diff
@@ -0,0 +1,133 @@
+### Reasonable solutions
+
+There are many _valid_ solutions to this problem. If the student has written
+something completely different than the options listed below, do not dismiss it.
+At this point in the track, we can except them to have developped their own
```
tejasbubane: `expect` to have `developed`

tracks/javascript/wordy/mentoring.md
```diff
@@ -0,0 +1,133 @@
...
+style. However, follow the guideliness listed in this document.
```
tejasbubane: `guidelines`

tracks/javascript/wordy/mentoring.md
```diff
@@ -0,0 +1,133 @@
...
+```javascript
+function ArgumentError() {}
+
+// or better
+class ArgumentError extends Error {}
+```
+
+The student can _not_ assign a dynamic message at this moment, because the test
+is less than perfect. Help them out if they don't get this as it's not vital to
+the exercise.
```
tejasbubane: We do have a better way to handle exceptions with messages: https://github.com/exercism/javascript/issues/280#issuecomment-299868621 It will need changing the tests though.

tracks/javascript/wordy/mentoring.md
```diff
@@ -0,0 +1,133 @@
...
+  tokens.filter(token => (token instanceof Operator)).forEach((token) => {
+    context.call(token)
+  })
+  ```
+
+Discourage `if/else` trees or `switches`.
+
+#### Error throwing
+
+Unfortunately the current implementation of checking the error is horrible so
```
tejasbubane: Prefer using less harsh words `s/horrible/not good (we are working towards changing this - meanwhile..)/`

tracks/javascript/wordy/mentoring.md
SleeplessByte: Yep. That's why it's worded like this for now.

tracks/javascript/wordy/mentoring.md
SleeplessByte: It's about the worst possible implementation. It's not testing for a class but an instance of an error. There is literally no other way to do it worse.

Should I sugarcoat it? 

I have zero issue adding a positive "change is coming" behind it tho :dancer:

PR #824, opened 2019-03-01
python/two-fer: add new lang features

dogweather:
Added the new String Interpolation and preferred single-quote style as used in PEP 498.

### DIFF
--- a/tracks/python/exercises/two-fer/mentoring.md
+++ b/tracks/python/exercises/two-fer/mentoring.md
@@ -14,7 +14,8 @@ def two_fer(name='you'):
 
 ### Common suggestions
 - Suggest using a default value instead of any form of conditionals.
-- Suggest [Literal String Interpolation introduced with Python 3.6](https://www.python.org/dev/peps/pep-0498/).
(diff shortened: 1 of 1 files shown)


tracks/python/exercises/two-fer/mentoring.md
```diff
@@ -7,8 +7,14 @@ def two_fer(name="you"):
...
 
+```python
+def two_fer(name='you'):
+    return f'One for {name}, one for me.'
+```
+
 ### Common suggestions
-- Suggest using a default value instead of any form of conditionals. 
+- Suggest using a default value instead of any form of conditionals.
+- Suggest [Literal String Interpolation introduced with Python 3.6](https://www.python.org/dev/peps/pep-0498/).
```
cmccandless: As track documentation (and source code, for that matter) still supports Python 2.7, it might be good idea to rephrase into something like:

```suggestion
- If the student is using Python 3? Suggest [Literal String Interpolation introduced with Python 3.6](https://www.python.org/dev/peps/pep-0498/).
- If the student is using Python <= 3.5, suggest upgrading to Python 3.7 for additional language features.
```

SleeplessByte:
Thank you @dogweather  ❤️

PR #826, opened 2019-03-02
TwoFer/csharp Fix broken URL in two-fer notes

Mattsi-Jansky:
Also fixed a windows whitespace character (hence no visible change
to the last line)


ErikSchierboom:
Thanks @Mattsi-Jansky !

PR #828, opened 2019-03-03
r: deduplicate examples

katrinleinweber:
I realised that with the (anti-)pattern we used in the first two mentoring notes, we'd be maintaining duplicate copies of example code.

How about we declare the [`exercism/r/../example.R` files](https://github.com/exercism/r/tree/master/exercises) as the ideal solutions and only hyperlink to them in each mentor note?

### DIFF
--- a/tracks/r/exercises/leap/mentoring.md
+++ b/tracks/r/exercises/leap/mentoring.md
@@ -1,12 +1,6 @@
 ### Reasonable solutions
 
-This is a nice, concise, efficient solution:
-
-```r
-leap <- function(year) {
-  year %% 4 == 0 & (year %% 100 != 0 | year %% 400 == 0)
-}
-```
+A nice, concise, efficient solution can be found in [`example.R`](https://github.com/exercism/r/blob/master/exercises/leap/example.R).
 
 
 ### Common suggestions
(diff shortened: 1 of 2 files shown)


SleeplessByte:
> How about we declare the exercism/r/../example.R files as the ideal solutions and only hyperlink to them in each mentor note?

Please don't make this change. The `example.xxx` are not meant as optimal solutions but only to make the CI pass and to see if there is anything wrong with the tests; and we don't want to set precedent that this is true across languages.

katrinleinweber:
Thanks for providing this perspective. I should have probably written `one exemplary solutions`. It's clear to me that there is no _one_ optimal solution. However, for auto-testing, I'd argue that _computationally efficient_ is probably best.

So, my suggestion morphs into: Should be maintain any `example.xxx` that's also useful to learners in parallel to other code examples in mentor notes.

I understand from your comment that it's probably best to decouple both, though. Even if that means maintaining larger notes with several code examples. Fine with me; Just wanted to clarify :-) If that's the direction, please feel free to close this, which also means https://github.com/exercism/r/pull/115 can be closed.

SleeplessByte:
@iHiD @kytrinyx just pinging you to confirm. I think per the last discussion we had with maintainers vs website-copy is that we *want* these two things to be decoupled.

@katrinleinweber thank you for the detailed response ❤️

kytrinyx:
From my perspective, yes: I would very much like these to be entirely decoupled, with the example solution being optimized for CI, and the mentor notes being something that can (and probably will) evolve based on discussions between mentors, as well as experiences with students.

PR #835, opened 2019-03-04
haskell/collatz-conjecture: use quot and rem over div and mod

b-mehta:
Mistake in collatz: `rem` was being used instead of `quot`.


SleeplessByte:
@rpottsoh are you looking for a second review?

rpottsoh:
> @rpottsoh are you looking for a second review?

No, the changes seem reasonable and simple enough to me.  I'm no Haskell expert, but I did look up `quot`, `rem`, `div` and `mod` as they pertain to the language and I agree with the assessment made in this PR.

PR #839, opened 2019-03-06
r: Add Two Fer mentor notes

katrinleinweber:
I mentored 2 very similar solutions this morning and summarise the advice here to start collecting any for `two-fer`.


katrinleinweber:
You're welcome, and thanks for merging :-)

PR #841, opened 2019-03-06
ruby hamming notes: each_char.with_index vs chars.each_with_index

yyyc514:
Can we all agree this is better as it avoid the need to create the extra array?  This brings solution 3 more in line with solution 2.  In my quick performance tests with small strings solution 3 (with each_char) seems to beat them all.

Or should we add this as a 4th strategy?  It's obviously the correct version of the 3rd strategy though IMHO.

If you want me to add some notes about chars.each_with_index I could do that also.

### DIFF
--- a/tracks/ruby/exercises/hamming/mentoring.md
+++ b/tracks/ruby/exercises/hamming/mentoring.md
@@ -167,7 +167,7 @@ either an empty string or empty regex. Turns out there's already a built-in
 method for splitting a string into it's characters: `String#chars`.
 
 *You don't need to split the string into characters to iterate over it though.*
-That's why Ruby also provides `each_char` (which returns an iterator, not an
+That's why Ruby also provides `each_char` (which returns an Enumerator, not an
 Array).
 
 ## Talking points
--- a/tracks/ruby/exercises/hamming/mentoring.md
+++ b/tracks/ruby/exercises/hamming/mentoring.md
@@ -51,13 +51,13 @@ Strategy 2: String Power
 Pros: uses string directly; no array conversion needed;   
 Cons: requires indices.
 
-Strategy 3: Each with index
+Strategy 3: Each char and with index
 
 ```ruby
-    strand1.chars.each_with_index.count {|letter, index| letter != strand2[index] }
+    strand1.each_char.with_index.count {|letter, index| letter != strand2[index] }
 ```
 Pros: more intuitive? (students seem to pick this often);  
-Cons: requires both indices and converting to arrays, so it's kind of the worst of both worlds from 1 and 2 😛 . 
+Cons: requires indices. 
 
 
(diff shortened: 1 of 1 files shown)


F3PiX:
@exercism/website-copy-ruby

SleeplessByte:
I think this is a good improvement. 

[Here](https://github.com/ruby/ruby/blob/593505ac6f802d2b5bff469425b7c76b65cc9b10/string.c#L8316) is the implementation of [`String#each_char`](https://github.com/ruby/ruby/blob/593505ac6f802d2b5bff469425b7c76b65cc9b10/string.c#L8316) and [here](https://github.com/ruby/ruby/blob/593505ac6f802d2b5bff469425b7c76b65cc9b10/string.c#L8334) is the implementation of [`String#chars`](https://github.com/ruby/ruby/blob/593505ac6f802d2b5bff469425b7c76b65cc9b10/string.c#L8334). You can see that in the link before the call to enumerate in the first case NO array is created and in the second case it is. Therefore, `.each_char` beats `.chars`. That said, I think you should _also_ add the common suggestion to replace `.chars` with `.each_char` with this exact reasoning.

pgaspar:
I like these changes. I wasn't aware of the performance difference.

However, I think you should:

- [ ] Update lines 14-15
- [ ] Add notes acknowledging `chars.each_with_index` is sometimes used and explaining why `each_chars.with_index` is an improvement over `chars.each_with_index` - it may not be immediately obvious

kotp:
@pgaspar Why `each_chars.with_index` and not `chars.each.with_index`?

pgaspar:
> @pgaspar Why each_chars.with_index and not chars.each.with_index?

I think @yyyc514's point is that `chars` forces the creation of an array that isn't necessary if you use enumerators. Right?

SleeplessByte:
@pgaspar that is correct. @kotp see the source links I've posted.

yyyc514:
I'll work on adding suggested changes this week. Thanks.

rpottsoh:
Pinging @yyyc514

sshine:
Pinging @yyyc514 again. :)

yyyc514:
Better?

tracks/ruby/exercises/hamming/mentoring.md
```diff
@@ -165,6 +166,10 @@ Students commonly try to turn a string into characters using `String#split` and
 either an empty string or empty regex. Turns out there's already a built-in
 method for splitting a string into it's characters: `String#chars`.
 
+*You don't need to split the string into characters to iterate over it though.*
+That's why Ruby also provides `each_char` (which returns an iterator, not an
```
kotp:
```suggestion
That's why Ruby also provides `each_char` (which returns an Enumerator, not an
```

PR #846, opened 2019-03-08
python: Notes for mentoring matrix

jeffdparker:
First draft of matrix notes

### DIFF
--- a/tracks/python/exercises/matrix/mentoring.md
+++ b/tracks/python/exercises/matrix/mentoring.md
@@ -1,0 +1,57 @@
+### Problem and challenges
+
+The problem asks you to write a Matrix class.
+The constructor takes as input a string holding 
+lines of integers. The problem asks you
+to write methods row() and column().
+
+### Reasonable Solution
+
+```python
+class Matrix(object):
+
+    def __init__(self, text: str):
+        self.rows = [[int(s) for s in word.split()]
+                     for word in text.split('\n')]
+
+    def row(self, index: int) -> list:
+        return self.rows[index - 1].copy()
+
+    def column(self, index: int) -> list:
+        return [row[index - 1] for row in self.rows]
+```
+
+### Talking Points
+
+#### Delayed contruction
+
+You will see solutions that store the matrix in 
+an intermediate state, and then find the integers
+when row() or column() are called. 
+You will even find many solutions like this in the 
+community area.
+
+We should encourage students to move from strings to integers in __init__. 
+While there are no test cases with non-integers,
+point out that if the input containted bad data,
+you would want the constructor to fail, rather 
+than failing at row() or column()
+
+#### Duplicate storage
+
+You will see solutions that build two copies of 
+the matrix: the original and the transposed matrix.
+Point out that this takes twice the storage, and 
+that constructing a column on the fly isn't that
+difficult.
+
+#### To copy or not to copy
+
+The solution presented copies the rows to avoid
+any chance of updating the matrix. 
+That may or may not be the model the programmer
+has in mind: they may want to take a row and
+update an entry.  
+
+In any case, this isn't a topic to burden a beginner with, 
+but you may want to comment about it if you see it in the solution. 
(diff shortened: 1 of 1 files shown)


tracks/python/exercises/matrix/mentoring.md
```diff
@@ -0,0 +1,57 @@
+### Problem and challenges
```
cmccandless:
```suggestion
### Problem and Challenges
```

tracks/python/exercises/matrix/mentoring.md
```diff
@@ -0,0 +1,57 @@
...
+    def row(self, index: int) -> list:
+        return self.rows[index - 1].copy()
+
+    def column(self, index: int) -> list:
+        return [row[index - 1] for row in self.rows]
+```
+
+### Talking Points
+
+#### Delayed contruction
```
cmccandless:
```suggestion
#### Delayed Contruction
```

tracks/python/exercises/matrix/mentoring.md
```diff
@@ -0,0 +1,57 @@
...
+You will even find many solutions like this in the 
+community area.
+
+We should encourage students to move from strings to integers in __init__. 
+While there are no test cases with non-integers,
+point out that if the input containted bad data,
+you would want the constructor to fail, rather 
+than failing at row() or column()
+
+#### Duplicate storage
```
cmccandless:
```suggestion
#### Duplicate Storage
```

tracks/python/exercises/matrix/mentoring.md
```diff
@@ -0,0 +1,57 @@
...
+
+You will see solutions that build two copies of 
+the matrix: the original and the transposed matrix.
+Point out that this takes twice the storage, and 
+that constructing a column on the fly isn't that
+difficult.
+
+#### To copy or not to copy
```
cmccandless:
```suggestion
#### To Copy or not to Copy
```

tracks/python/exercises/matrix/mentoring.md
```diff
@@ -0,0 +1,57 @@
...
+lines of integers. The problem asks you
+to write methods row() and column().
+
+### Reasonable Solution
+
+```python
+class Matrix(object):
+
+    def __init__(self, text: str):
+        self.rows = [[int(s) for s in word.split()]
```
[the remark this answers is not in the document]
BethanyG: @cmccandless  - .split() in this example is removing whitespace.  The digits in the rows of the matrix string are space delimited. 

@jeffdparker  - strongly recommend that the first Solution  in this section be the one from https://github.com/exercism/python/blob/master/exercises/matrix/example.py, which can then be followed by other variants.   The extemporaneous use of `.copy()` in the current example is a little problematic, IMO.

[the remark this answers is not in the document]
cmccandless: > .split() in this example is removing whitespace. The digits in the rows of the matrix string are space delimited.

:+1: thanks for that; I wasn't looking at the test suite or README when I wrote that, and as I always try to provide an argument for `str.split()`, I forgot the default was `' '`, not `''` (I've been doing more JS lately, which has empty string for its default in string split).

> The extemporaneous use of .copy() in the current example is a little problematic, IMO.

Respectfully, I disagree. I think it is cleaner, despite being an extra step. The test suite does not test for the following case, but in the real world, it would be a concern if a user could modify the original matrix with the following:
```Python
m = Matrix('1 2\n3 4')
m.rows[0][0] = 2
```

Exercism is meant for teaching fluency in a language, and that includes best practices. I would certainly not discourage the use of `.copy()` (I might even praise the foresight of students who include it), but that said, I would not withhold approval from solutions that didn't use it (or equivalent code).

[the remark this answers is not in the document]
BethanyG: The use of `.copy()` in the `rows()` function doesn't prevent modification of  `self.rows` directly in any way.  Granted, it _does_ prevent the caller from inadvertently mutating by:

```python
m = Matrix('1 2\n3 4')
foo = m.rows(0)
foo[0] = 7
```
Which is a good consideration.  I think I am more used to seeing attributes protected via `@property` than via copying.

SleeplessByte:
Thank you @jeffdparker for another great document and kudos @cmccandless for the review 🍔

PR #848, opened 2019-03-09
Add mentoring notes for scala sum-of-multiples

ricemery:
Add mentoring notes for scala sum-of-multiples.

First attempt add adding mentor notes.. Please let me know if anything is messed up.

Thx

### DIFF
--- a/tracks/sum-of-multiples/mentoring.md
+++ b/tracks/sum-of-multiples/mentoring.md
@@ -1,5 +1,6 @@
 ### Reasonable solutions
 
+#### Solution using `filter` and `exists` 
 ```scala
 object SumOfMultiples {
   def sum(factors: Set[Int], limit: Int): Int =
@@ -9,6 +10,7 @@ object SumOfMultiples {
 }
 ```
 
+#### Solution using `flatMap`
 ```scala
 object SumOfMultiples {
   def sum(factors: Set[Int], limit: Int): Int = 
--- a/tracks/sum-of-multiples/mentoring.md
+++ b/tracks/sum-of-multiples/mentoring.md
@@ -1,0 +1,27 @@
+### Reasonable solutions
+
+```scala
+object SumOfMultiples {
+  def sum(factors: Set[Int], limit: Int): Int =
+    (1 until limit)
+      .filter(n => factors.exists(n % _ == 0))
+      .sum
+}
+```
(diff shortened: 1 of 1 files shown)


tracks/sum-of-multiples/mentoring.md
```diff
@@ -0,0 +1,27 @@
...
+    factors flatMap(x => x until limit by x) sum
+}
+```
+
+
+### Common suggestions
+
+- As an additional challenge, this exercise can be solved without using a local mutable var. If instead of using the for comprehension, filter over the (1 until limit) range, then checking if there exists n that are divisible by one of the factors.
+- As an additional challenge, this exercise can be solved without using recursion. If instead of using recursion, 'filter' over the (1 until limit) range, then checking if there 'exists' n that are divisible by one of the factors. 
+- Did you know that this solution does not pass all the unit test cases? Note that most of the test cases are marked pending. And, it is up to you to turn the test cases on as you develop your solution.
```
ErikSchierboom: I'm not sure about this section. Is this something to happens a lot when mentoring, that students have to be told that they have to modify the unit test cases?

tracks/sum-of-multiples/mentoring.md
```diff
@@ -0,0 +1,27 @@
+### Reasonable solutions
+
+```scala
```
ErikSchierboom: Just a suggestion. In the C# track, I use a sub-header to describe the various reasonable solutions if there is more than one (what makes them unique). For an example, see [the notes](https://github.com/exercism/website-copy/blob/master/tracks/csharp/exercises/bob/mentoring.md). You could consider adding those, but feel free to ignore.

tracks/sum-of-multiples/mentoring.md
ricemery: Yes. If I had to estimate, the full suite of unit tests have not been run about 10% of the time.

tracks/sum-of-multiples/mentoring.md
ricemery: Thanks!

tracks/sum-of-multiples/mentoring.md
ErikSchierboom: Oh wow, that's huge!

PR #849, opened 2019-03-09
Add mentoring notes for scala leap

ricemery:
Add mentoring notes for scala leap

### DIFF
--- a/tracks/scala/leap/mentoring.md
+++ b/tracks/scala/leap/mentoring.md
@@ -1,0 +1,25 @@
+### Reasonable solutions
+
+```scala
+object Leap {
+  def leapYear(year: Int): Boolean = {
+    val isYearDivisibleBy: (Int) => Boolean = year % _  == 0
+
+    (isYearDivisibleBy(4) && !isYearDivisibleBy(100)) ||  isYearDivisibleBy(400)
+  }
+}
+```
+
+```scala
+object Leap {
+  def leapYear(year: Int): Boolean = {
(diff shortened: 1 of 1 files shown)


tracks/scala/leap/mentoring.md
```diff
@@ -0,0 +1,25 @@
+### Reasonable solutions
+
+```scala
```
ErikSchierboom: Just a suggestion. In the C# track, I use a sub-header to describe the various reasonable solutions if there is more than one (what makes them unique). For an example, see [the notes](https://github.com/exercism/website-copy/blob/master/tracks/csharp/exercises/bob/mentoring.md). You could consider adding those, but feel free to ignore.

tracks/scala/leap/mentoring.md
```diff
@@ -0,0 +1,25 @@
...
+  def leapYear(year: Int): Boolean = {
+    (year % 4 == 0 && !(year % 100 ==0)) || year % 400 == 0
+  }
+}
+```
+
+
+### Common suggestions
+
+- True and false do not need to be returned explicitly. Instead return the result of combining expressions with &&, || and !
```
ErikSchierboom:
```suggestion
- `true` and `false` do not need to be returned explicitly. Instead return the result of combining expressions with `&&`, `||` and `!`
```

tracks/scala/leap/mentoring.md
ricemery: Thanks!

ErikSchierboom:
Nice!

PR #859, opened 2019-03-11
[ruby/luhn] Create initial mentoring.md

iHiD:


SleeplessByte:
Nice! I would suggest you add some `variations include` statements. Since you're not _really_ using the class to hold state, a module here would be as idiomatic.

kotp:
Let's get it in, and additions such as suggested can be made.  We should get this in view of the mentors as soon as possible. (It encourages them to also submit PR's with suggestions such as @SleeplessByte made above.)

PR #861, opened 2019-03-11
Hyperlink email addresses

katrinleinweber:
Are they intentionally not easily clickable in order to make the drastic actions not too easy?

If not, we could also add a `?subject=` parameter at least to `mentoring@`, because it seems usable for different topics. How about `I%27d%20like%20to%20stop%20mentoring%20the%20track%3A%20` for example?


SleeplessByte:
@katrinleinweber nah, I think the issue comes that GitHub autolinks, and we don't on exercism (with reason, multliple dicussions about this in multiple repos).

Thanks for this improvement 🏅

katrinleinweber:
You're welcome, and thanks for merging :-)

PR #865, opened 2019-03-12
Add mentoring notes for Scala space-age

ricemery:


ricemery:
Thanks @ErikSchierboom !

PR #867, opened 2019-03-12
add me as a js mentor

bdjnk:


rpottsoh:
I think technically your link_url doesn't need the trailing `/`.  If you removed it then link_text would be required to be `null`.  If you made those changes I am not sure how your link_url would appear on the site.  would it be:
[michael.plotke.me](https://michael.plotke.me)
or would it appear like
https://michael.plotke.me/

I am not really concerned.  If you don't want to change it that is fine.  If you decide to change it and don't like how it looks on the site after it is merged then another PR can be opened to tweak the json.

PR #870, opened 2019-03-13
JS & TS: Fix typos in bio for Inkdpixels

Inkdpixels:


mentors/javascript/i.json
```diff
@@ -5,6 +5,6 @@
     "link_text": null,
     "link_url": null,
     "avatar_url": null,
-    "bio": "I love JavaScript for for it's flexibility and fast-paced community. I've written and maintained quite a bit of TypeScript projects, both personally and professionally."
+    "bio": "I love JavaScript for it's flexibility and fast-paced community. I've written and maintained quite a bit of TypeScript projects, both personally and professionally."
```
rpottsoh: Just a suggestion.  Maybe you would like to say something different for the JS track and not something as TS centric.  You are not required to make this change, it is just a suggestion.  If you are content with what is here now just say the word and I'll get this merged.

PR #881, opened 2019-03-16
python/general: Add notes on Returns from functions

jeffdparker:
Multiple returns - Threat or Menace?  and Avoiding the Else Return.

### DIFF
--- a/tracks/python/exercises/General/mentoring.md
+++ b/tracks/python/exercises/General/mentoring.md
@@ -165,3 +165,54 @@ edits.  The single most common example is a trailing
 `pass` left over from the downloaded sourcefile.  
 Leaving the `pass` is a sign that student doesn't
 understand the keyword.  
+
+#### Many Unhappy Returns
+
+There is an rule of thumb that functions should have a single return.
+This is an aid in debugging: you can place a breakpoint on the statements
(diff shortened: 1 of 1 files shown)


tracks/python/exercises/General/mentoring.md
```diff
@@ -165,3 +165,54 @@ edits.  The single most common example is a trailing
...
+```python
+def distance(strand_a, strand_b):
+    if (len(strand_a) != len(strand_b)):
+        raise ValueError('Strands must have the same length')
+    else:
+        strand_a = strand_a.lower()
+        ...
+```
+
+Of course, the else is optional at this point: Fat Elvis
```
cmccandless: I realize this is just a saying, but it's probably best to leave the adjective out here.

```suggestion
Of course, the else is optional at this point: Elvis 
```

jeffdparker: Well, I suppose I deserved this.  The saying is "Elvis has left the building" - I added Fat to show that it wasn't the Elvis anyone wanted.  

I agree that drawing Elvis into the discussion is going to puzzle some of our non-American mentors.

SleeplessByte:
💯 merged

PR #882, opened 2019-03-16
[Ruby/luhn] Fix missing use of attr_reader

iHiD:
I think this change is quite uncontroversial. However, my rational:

- The current mentor notes pass an instance variable around, which I consider a very bad pattern. 
- The notes also say `the minimal solution for approval is a variant with an instantiated class.` I don't feel comfortable teaching someone to instantiate a class if we don't go on to actually use any OOP features. Without using an instance variable in `valid_format?` and `luhn_sum` it would be much more efficient to just use case methods.

### DIFF
--- a/tracks/ruby/exercises/luhn/mentoring.md
+++ b/tracks/ruby/exercises/luhn/mentoring.md
@@ -13,21 +13,21 @@ class Luhn
   end
 
   def valid?
-    return unless valid_format?(candidate)
-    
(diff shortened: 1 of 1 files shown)


iHiD:
@kotp It's `>1` not `>0` so I don't think `empty?` would work.

kotp:
I know you are right.  (Not tired now.)

PR #883, opened 2019-03-17
ruby/raindrops: tweak optimal solution for raindrops

tenebrousedge:
tweak optimal solution by using `zero?`, the placeholder variable `_`, and deferring the call to `values` until the final statement.

### DIFF
--- a/tracks/ruby/exercises/raindrops/mentoring.md
+++ b/tracks/ruby/exercises/raindrops/mentoring.md
@@ -20,8 +20,8 @@ module Raindrops
   SOUNDS = {3 => "Pling", 5 => "Plang", 7 => "Plong"}.freeze
 
   def self.convert(num)
-    rhythm = SOUNDS.select{|key, sound| num % key == 0}.values
-    rhythm.empty? ? num.to_s : rhythm.join
+    rhythm = SOUNDS.select{|key, _| (num % key).zero? }
+    rhythm.empty? ? num.to_s : rhythm.values.join
   end
 end
 ```
--- a/tracks/ruby/exercises/raindrops/mentoring.md
+++ b/tracks/ruby/exercises/raindrops/mentoring.md
@@ -13,8 +13,6 @@ module Raindrops
 end
 ```
 
-This is the optimal solution:
-
 ```ruby
 module Raindrops
   SOUNDS = {3 => "Pling", 5 => "Plang", 7 => "Plong"}.freeze


SleeplessByte:
Looks good to me. `_` might be a nice talking point (unused variables).

Pinging @exercism/website-copy-ruby.

tracks/ruby/exercises/raindrops/mentoring.md
```diff
@@ -20,8 +20,8 @@ module Raindrops
   SOUNDS = {3 => "Pling", 5 => "Plang", 7 => "Plong"}.freeze
 
   def self.convert(num)
-    rhythm = SOUNDS.select{|key, sound| num % key == 0}.values
-    rhythm.empty? ? num.to_s : rhythm.join
+    rhythm = SOUNDS.select{|key, _| (num % key).zero? }
+    rhythm.empty? ? num.to_s : rhythm.values.join
```
kotp: The join should probably be done the block above. No "building" in this line (or as little as possible) as line 23 seems to be where the building aspect belongs.

```ruby
    rhythm = SOUNDS.select{|key, _| (num % key).zero? }.values.join
    rhythm.empty? ? num.to_s : rhythm
```

tenebrousedge: @kotp the `values` and `join` do not need to be performed if the set is empty, and it's not clear to me what design principle(s) are informing your criticism.

tenebrousedge:
@F3PiX The underscore is [handled differently by the Ruby parser](https://github.com/ruby/ruby/blob/trunk/parse.y#L9496). It's not merely a naming convention.

tenebrousedge:
@F3PiX The [Ruby Style Guide](https://github.com/rubocop-hq/ruby-style-guide) also recommends predicate methods like `zero?` over direct comparisons.

SleeplessByte:
> @F3PiX The underscore is [handled differently by the Ruby parser](https://github.com/ruby/ruby/blob/trunk/parse.y#L9496). It's not merely a naming convention.

Whilst this it true, from what I can tell it's an MRI implementation detail. Significant tho, but might not be what we want to teach here? At least at a _talking point_ like I said about this subject. I think it's very interesting and I definitely agree that it should be either `_` or `_sound`, leaning towards `_` because it's more than just idiomatic like you say.

I have no opinion on the `zero?`; You're correct that it's preferred by the default `rubocop`. I think you should put a line under the reasonable solution that both are fine (`== 0` and `zero?`).

tenebrousedge:
@SleeplessByte whether the placeholder should be taught in this exercise is an interesting question. Offhand, I can't think of another exercise where this would be part of an optimal solution.

tracks/ruby/exercises/raindrops/mentoring.md
kotp: I guess my real criticism is that `select` is used and indicated as `optimal`.   And a bit of an aversion to "complicating" the ternary operator branch.  If they were together, rather than split (as the unchanged code is) it may be more apparent of what could be done to be simplified.  Doing a block, and continuing work that should have been done in the block, if possible.

The values and the join should not be done at all, but the building of the string should be done together.  It can be done with `select` but `select` is probably not optimal.  Using each_with_object will build the string directly, iterating those rules, and not needing to decide if we need the values or even to join.

It also avoids creating an Array, to immediate convert it (via join) to a string.

```ruby
# Given a drop of rain (a raindrop as represented by an integer)
# build  a sound (a string, perhaps empty) based on `rules` such 
# as `{3 => 'Pling'}`
    rules.each_with_object('') do |(key, sound), memo|
      memo << sound if key.factor?(drop)
    end
```

This treats the rules without breaking it out (iterating values, for instance) and does the join in the block, only if needed.

It also avoids making the following line, the ternary, more complicated.

- - -

As you can see, hopefully, my criticism really isn't with your change directly, but more of a "we aren't going to notice the problem if we make it slightly better" kind of criticism.  I don't really like the current solution presented using select, I think chains methods that aren't necessary if we use the block to our advantage, and do the work that is needed, rather than piecing it together.

SleeplessByte:
Then I vote yes, add it as a _talking point_ 👍

tracks/ruby/exercises/raindrops/mentoring.md
SleeplessByte: I actually 💯 % prefer the `each_with_object` solution presented by @kotp . I think it's clearer and has way lower cognitive complexity.

SleeplessByte: In the `javascript`  mentor notes I make a habit of listing _both_ versions; one with `select` and one even more optimal. I tend to steer mentors to approve the "lesser yet okay" versions -- with challenges for the student to use something better, in this case `each_with_object`.

kotp: It mutates the string that the `each_with_object` is provided, which is a brand new object in the argument list.  `memo` is that string (object) given in the argument list to `each_with_object`.

kotp: Perhaps a side conversation about `each_with_object`, and why `memo += sound` does not work inside the block is because `+=` returns , the original object remaining the unchanged.  So mutation here is the desired (required) behavior.  Think of it in terms of a collection (of characters) being built.  Like an array collection, we would use `<<` to mutate that array (in Ruby, which is where this note is going), and we are mutating the string we want to build.

In other words, this method is meant to build something, and it does so without creating intermediate objects.  The mutation is fine and expected.

kotp: Approving the PR as the change given the context makes sense for the reasons given, with the stated distaste that I have that this is marked as "optimal" as a solution.  It is reasonable as a solution, but with room for discussion.

PR #888, opened 2019-03-19
julia/difference-of-squares: Use ÷ and Gauss's formula

SaschaMann:


SaschaMann:
Do images work in the mentor notes?

SleeplessByte:
Hi @SaschaMann. Is there a reason the solution is missing? (just like in [custom-set](https://github.com/exercism/website-copy/blob/master/tracks/julia/exercises/custom-set/mentoring.md)).

SaschaMann:
Yes, I don't have a decent example solution for it, but wanted to add the hint anyway.

---
The review request won't work because I'm the only one in that team and I can't approve my own PR.

SleeplessByte:
Can you remove the empty block (`Reasonable Solution`) and preferably add a blank line after each header (see markdown lint).

I'll merge it afterwards.

SaschaMann:
Done, thank you.

PR #889, opened 2019-03-19
python/hamming: add an example using sum()

jeffdparker:


SleeplessByte:
Another great one @jeffdparker!

Thanks for the review @cmccandless .

PR #894, opened 2019-03-20
python/luhn: add example seen in the wild

jeffdparker:
First example I've seen of a map from i to 2*i % 9 unless i == 9

### DIFF
--- a/tracks/python/exercises/luhn/mentoring.md
+++ b/tracks/python/exercises/luhn/mentoring.md
@@ -86,6 +86,16 @@ for n in self.card_num[-2::-2]:
     else:
         every_other_num.append((int(n)*2) % 9)
 ```
+
+Here is an example that uses a tuple of values.  Why this is a default parameter is beyond my ken.  
+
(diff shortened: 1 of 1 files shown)


tracks/python/exercises/luhn/mentoring.md
```diff
@@ -86,6 +86,16 @@ for n in self.card_num[-2::-2]:
     else:
         every_other_num.append((int(n)*2) % 9)
 ```
+
+Here is an example that uses a tuple of values.  Why this is a default parameter is beyond my ken.
```
kotp: Might want to identify what "ken" is or means.  Also, may want to explain why it is beyond "you" (or whomever "my" refers to.  Also, may want to identify the source of the example (even if it is "an example that was found 'in the wild'").  The title of the PR won't carry into the example here, so it will be lost knowledge.

SleeplessByte:
Thanks @jeffdparker for writing and @kotp for reviewing ❤️

PR #903, opened 2019-03-29
WIP: Started mentoring notes for cpp binary_search_tree

elyashiv:
This is a start for mentoring notes for binary_search_tree. I will love to add whatever you think is needed.
I will go over the points that came up in the PR.

@patricksjackson I will love to hear your opinion.

### DIFF
--- a/mentors/julia/s.json
+++ b/mentors/julia/s.json
@@ -6,5 +6,13 @@
     "link_url": "https://www.bromberger.com",
     "avatar_url": null,
     "bio": "I am the founder of JuliaGraphs and one of the core developers of LightGraphs.jl. I've been using Julia at work and for personal projects since v0.3."
+  },
+  {
+    "github_username": "ScottPJones",
+    "name": "Scott Paul Jones",
+    "link_text": "LinkedIn",
+    "link_url": "https://linkedin.com/in/scottgandalf",
+    "avatar_url": null,
+    "bio": "I've been programming for 47 years, always focusing on performance.  I love learning both computer and human languages, my favorites are Scheme, CLU, and now Julia. I respect C as a great tool over the years, now I want to learn Rust to complement Julia,  which I've been using 100% of the time (professionally & developing open source packages) for over 5 years, since version 0.3. I am the creator of the JuliaString.org GitHub organization"
   }
 ]


patricksjackson:
I started some discussion with you in the update PR in exercism/cpp#245.

The two things we need to have here are talking about the iterator and the constructor, and we should try to link to relevant sections in the [CppCoreGuidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md) if they're available.

To support why I link there, those guidelines aren't just some random person doing it, it's led by the creator of the language and has contributions by many of the core committee members. It's generally agreed upon good advice, and helps push for the direction of the language as well. For example, the `span` type I think started in the first iteration of those guidelines, though some companies had similar things in use, and it will be in C++20.

tracks/cpp/exercises/binary-search-tree/mentoring.md
```diff
@@ -0,0 +1,20 @@
+### Reasonable solution ###
```
SleeplessByte: Add empty white line after each heading.

SleeplessByte: Remove trailing `###`

tracks/cpp/exercises/binary-search-tree/mentoring.md
```diff
@@ -0,0 +1,20 @@
...
+There is an example [in the github repo](https://github.com/exercism/cpp/blob/master/exercises/binary-search-tree/example.h) for this track.
```
SleeplessByte: The `example.h` is _not_ the same as the reasonable solution. Remove this and copy the solution here.

tracks/cpp/exercises/binary-search-tree/mentoring.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Common Suggestions ###
+ * Use std::make_unique (c++ 14)
+### Talking Points ###
+ * How to pass parameters to constructor?
+ 
+ method       | arguemnt is lvalue    | argument is rvalue | notes
```
SleeplessByte: Format this as a Github Markdown table, like so:
```
| method       | argument is lvalue | argument is rvalue | notes |
| ------------ | ------------------ | ------------------ | ----- |
| by const ref | copy ctor          | copy ctor          |       |
...
```

It will display like this:

| method       | argument is lvalue | argument is rvalue | notes |
| ------------ | ------------------ | ------------------ | ----- |
| by const ref | copy ctor          | copy ctor          |       |

tracks/cpp/exercises/binary-search-tree/mentoring.md
elyashiv: Hi. I didn't quite understand you... Do you mean the file doesn't contain the solution, or that the solution should be embedded?

I noticed most commentary has the solution embedded in the file, but I think this will be to robust (the file is 227 lines), and the github file can get updates...

SleeplessByte: > that the solution should be embedded?

The solution should be embedded. The example files in each track are for passing the CI, not necessarily a reasonable solution. The issue that keeping them in sync is a lot of overhead for both maintainers and the `website-copy` team.

SleeplessByte: > I noticed most commentary has the solution embedded in the file, but I think this will be to robust (the file is 227 lines)

Personally I don't think that's an issue, but I _understand_ your point. The only other reasonable solution would be to duplicate the file in `cpp` track and rename it, with a contributing note that that file has to be the reasonable solution.

Pinging @exercism/cpp.

KevinWMatthews: If we duplicated the file in the `cpp` track, would the new file be sent to students? I thought that exercism sends all files except for the `example`s.

While 200+ lines is a lot to paste into a `README`, an embedded copy would help keep the reasonable solution in sync with its discussion. I'm concerned that the contents of `example.h` could drift over time, leaving the Talking Points out of date.

@patricksjackson may have a more educated opinion - he's been more involved in reviewing this exercise.

patricksjackson: Well this is interesting. I was very much under the impression that the example solutions in the github repo were used by mentors if they needed to look at a reasonable solution, and have been operating under that assumption for a few _years_ now. I'm fairly sure that assumption is true for all the maintainers in both the C and C++ tracks.

I assumed that track maintainers and track mentors would work together to keep reasonable solutions in the track, not a copy inside of a markdown file _where it would never be tested against new versions of the tests_. So I fail to see why the README file would be more up to date than the example file. It makes a lot more sense to me to have the example be the 'reasonable' solution.

All that said, I've never had time to mentor so I've only been maintaining the track. Obviously we'll defer to you here.

elyashiv: The reasonable solution might not be the one in the repo because:
1. There might be a few reasonable solutions
2. The solution in the repo might be better/worse than the one we expect students to use.
Specifically, in our case, we can approve solutions where the constructor gets a ref to const, or overloaded on lvalue/rvalue refs, but the solution in the repo uses universal refs.

SleeplessByte: What @elyashiv is correct! Apart from that, they are _not_ kept in sync. 

@patricksjackson I've heard that before here and there. Sometimes on slack a mentor links one and usually this happens when the example.xxx file is sub-optimal. People just infer it, but it's not said anywhere.

The example.xxx files are really just there for CI to pass, so that we know the exercise _can_ be solved. But because most maintainers are not part of the website-copy group, we can no longer assume the reasonable solution equals the example.xxx file.

SleeplessByte:
Pinging @elyashiv . Is this still WIP?

sshine:
@SleeplessByte, @elyashiv: In order to progress this merge, can we perhaps say that linking to example solutions in the track repo is less ideal but solved the problem of 227 line header files?

I think mentors of a track should use the mentor notes as they see fit.

SleeplessByte:
@sshine I would prefer it then if they'd rename `example` to `solution`. The example _in all other tracks_ are to make CI pass, and not to be optimal solutions. If this is named differently, that would already indicate it's an actual proper solution. Apart from that, I'm whatevs and they should indeed just do whatever :)

elyashiv:
HI, I'm sorry I was inactive for so long. Can we still merge this? what work needs to be done?

kotp:
> HI, I'm sorry I was inactive for so long. Can we still merge this? what work needs to be done?

I would probably rebase the current master and wring out any problems presented.  And then go from there @elyashiv .

PR #907, opened 2019-03-31
Ruby: Raindrops - reasonable with each_with_object

kotp:
Ref: #883

### DIFF
--- a/tracks/ruby/exercises/raindrops/mentoring.md
+++ b/tracks/ruby/exercises/raindrops/mentoring.md
@@ -1,5 +1,7 @@
 ### Reasonable solutions
 
+- the first solution uses a mutable string
+
 ```ruby
 module Raindrops
   def self.convert(integer)
@@ -12,6 +14,8 @@ module Raindrops
   end
 end
 ```
+
+- using `each_with_object` over the rules object, then add them if they pass a test
 
 ```ruby
 module Raindrops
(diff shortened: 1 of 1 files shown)


SleeplessByte:
I personally like it; perhaps it could helpt to add a line _why_ this is nice? -- there is a conflict.

kotp:
> there is a conflict.

Introduced due to the other being merged into master, taken care of.

As far as a comment as to why is this nice?  Can a mentor identify that via the code?  There are several aspects in the code that aren't relevant to the problem itself, such as the vertical layout of the rules, the mutable allow indicator of the string literal which is only valid on some versions of Ruby, etc.

SleeplessByte:
Gotcha! Good question.

Basically I am saying, perhaps you want to
- add at line 2 or line 15 that the first solution uses a mutable string
- add at line 15 or line 32 that the second solution uses `each_with_object` over the rules object, then add them if they pass a test
- add at line 32 or line 43 that the third solution selects the correct rules by testing them and then joining the selected rules

I know it's apparent from the code, but writing it out might help a mentor as the cognitive complexion is way lower that way. As you said yourself: many things aren't relevant and adding a line of clarification _what is relevant_ goes a long way.

SleeplessByte:
Thanks @kotp ! Great addition.

PR #908, opened 2019-03-31
Add 0Thomas to Python Mentors

0thomas:


0thomas:
> Your information looks fine, I just have a minor formatting request. Please increase indentation by two spaces on lines 2 through 9.

Added, thanks for the request, it seems I completely missed this point. Maybe we can add it to the Readme.

rpottsoh:
I think the [README](https://github.com/exercism/website-copy/tree/master/mentors) covers it, but there may be room for improvement.  If you can think of another way/better way I certainly would appreciate a PR.  You are not the first to run into this situation.  The first to post their bio to a file is the most vulnerable, without an example to emulate.

Thanks for making  the changes here to the JSON.  I'll be merging now.

PR #910, opened 2019-04-01
Rust: add mentoring notes for reverse-string

TheDarkula:
With regards to the turbofish operator  (`::<>`), I'm aware that it can be elided and the compiler can fill in the blanks for us. The goal is to assume students know nothing at every pass so we can teach them all of the idiosyncrasies of the language. I have personally mentored over 70 students on "Reverse String", since we promoted it to be a core exercise, and the vast majority of them hadn't even heard of the turbofish operator; they have also thanked me profusely for teaching them about it.

The best approach to really showing students what they can do with Rust is by being explicit. If they want to omit what some would regard as "unnecessary" typing in their work after exercism, that's fine. But, our goal should be to give them all of the information up front, and they can decide their own style afterwards. I feel that more information creates self-documenting code and is a great practise to follow :)

### DIFF
--- a/tracks/rust/exercises/reverse-string/mentoring.md
+++ b/tracks/rust/exercises/reverse-string/mentoring.md
@@ -1,15 +1,17 @@
 ### Concepts
 
-- &str
-- String
-- Iterators
-- Iterator Methods
+- [&str](https://doc.rust-lang.org/std/primitive.str.html)
+- [String](https://doc.rust-lang.org/std/string/struct.String.html)
+- [Iterators](https://doc.rust-lang.org/std/iter/trait.Iterator.html)
+- [Iterator Methods](https://doc.rust-lang.org/std/iter/trait.Iterator.html#provided-methods)
 
 ### Reasonable solutions
 
 A reasonable solution should:
 
-- Use `.chars()`, `.rev()`, and `.collect()`
+- Use [.chars()](https://doc.rust-lang.org/std/primitive.str.html#method.chars),
+[.rev()](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.rev), and
+[.collect()](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect)
 - Be explicit with `.collect()` by using the turbofish operator `::<>`
 - Not use other iterators or iterator methods like `.into_iter()` or
 `.flat_map()`
@@ -18,11 +20,12 @@ A reasonable solution should:
 
 A bonus solution should:
 
-- Use the `unicode_segmentation` crate
+- Use the [unicode_segmentation crate](https://unicode-rs.github.io/unicode-segmentation/unicode_segmentation/index.html)
 - Use `input.graphemes(true)` instead of
-`UnicodeSegmentation::graphemes(input, true)`
+`UnicodeSegmentation::graphemes(input, true)`.
+The documentation for this method can be found [here](https://unicode-rs.github.io/unicode-segmentation/unicode_segmentation/trait.UnicodeSegmentation.html#tymethod.graphemes).
 
-### Examples
+#### Examples
 
 ```rust
 pub fn reverse(input: &str) -> String {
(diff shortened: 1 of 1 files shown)


tracks/rust/exercises/reverse-string/mentoring.md
```diff
@@ -0,0 +1,111 @@
...
+- Not use variables or loops
+- Not use `return`
+
+A bonus solution should:
+
+- Use the `unicode_segmentation` crate
+- Use `input.graphemes(true)` instead of
+`UnicodeSegmentation::graphemes(input, true)`
+
+### Examples
```
SleeplessByte: Turn into level 4 heading (because these are examples of reasonable solutions)

tracks/rust/exercises/reverse-string/mentoring.md
```diff
@@ -0,0 +1,111 @@
...
+
+```rust
+pub fn reverse(input: &str) -> String {
+    input.chars().rev().collect::<String>()
+}
+```
+
+##### Bonus solution:
```
SleeplessByte: Turn into level 4 heading as it's an example of the bonus solution which is reasonable

tracks/rust/exercises/reverse-string/mentoring.md
```diff
@@ -0,0 +1,111 @@
...
+
+```rust
+use unicode_segmentation::UnicodeSegmentation;
+
+pub fn reverse(input: &str) -> String {
+    input.graphemes(true).rev().collect::<String>()
+}
+```
+
+### Example Comments
```
SleeplessByte: Change to `### Common Suggestions`

tracks/rust/exercises/reverse-string/mentoring.md
```diff
@@ -0,0 +1,111 @@
...
+We don't need to use the `return` keyword here, we only use that for early returns.
+```
+
+If they haven't done the bonus:
+
+```
+Do you want to give the bonus a go? :)
+```
+
+##### For the bonus:
```
SleeplessByte: Make this level 4 (common suggestions for the bonus).

tracks/rust/exercises/reverse-string/mentoring.md
```diff
@@ -0,0 +1,111 @@
+### Concepts
+
+- &str
```
SleeplessByte: You might want to consider linking to the official docs, so mentors have a quick reference. This goes for all these and all the method calls in this document.

tracks/rust/exercises/reverse-string/mentoring.md
TheDarkula: Done :)

tracks/rust/exercises/reverse-string/mentoring.md
TheDarkula: Done :)

tracks/rust/exercises/reverse-string/mentoring.md
TheDarkula: Done :)

tracks/rust/exercises/reverse-string/mentoring.md
TheDarkula: Done :)

tracks/rust/exercises/reverse-string/mentoring.md
TheDarkula: Great idea! I only linked to the methods that we want mentors/students to use :)

TheDarkula:
@SleeplessByte Thanks so much for the great review!

SleeplessByte:
Perfect @TheDarkula! Do you perhaps want to wait on a review by @coriolinus? I know he's the other one on @exercism/website-copy-rust. If not, I'm okay with merging this instead of waiting and making subsequent changes on a new PR, as having some notes are better than no notes.

TheDarkula:
@SleeplessByte I agree with merging this now since it is a core exercise and we have no notes currently. Thanks again for the help and approving this :)

@coriolinus If there are any improvements you can think of, let me know in the #track-rust channel.

Merged :)

PR #917, opened 2019-04-04
python/high-scores: Minor Change to 3rd possible Solution.

0thomas:
Solution also works out without the first -1. Saw it first in a solution I am mentoring.

Passing all the tests without any issues.

### DIFF
--- a/tracks/python/exercises/high-scores/mentoring.md
+++ b/tracks/python/exercises/high-scores/mentoring.md
@@ -59,8 +59,11 @@ class HighScores(object):
     def personal_top_three(self):
         return self.sorted_scores[:-4:-1]
 ```
-
-
+Another variation of this solution includes the start at -1 for slicing the result to personal_top_three. The last solution works due the step being -1. Since no start is defined the start will default to 0 and the first step will start at -1 due to the step being defined.
+```python
+    def personal_top_three(self):
+        return self.sorted_scores[-1:-4:-1]
+```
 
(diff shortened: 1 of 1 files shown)


tracks/python/exercises/high-scores/mentoring.md
```diff
@@ -57,7 +57,7 @@ class HighScores(object):
         return self.sorted_scores[-1]
 
     def personal_top_three(self):
-        return self.sorted_scores[-1:-4:-1]
+        return self.sorted_scores[:-4:-1]
```
SleeplessByte: Below this solution you probably want to mention that variation include:

```python
return self.sorted_scores[-1:-4:-1]
```

Might also be worthwhile to explain _why_ it works.

SleeplessByte: @0thomas can you make this change?

0thomas: @SleeplessByte Yes, sure. Was a bit busy. Here comes the new commit. Suggest any words you would like to change in the reason. I tried to express it as clearly and as simple as possible.

SleeplessByte: No problem! 

I think the new wording is just fine. We can always change in a future PR.

SleeplessByte:
Thank you @0thomas :heart: (and @cmccandless for the review)

PR #920, opened 2019-04-05
scala: add mentoring notes for perfect-numbers

ricemery:


tracks/scala/exercises/perfect-numbers/mentoring.md
```diff
@@ -0,0 +1,59 @@
+### Reasonable solutions
+
+#### Solution using case objects
+
+```scala
+object PerfectNumbers {
+  import NumberType._
+
+  def classify(number: Int): NumberType = {
```
ErikSchierboom: The return type is `NumberType`, whereas the next solution's `classify` method's return type is `: Either[String, NumberType.Value]`. Is that correct?

ricemery: Good catch. I pulled an old solution as an example that needs to be updated to the new signature/unit tests.

ErikSchierboom:
Merged!

PR #921, opened 2019-04-05
csharp/two-fer: update to latest spec

ErikSchierboom:


ErikSchierboom:
Thanks for reviewing!

PR #924, opened 2019-04-05
python/isogram: Add example seen in wild

jeffdparker:
I discussed sorting, but didn't have an example


tracks/python/exercises/isogram/mentoring.md
```diff
@@ -14,8 +14,8 @@ def is_isogram(string):
         ch = string[i]
 
         if ch.isalpha():
-		    if (s.find(ch) > -1):
-			    return False
+	    if (s.find(ch) > -1):
+	        return False
```
cmccandless: This indentation should not have been reduced.

sshine:
Ping @jeffdparker

tracks/python/exercises/isogram/mentoring.md
jeffdparker: Agreed.  

Sorry - I have been away from the keyboard for most of a month.

NobbZ:
I'm not sure if sorting and then searching for duplicates is a solution that should be mentioned at all.

You can not sort "better" than `O(n log n)`, searching for duplicates is another `O(n)`.

Just iterating and accumulating a bitset keeping the already seen letters is `O(n)`.

In my opinion, the final goal of this exercise should always be a solution that has `O(n)` time.

jeffdparker:
> I'm not sure if sorting and then searching for duplicates is a solution that should be mentioned at all.
> 
> You can not sort "better" than `O(n log n)`, searching for duplicates is another `O(n)`.
> 
> Just iterating and accumulating a bitset keeping the already seen letters is `O(n)`.
> 
> In my opinion, the final goal of this exercise should always be a solution that has `O(n)` time.

I'm happy to take a suggested update.  Others may want to chime in, but I have a two questions.

Do all of the problems have a single right solution?  (AKA Is there a best bicycle?  Mountain?  Road?  Hauling Groceries?)

Which is more important, space or time? 

I agree that O(n) is better than O(n log n),  but sometimes implementation details are more important than theory.  In this case, n = 26, so sorting is pretty quick, but a general set implementation may be optimized with a Hash table for much largest N.  

Would you like to suggest prose that weighs time and space for this solution?  Perhaps something to nudge the programmer to a solution that uses sets?

Jeff Parker

NobbZ:
Personally, I'm strictly against anything that uses a set that needs more than 32 bit of space (or it next bigger integer size that is idiomatic in a language).

I do see though, that some more inefficient sets are sometimes necessary for the student to understand the inital concept of using a set at all for this exercise.

But sorting the text of the bible to search for duplicate letters, is a waste of time.

By simply reducing over the text and having the set as accumulator you can probably short circuit after the first couple of verses…

No discussion necessary…

jeffdparker:
> Personally, I'm strictly against anything that uses a set that needs more than 32 bit of space (or it next bigger integer size that is idiomatic in a language).
> 
> I do see though, that some more inefficient sets are sometimes necessary for the student to understand the inital concept of using a set at all for this exercise.
> 
> But sorting the text of the bible to search for duplicate letters, is a waste of time.
> 
> By simply reducing over the text and having the set as accumulator you can probably short circuit after the first couple of verses…
> 
> No discussion about space vs time necessary…



> Personally, I'm strictly against anything that uses a set that needs more than 32 bit of space (or it next bigger integer size that is idiomatic in a language).
> 
> I do see though, that some more inefficient sets are sometimes necessary for the student to understand the inital concept of using a set at all for this exercise.
> 
> But sorting the text of the bible to search for duplicate letters, is a waste of time.
> 
> By simply reducing over the text and having the set as accumulator you can probably short circuit after the first couple of verses…
> 
> No discussion about space vs time necessary…

Forgive my stupidity - you are right that it is length of text, not size of the set, that counts.  A bucket sort with be efficient, but we can't assume that any more than we can assume that Python is going to use a 32 bit word to store a set representing the 26 letters.  

Do you think we should add an example that uses a word and bit manipulations to solve the problem?  Please feel free to suggest some.  

Again, if you have text you'd like to suggest for the sorting section, I'd be pleased to take it.  We are going to see solutions that use sorting: we can decide what we want to say when we see them.  

This problem can be solved many ways.  We will need to address the solutions we are presented with.

NobbZ:
I'd not even mention sorting solutions.

If at all, I'd add something like "if the input is sorted first, tell the student about the bible as input."

jeffdparker:
> I'd not even mention sorting solutions.
> 
> If at all, I'd add something like "if the input is sorted first, tell the student about the bible as input."

Of course, with Python 3, we can no longer assume that a character fits into 8 bits much less 5.  

I can add some text describing the drawbacks to sorting, mention the coding of a set in 32 bits, and mention the limits of that approach given Unicode characters.

jeffdparker:
Sorry to be so long out of the loop.  I hope to return to help out some day, but I'm currently swamped.

If I'm preventing someone else form taking this on, please let me know.

jdp

sshine:
@jeffdparker: No problem, just writing in the interest of merging or closing PRs.

I've asked the Python track Slack if anyone is interested in pushing to this feature branch.

This repository sees a bit of abandoned PRs, so I'm thinking about a protocol to avoid that.

Let's keep the PR open for now.

yawpitch:
Much of what's in this file makes no sense.

The "Reasonable Solution" isn't at all reasonable ... for an N-char string it will allocate N individual strings then search each of them in O(n) time ... its pathological.

Sorting isn't reasonable, for all the reasons mentioned above.

The section on trimming a string seems nonsensical... all that's needed is list of the legal characters, there's no reason to use either string concatenation or even mention str.join.

The introduction of list comprehensions is fine, but also suggests joining for no reason.

The Counter example would throw an AttributeError because the **string** module doesn't have a **lower** function. Fix that and it could be written _considerably_ more sensibly.

The **set** version is the closest to idiomatic, but also could use a rewrite.

And the **str.count** approach is going to search the entire string for every character in the string ... it's also pathological.

Honestly this probably needs to be completely rewritten.

yawpitch:
A bitset, BTW, is only a good idea if we assume the character set is ASCII only, or some other short-ish alphabet ... it's going to fall apart as soon as there's more than 64 characters in the alphabet.

yawpitch:
The "canonical" data structure for this problem is a **set**, BTW, in Python ... the **Counter** is a waste of both space and effort, as the counts are irrelevant.

The "simplest" reasonable approach is to compare the length of the set of the valid letters to the length of the list of those letters, but the "most reasonable" solution is something like the "unique ever seen" walk that @sshine mentioned above.

BethanyG:
I am going to close this due to inactivity/abandonment.  If anyone would like to revisit, please let me know.

PR #944, opened 2019-04-17
python/hamming: Include a recursive example

jeffdparker:
Discourage the use of tail recursion here.

### DIFF
--- a/tracks/python/exercises/hamming/mentoring.md
+++ b/tracks/python/exercises/hamming/mentoring.md
@@ -63,6 +63,22 @@ def distance(strand_a, strand_b):
     return sum(a != b for a,b in zip(strand_a,strand_b))
 ```
 
+#### Recursion
+
+It is possible to write a recursive solution.
+
+```python
+def distance(strand_a, strand_b):
+    if len(strand_a) != len(strand_b):
+        raise ValueError("Different length")
+    elif len(strand_a) == 0:
+        return 0
+    else:
+        return (0 if strand_a[0] == strand_b[0] else 1) + distance(strand_a[1:], strand_b[1:])
+```
+
+This take more time and more stack space than an iterative solution, and should be discouraged.
+
 ### Common Suggestions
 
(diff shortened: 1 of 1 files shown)


NobbZ:
That version is not tail recursive, but yes python does not support recursive functions very well and a loop should always be preferred over recursion.

NobbZ:
I'm also not sure if this should be listed as "reasonable" solution with the remark to discourage it.

jeffdparker:
It's not Tail Recursive?  Really?

NobbZ:
Yes, it is not tail recursive, as tail recursiveness means that the last function call is to the function itself, but in your case, the last function called is `+`.

In FP it is common to create a helper function which carries an additional accumulating value or to use some kind of `reduce`/`fold`.

NobbZ:
You are right on the notes not mentioning tail recursiveness, still I think it's incorrect to have the example as a reasonable solution with a comment discouraging it at the same time.

cmccandless:
> still I think it's incorrect to have the example as a reasonable solution with a comment discouraging it at the same time.

Maybe it would be best to reword that comment, along the lines of

`This solution works, but will take more time and more stack space than an iterative solution. Encourage the student to think in terms of performance as well as correctness.`

tracks/python/exercises/hamming/mentoring.md
```diff
@@ -63,10 +63,31 @@ def distance(strand_a, strand_b):
...
+Encourage the student to think in terms of performance as well as correctness.
+
 ### Common Suggestions
 
 A List Comprehension can be suggested, but isn't an obvious win.
 	 
 ### Talking points
 
 Running time is not a concern here. It is hard to make this worse than O(N).
+
```
cmccandless: This extra trailing whitespace is unnecessary.

cmccandless: I'm referring to the added 2 newlines at the end of the file.

PR #950, opened 2019-04-21
python/tournament: mentoring.md

jeffdparker:
A first draft describing the problem

### DIFF
--- a/tracks/python/exercises/tournament/mentoring.md
+++ b/tracks/python/exercises/tournament/mentoring.md
@@ -1,0 +1,120 @@
+### Problem and challenges
+
+You are given the results of a series of games, and
+asked to print the results for the season.
+
+You need to parse alternative ways to describe a game,
+keep track of data, and print the data in a fixed form.
+The data must be sorted by points and team names.
+
+### Reasonable solutions
+
+```python
+import collections 
+
+def defaultValue():
+    return [0, 0, 0]
+
+def points(standings: dict, team: str) -> int:
+    return 3*standings[team][0] + standings[team][1]
+
+def tally(results):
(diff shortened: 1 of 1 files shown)


tracks/python/exercises/tournament/mentoring.md
```diff
@@ -0,0 +1,120 @@
...
+    lst = sorted(lst, key=itemgetter(0))
+    lst = sorted(lst, key=itemgetter(1), reverse=True)
+```
+
+We then sort by names, followed by sorting by value.
+Python's sorts are stable: that is, two teams with
+the same points will remain in the same relative order.
+
+Another approach would be to build a list of tuples,
+and use the natural odering for tuples to simplify the sort.
```
cmccandless:
```suggestion
and use the natural ordering for tuples to simplify the sort.
```

tracks/python/exercises/tournament/mentoring.md
```diff
@@ -0,0 +1,120 @@
...
+However, we want the teams in descending order of points,
+and then in ascending order by name.
+We handle that below by negating the points.
+
+```python
+    lst = [(-points(team), team) for team in standings]
+    def by_points_and_name(self):
+        return (-self.points(), self.name)
```
cmccandless: Not sure if I'm just missing something, or if there is a mistake here. Shouldn't this be something like the following?

```
def by_points_and_name(self):
    return (-self.points(), self.name)

lst = sorted(standings, key=lambda t: t.by_points_and_name())
```

If I'm missing the point, please defend your current iteration.

BethanyG:
Closing this since it's had no activity since early 2019.  @jeffdparker  -- please give a shout if you'd like to continue, and I can re-open this.  Thanks!

PR #952, opened 2019-04-22
Rust Clock mentoring notes

ExpHP:
cc @TheDarkula

### DIFF
--- a/tracks/rust/exercises/clock/mentoring.md
+++ b/tracks/rust/exercises/clock/mentoring.md
@@ -7,7 +7,14 @@
 ### Reasonable solutions
 
 A reasonable solution should do the following:
-`items needed here`
+
+* Use `#[derive]`
+* Implement `fmt::Display`
+* Reuse logic by implementing `Clock::add_minutes` in terms of `Clock::new`, or vice versa.
+* Prefer `24 * 60` over the rather magical `1440`.
+* Use `%` instead of a loop.
+* Have clean logic for performing the modulus.
+* (more items needed here)
 
 ### Examples
 
@@ -40,3 +47,13 @@ impl fmt::Display for Clock {
     }
 }
 ```
+
+### Notes
+
+If somebody visibly laments the behavior of `%` on negative arguments, you can tell them how to use unstable features like `rem_euclid`.
+
+"Clean logic" for the modulus ideally means that there is usage of a function that returns a positive modulus.  (basically an implementation of `rem_floor`, or usage of the unstable `rem_euclid`).  Notice that, if the clock stores hours in addition to minutes, the corresponding `div_floor` or `div_euclid` are also necessary.
+
+If somebody already has such functions in their code, you can ask them to take it a step above and add a test for it.  (this is a good opportunity to show that tests can be in `src/lib.rs` where they can see private functions, rather than always having to be in `tests/`)
+
+It is fine if the clock stores both hours and minutes, but if somebody's logic is hopelessly complicated, suggest that it might be easier to only store `minutes`.
--- a/tracks/rust/exercises/clock/mentoring.md
+++ b/tracks/rust/exercises/clock/mentoring.md
@@ -95,7 +95,7 @@ impl fmt::Display for Clock {
 }
 ```
 
-### Notes
+### Common Suggestions
 
 If they implement Debug or PartialEq manually:
 
--- a/tracks/rust/exercises/clock/mentoring.md
+++ b/tracks/rust/exercises/clock/mentoring.md
@@ -101,26 +101,35 @@ impl fmt::Display for Clock {
 
 If they implement Debug or PartialEq manually:
 
-```
-We can use the `#[derive]` attribute on the `Clock` struct to automatically generate impls of [Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html) and [PartialEq](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html#derivable).
+```markdown
+We can use the `#[derive]` attribute on the `Clock` struct to automatically
+generate impls of [Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html)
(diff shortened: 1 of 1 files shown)


tracks/rust/exercises/clock/mentoring.md
```diff
@@ -7,7 +7,14 @@
 ### Reasonable solutions
 
 A reasonable solution should do the following:
```
TheDarkula: Let's amend it to say:

> A reasonable solution should:
> 
> 	- Use `#[derive]`
> 	...

Also, I feel it reads most straightforward to use `1440` as the value for the constant, because the variable name `MINUTES_PER_DAY` tells us exactly what we're looking at. Also, having it set to a plain integer is "simpler" because there are no maths that need to be evaluated.

tracks/rust/exercises/clock/mentoring.md
```diff
@@ -7,7 +7,14 @@
...
+
+* Use `#[derive]`
+* Implement `fmt::Display`
+* Reuse logic by implementing `Clock::add_minutes` in terms of `Clock::new`, or vice versa.
+* Prefer `24 * 60` over the rather magical `1440`.
+* Use `%` instead of a loop.
+* Have clean logic for performing the modulus.
+* (more items needed here)
 
 ### Examples
```
TheDarkula: We need to change ### Examples to #### Examples since it's a subsection of "Reasonable solutions".

With that, I would prefer the constant be `MINUTES_PER_DAY` in the example, just because it reads clearer.

tracks/rust/exercises/clock/mentoring.md
```diff
@@ -40,3 +47,13 @@ impl fmt::Display for Clock {
     }
 }
 ```
+
+### Notes
```
TheDarkula: Next, can you update the "Notes" section to be the "Common Suggestions" section?

For example:

### Common Suggestions

If they implement Debug or PartialEq manually:
```
We can use the `#[derive]` macro for the `Clock` struct instead of implementing features like [Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html) and [PartialEq](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html#derivable) manually.
```

If they visibly lament the behavior of % on negative arguments:
```
We can use unstable features like [.rem_euclid()](https://doc.rust-lang.org/std/primitive.i32.html#method.rem_euclid).
```

If they already have such functions in their code:
```
We can take it a step above and add a test for it.
```
(this is a good opportunity to show that tests can be in src/lib.rs where they can see private functions, rather than always having to be in tests/)

It is fine if the clock stores both hours and minutes, but if their logic is hopelessly complicated:
```
It might be easier to only store minutes :)

File: tracks/rust/exercises/clock/mentoring.md [hunk 4]
```diff
@@ -40,3 +47,13 @@ impl fmt::Display for Clock {
...
+
+If somebody visibly laments the behavior of `%` on negative arguments, you can tell them how to use unstable features like `rem_euclid`.
+
+"Clean logic" for the modulus ideally means that there is usage of a function that returns a positive modulus.  (basically an implementation of `rem_floor`, or usage of the unstable `rem_euclid`).  Notice that, if the clock stores hours in addition to minutes, the corresponding `div_floor` or `div_euclid` are also necessary.
```
TheDarkula: I feel that a code example here would be best, instead of elaborating this "verbally". Also, providing a sample response to give to the student, in this case, would be especially helpful.

tracks/rust/exercises/clock/mentoring.md
ExpHP: I think the suggestions look a bit strange in code fences, especially without line breaks.  Could I suggest:

#### Blockquote with escapes

If they implement Debug or PartialEq manually:

> We can use the \`#[derive]\` macro for the \`Clock\` struct instead of implementing features like \[Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html) and \[PartialEq](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html#derivable) manually.

or maybe 

#### Code with manual linebreaks and markdown highlighting

If they implement Debug or PartialEq manually:

```markdown
We can use the `#[derive]` macro for the `Clock` struct instead of implementing
features like [Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html) and
[PartialEq](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html#derivable) manually.
```

---

(or maybe this can be revisited in a later PR that affects numerous exercises)

TheDarkula: I think the format is so it's easy to just copy and paste from inside the code block to respond to students, but whatever works :)

ExpHP:
Heh, To be honest I dislike the canned responses on some of the mentor notes since it invites copying-and-pasting, when it is usually necessary to adjust the phrasing on an individual basis (based on what the student appears to already know, or in order to work in complements).  But I'll oblige.

There are now two examples, that differ in two ways:

* just minutes, versus hours and minutes
* a manual `rem_floor` function versus the unstable feature

I added two more bullets to "Reasonable solutions" to complement some of the suggestions.

ExpHP:
Oh, I missed the bit about the Concepts section.

tracks/rust/exercises/clock/mentoring.md
ExpHP: In that case I'll add line breaks :D

TheDarkula:
Awesome, thanks again!

I'm in no way saying to only use the copy/paste "canned" responses. It just helps having them in the mentoring notes to get started with feedback, especially when the same issue keeps coming up and you keep typing the same thing over and over again. I tend to mix and match the notes as necessary, and of course, add compliments :)
I very much encourage and want mentors to add their own flavour!

For the "Concepts" section, be sure to add the links for each one as a quick and easy reference.

TheDarkula:
Thanks so much! :)

PR #954, opened 2019-04-23
Elixir: Update `link_name` from Sascha Wolf

Zeeker:
So, I've noticed that providing a `link_url` but not a `link_name` leads to this funny situation: 

image

As you can see the link text points to github but the URL to my website. I've expected that the `link_text` would simply equal the `link_url`. In my original PR (#953) I've tried to set the `link_name` to `saschawolf.me` but that was disallowed by the tests, as such I now chose `Website`.

### DIFF
--- a/mentors/elixir/z.json
+++ b/mentors/elixir/z.json
@@ -10,7 +10,7 @@
   {
     "github_username": "zeeker",
     "name": "Sascha Wolf",
-    "link_text": "saschawolf.me",
+    "link_text": null,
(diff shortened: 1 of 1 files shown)


rpottsoh:
Cc @ihid

rpottsoh:
@zeeker, this is strange, let's see what Jeremy has to say about this.

iHiD:
Thanks. Will be fixed for future updates by https://github.com/exercism/website/pull/628 once I merge/deploy.

PR #961, opened 2019-04-24
Create readme.md for Pharo-Smalltalk

macta:
give mentors some guidance on where to find pharo mentor notes

### DIFF
--- a/tracks/pharo-smalltalk/exercises/readme.md
+++ b/tracks/pharo-smalltalk/exercises/readme.md
@@ -1,0 +1,5 @@
+## Notes
+
+Unlike other tracks, Pharo has an IDE built in itself, and so mentoring tools have been built into Pharo to retrieve user exercises and create feedback notes.
+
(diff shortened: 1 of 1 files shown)


iHiD:
>  I have no issues with a readme here but don't know if the site will have issues.

It should be fine. But thank you for checking @rpottsoh.

I'm :+1: on this PR as Pharo is so different to everything else here. But I'll let someone in the website-copy team make the final decision :)

macta:
I neglected to follow up on the question (holiday).

Currently - my proposal is that Pharo mentors load a development image (like a docker or VMware image) which has the reference examples. It’s an easy load, and it gives a mentor browser in a menu option in that environment (Pharo supports the big 3, so we are cross platform).

I found navigating code on the web a bit tricky for more complex exercises as I wanted to see references to variables, or senders or implementors of a method. 

I’m also tuning the ability to comment on methods or classes to give easier feedback to paste into the web portal.

If it turns out we need both, then will revert this change and add the feedback we have accumulated.

PR #973, opened 2019-04-28
add initial mentor notes for Crystal/Bob

tenebrousedge:

### DIFF
--- a/tracks/crystal/exercises/bob/mentoring.md
+++ b/tracks/crystal/exercises/bob/mentoring.md
@@ -1,0 +1,74 @@
+### Algorithm
+
+There are two general approaches, either using regular expressions or `String` methods.
+```crystal
+class Bob
+  getter comment
+  
+  def self.hey(comment : String) : String
+    self.new(comment).response
+  end
+
+  def initialize(@comment : String) : String
+  end
+
+  def response : String
+    if shouting? && questioning?
+      "Calm down, I know what I'm doing!"
+    elsif shouting?
+      "Whoa, chill out!"
+    elsif questioning?
+      "Sure."
+    elsif silent?
+      "Fine. Be that way!"
+    else
+      "Whatever."
+    end
+  end
+
+  private def shouting?
+    comment.each_char.any?(&.ascii_letter?) && comment == comment.upcase
+  end
+
+  private def questioning?
+    comment.ends_with?("?")
+  end
+
+  private def silent?
+    comment.blank?
+  end
+end
+
+```
+```crystal
+module Bob
+  SHOUTING = "\\A(?=.*[A-Z])(?!.*[a-z])"
+
+  QUESTIONING = ".*\\?\\z"
+
+  SILENT = "\\A\\s*\\z"
+
+  def self.hey(comment : String) : String
+    case comment
+      when %r{#{SHOUTING + QUESTIONING}}
+        "Calm down, I know what I'm doing!"
+      when %r{#{SHOUTING}}
+        "Whoa, chill out!"
+      when %r{#{QUESTIONING}}
+        "Sure."
+      when %r{#{SILENT}}
+        "Fine. Be that way!"
+      else
+        "Whatever."
+      end
+  end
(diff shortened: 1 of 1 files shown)


tracks/crystal/exercises/bob/mentoring.md
```diff
@@ -0,0 +1,74 @@
...
+  
+  def self.hey(comment : String) : String
+    self.new(comment).response
+  end
+
+  def initialize(@comment : String) : String
+  end
+
+  def response : String
+    if shouting? && questioning?
```
SleeplessByte: I don't know if this is _the_ way to do it in crystal, but in most other languages, one should not test for shouting or question _twice_. e.g. it should be a clever match or nested conditional.

tenebrousedge: Well, I admit to some stupidity on this point, but I don't know how one would reorder or nest the conditions to avoid testing any condition twice.

SleeplessByte: Given your code:

```crystal
  def response : String
    if shouting?
      if questioning?
        "Calm down, I know what I'm doing!"
      else
        "Whoa, chill out!"
      end
    elsif questioning?
      if shouting?
        "Sure."
       else
        "Whoa, chill out!"
      end
    elsif silent?
      "Fine. Be that way!"
    else
      "Whatever."
    end
  end
```

But now it isn't very DRY; so perhaps a _better_ way would to keep what you have but _memoize_ `shouting?` and `questioning?`. I think this is _why_ the test case was added, to explore a discussion like this. I don't know what the canonical way of solving this would be in Crystal -- it might even be "values are cached by the compiler / stack frame / whatevs so therefore we don't need to worry about this".

Naturally it is a micro-optimisation in terms of THIS exercise, but replace any of the methods with `expensive_call?` and then it becomes relevant.

tenebrousedge: Ah, I see, this solution would avoid testing the conditions, but duplicate the responses.

I don't think that discussions regarding performance are very useful in the context of Exercism problems, since there are no limits on the expected size of any of the inputs. One can add memoization to almost every value generated in every problem, and as you say, the compiler may do that automatically. Making this explicit simply adds "noise" to the solutions. However, I believe that the stated reason for adding the test case was that the condition of a string that was both "shouting" and "questioning" was previously ambiguous.

SleeplessByte: > However, I believe that the stated reason for adding the test case was that the condition of a string that was both "shouting" and "questioning" was previously ambiguous.

Ah yes, this is correct.

> One can add memoization to almost every value generated in every problem

If that's idiomatic in crystal, then that _should_ be preferred; if it makes sense. 

>  [...] you say, the compiler may do that automatically

Yes, in that case it would probably not be idiomatic to do, as it would most likely be idiomatic to rely on the compiler to do that. e.g. the JIT compiler in Ruby does _not_ do this.

>  Making this explicit simply adds "noise" to the solutions. 

I actually made a mistake to my example solution; it doesn't have duplicated output.

tenebrousedge: It would certainly not be idiomatic to use memoization for every value, but it would be equally impossible to say that memoization is not idiomatic. Crystal does have some neat options for that, but I'm not entirely sure they would work here; I'll look into it. The revised code you have shared tests `questioning?` twice. Is it possible to write this without duplicating either a conditional test, or a response?

SleeplessByte: It will only ever hit one questioning? Code path so it will only test it once!

tenebrousedge: Ah, you're right. I'll add your version in any case. The `OR` solution seems interesting. I don't suppose you would have an example to hand?

SleeplessByte: Not right now, but if I run into it, I'll forward it 🚀

PR #784, opened 2019-02-12
Ruby/Acronym: Experiment with mentor snippets

F3PiX:
I added some 'get you started'-snippets, that work for the vast majority of the Acronyms.

PR #975, opened 2019-04-28
add initial mentor notes for Crystal/Pangram

tenebrousedge:

### DIFF
--- a/tracks/crystal/exercises/pangram/mentoring.md
+++ b/tracks/crystal/exercises/pangram/mentoring.md
@@ -1,0 +1,56 @@
+### Algorithm
+
+There are lots of ways to solve this one. Methods that operate on the whole collection are preferred to iterative methods.
+
+```crystal
+module Pangram
+  ALPHABET = ('a'..'z').to_a
+  def self.pangram?(phrase)
+    (ALPHABET - phrase.downcase.chars).empty?
+  end
+end
+```
+```crystal
+module Pangram
+  ALPHABET = ('a'..'z').to_a
+  def self.pangram?(phrase)
+    ALPHABET & phrase.downcase.chars == ALPHABET
+  end
+end
+```
+```crystal
+module Pangram
+  ALPHABET = Set.new('a'..'z')
+  def self.pangram?(phrase)
+    ALPHABET.subset? Set.new(phrase.downcase.chars)
+  end
+end
+```
+Iterative:
+```crystal
+class Pangram
+  ALPHABET = ('a'..'z')
+
+  getter chars
+
+  def self.pangram?(phrase)
+    self.new(phrase).valid?
+  end
+
+  def initialize(phrase)
+    @chars = phrase.downcase.chars
+  end
+
+  def valid?
+    ALPHABET.all?(&->chars.includes?(Char))
+  end
+end
+```
+
+### Structure
+
+The exercise does not require the instantiation of any objects, so this should be considered optional. However, this problem does divide itself into two operations, one of which is transforming the input into a usable form. There is a symmetry between that idea and the idea that initializers prepare objects for use. On the other hand, it's difficult to see other purposes for this instantiated object, and it does make the code longer. When no clear arguments present themselves in either direction, the matter may be left to the judgement of the mentor and the student, but a discussion of the issue is probably worthwhile.
+
+### Style
+
+This problem requires some representation of the alphabet, either as a `Set`, `Range`, or `Array`. This should be stored as a constant. Iterative solutions probably lend themselves to the use of `&->`, as shown above.


tracks/crystal/exercises/pangram/mentoring.md
```diff
@@ -0,0 +1,56 @@
+### Algorithm
```
SleeplessByte: Please use the following as per the [README.md](https://github.com/exercism/website-copy/blob/master/README.md) of this repository:

```suggestion
### Reasonable solutions
```

tracks/crystal/exercises/pangram/mentoring.md
```diff
@@ -0,0 +1,56 @@
...
+
+There are lots of ways to solve this one. Methods that operate on the whole collection are preferred to iterative methods.
+
+```crystal
```
SleeplessByte: It would be helpful to add a line above or below each solution so mentors can _skim_ for the correct match. You can use heading level 4, or regular text.

tracks/crystal/exercises/pangram/mentoring.md
```diff
@@ -0,0 +1,56 @@
...
+    @chars = phrase.downcase.chars
+  end
+
+  def valid?
+    ALPHABET.all?(&->chars.includes?(Char))
+  end
+end
+```
+
+### Structure
```
SleeplessByte: Please use the following as per the [README.md](https://github.com/exercism/website-copy/blob/master/README.md) of this repository:
```suggestion
### Reasonable solutions
```

```md
### Common suggestions
```

In this case that might not be optimal for what you wrote; you can add the paragraph to the `Reasonable solutions` section, either at the top or bottom.

tracks/crystal/exercises/pangram/mentoring.md
```diff
@@ -0,0 +1,56 @@
...
+
+The exercise does not require the instantiation of any objects, so this should be considered optional. However, this problem does divide itself into two operations, one of which is transforming the input into a usable form. There is a symmetry between that idea and the idea that initializers prepare objects for use. On the other hand, it's difficult to see other purposes for this instantiated object, and it does make the code longer. When no clear arguments present themselves in either direction, the matter may be left to the judgement of the mentor and the student, but a discussion of the issue is probably worthwhile.
+
+### Style
```
SleeplessByte: Same as above.

The final missing heading is `### Talking points`. Although not required, you can add questions to challenge more advance learners with.

tenebrousedge:
Is there a previous discussion of the mentor notes structure that would give me a better understanding of why those headings are preferred? The README notes that it was copied from another repository, but I'm not sure which one that would be.

SleeplessByte:
> Is there a previous discussion of the mentor notes structure that would give me a better understanding of why those headings are preferred? The README notes that it was copied from another repository, but I'm not sure which one that would be.

@F3PiX Were there discussions on GitHub about these headings? (cc @iHiD) I don't think these were "copied" from another repo; the readme only gives a Ruby example which might have introduced them.

For me it's just the consistency across tracks. Every track's mentor notes are using these.

rpottsoh:
@SleeplessByte  are you perhaps thinking of #784?

F3PiX:
> @F3PiX Were there discussions on GitHub about these headings?

There was an initial format. In Ruby I started an experiment with a different format, that has been improved upon and other tracks like JS :-)))) are introducing different formats as well.  So, we're diverging. I would strive for consistency _within_ a track more than _between_ tracks.

SleeplessByte:
@exercism/crystal Any preference?

If we're changing this, I also suggest we change the wording [in the README.md](https://github.com/exercism/website-copy/blob/master/README.md).

mhelmetag:
I'm not sure I really understand what this is... Where does this show up on the site? That said, if it helps make something more clear, I'm all for it!

tenebrousedge:
@mhelmetag There's a link displayed near the top of the page when you're reviewing a problem, that will show you the mentor notes associated with that problem. Most tracks have notes for at least the core problems. We are working on introducing some of these to the Crystal track; I have submitted a PR for each of the core exercises. The way that I have structured these notes differs from the structure used until now by other tracks, which we are told is acceptable, but since we want to use a consistent format going forward, comments are being solicited as to whether the format used is acceptable, or whether some other structure would be preferable.

Personally, I think that it makes sense to discuss code in terms of three components, namely, [1] the algorithm being used: the sequence of operations, without consideration of structure; [2] how those operations may be arranged into objects and methods; and [3] finer points of style, which might increase clarity but not affect correctness: essentially those things which would be fixed by Crystal's formatting tool. These categories can usually be clearly distinguished.

The common structure on the other hand uses the headings of "Reasonable Solutions", "Common Suggestions", and "Talking Points". These to me are much more subjective categories. I am not sure what a "reasonable" solution might be, particularly as Exercism does not give any bounds on inputs. Some solutions might be reasonable for low `n`, but one can choose pathological inputs for nearly any algorithm. "Common Suggestions" is less subjective, but requires experience on the part of the note-writer to know what might be commonly suggested. The category of "Talking Points" is personally mystifying to me: I don't feel that I have the necessary insight into the intentions of the track maintainers to suggest what the primary purpose of a given problem might be, and I tend to think that the scope of what may be learned from a problem is pretty broad. There are people who will turn in hundreds of lines of highly abstract code for the "Raindrops" problem, which is not something that could reasonably be encouraged in a "talking point", but would still be worth talking about.

I'm not averse to rewriting the existing pull requests into whatever format the community prefers, but if we are going with "Reasonable Solutions" and "Talking Points", then I would humbly request assistance with determining what those might be for each problem.

SleeplessByte:
> These categories can usually be clearly distinguished.

This is _not_ the case in (most of) the other tracks, but if that's the case for crystal, I do opt for having that structure. Note that my suggestion was never to _remove_ the structure, but to rename the heading and add the others.

>  I am not sure what a "reasonable" solution might be, particularly as Exercism does not give any bounds on inputs. Some solutions might be reasonable for low n, but one can choose pathological inputs for nearly any algorithm. 

Actually, no, this is not as subjective as you describe. In lieu with `project-track-anatomy` which will come to the Crystal track, each exercise will specifically try to teach one or several subjects. The input bound is both the given README.md + given test suite. If the README.md says a solution should be of a certain big O or less, then that's the input bound. If it's not, it should be considered in the story:

e.g. `resistor-color-duo` will only ever hit 2 colors, and the list of colors will never grow, as per the story. This means that _over optimised_ solutions using a hash structure are completely unnecessary. That doesn't mean they're wrong, but it means that an O(n) lookup using linear search is completely fine. 

e.g. `grains` currently uses big integer (> 2^64) in some tracks. In these tracks that is made clear through the tests ánd the readme. The solution should be able to handle that as per the exercise.

> ... but requires experience on the part of the note-writer to know what might be commonly suggested

Absolutely correct. It's common that first iteration of notes doesn't _have_ this heading at all, but is added after a certain mentor or maintainer has mentored a sufficiently large set of exercises.

> I don't feel that I have the necessary insight into the intentions of the track maintainers to suggest what the primary purpose of a given problem might be, and I tend to think that the scope of what may be learned from a problem is pretty broad.

I've heard this before and I think you're right 😞. We're trying to solve this going forward, and it's tangent on the `project-track-anatomy`. Currently there might actually _not_ be a single subject to teach for these exercises but that **will** change in the future. Nonetheless, talking points generally hold those things such as mentioning alternative ways, performance, codepoints/graphmemes, and so forth.

---------

After reading your last comment, I understand better _why_ you want the heading you have. I don't think it's correct to **not** give a set of reasonable solutions or at least describe what is approvable and what is not, **but** I also see why you want the headings as you chose them. 

As a first iteration, we might want to keep what you currently have -- but I would suggest to:
- either be clear about what is approvable and what's not
- add the three headings from the main readme as you go, where applicable. 

However, as an example:

> The exercise does not require the instantiation of any objects, so this should be considered optional. However, this problem does divide itself into two operations, one of which is transforming the input into a usable form. There is a symmetry between that idea and the idea that initializers prepare objects for use. On the other hand, it's difficult to see other purposes for this instantiated object, and it does make the code longer. When no clear arguments present themselves in either direction, the matter may be left to the judgement of the mentor and the student, but a discussion of the issue is probably worthwhile.

The first phrase: is it really optional? If so, then it's fine as is. If one is more idiomatic than the other, it should probably not be optional (either never instantiate if you don't need to, or always do). There are many tracks that also leave this optional, so that's fine.

> the matter may be left to the judgement of the mentor and the student, but a discussion of the issue is probably worthwhile

Basically this entire paragraph would therefore be a talking point :)

> This problem requires some representation of the alphabet, either as a `Set`, `Range`, or `Array`. This should be stored as a constant. Iterative solutions probably lend themselves to the use of `&->`, as shown above.

Here are two common suggestions:
- If the alphabet is not represented as a constant, suggest they use a `Set`, `Range`, or `Array` and store it as a constant
- If the student has an iterative solution, make sure they know about `&->`, as shown above.

iHiD:
Hello. Sorry I didn't respond to this earlier.  I'd love to get it merged!

The [README](https://github.com/exercism/website-copy#mentor-notes) defines how these headings should be used:

> - Reasonable solutions: 1+ solutions that solve this problem well.
> - Common suggestions: Good suggestions specific to this exercise. Good lessons that emerge from it.
> Talking points: Questions to challenge more advance learners with.


The aim with mentoring is to teach a student the idiomatic way to solve a problem in a language - it's not to teach specific algorithms or specific ways of approaching the general problem. "Reasonable Solutions" are solutions that display a language's idioms and can therefore be approved without needing to ask the student for a second iteration. Where a solution is not reasonable, I would expect to see ideas in the "Common suggestions" section that helps guide mentors as to what to say. 

For example, the [Ruby notes for this exercise](https://github.com/exercism/website-copy/blob/master/tracks/ruby/exercises/pangram/mentoring.md#reasonable-solutions) show to "Reasonable solutions" - on that is maybe more "advanced" than the other, and uses a different approach/algorithm but both are idiomatic so both would be approvable.

With regards to your notes, I think that just changing the top heading to "Reasonable Solutions" and then putting the remaining points under "Common suggestions" works absolutely fine.

As a final note, the main reasons for insisting on consistency are the ability for mentors to easily mentor other languages with a common schema of notes, and also the future ability for us to do "clever" things in the UI based on the fact we have common headings amongst these files.

@tenebrousedge Would you like me to make the changes for you, or do you want to add another commit to do it?

tenebrousedge:
I decline to change the section headers.

iHiD:
Fair enough. Closing in favour of https://github.com/exercism/website-copy/pull/1193

PR #977, opened 2019-04-28
add initial mentor notes for Crystal/Hamming

tenebrousedge:


tracks/crystal/exercises/hamming/mentoring.md
```diff
@@ -0,0 +1,57 @@
...
+    end
+  end
+end
+```
+```crystal
+module Hamming
+  def self.compute(strand, other)
+    raise ArgumentError.new unless strand.size == other.size
+    strand.each_char.with_index do |base, index|
+      base != other[index]
```
SleeplessByte: Does this work? It doesn't use `count`

PR #978, opened 2019-04-28
add initial mentor notes for Crystal/Leap

tenebrousedge:

### DIFF
--- a/tracks/crystal/exercises/leap/mentoring.md
+++ b/tracks/crystal/exercises/leap/mentoring.md
@@ -5,7 +5,7 @@ There are technically two ways to do this, but only one is useful.
 ```crystal
 module Year
   def self.leap?(year : Int32) : Bool
-    year.divisible_by?(4) && !year.divisible_by?(100) || year.divisible_by?(400)
+    year.divisible_by?(4) && (!year.divisible_by?(100) || year.divisible_by?(400))
   end
 end
 ```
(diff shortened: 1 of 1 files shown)


tracks/crystal/exercises/leap/mentoring.md
```diff
@@ -0,0 +1,27 @@
+### Algorithm
+
+There are technically two ways to do this, but only one is useful.
+
+```crystal
+module Year
+  def self.leap?(year : Int32) : Bool
+    year.divisible_by?(4) && !year.divisible_by?(100) || year.divisible_by?(400)
```
SleeplessByte: Depending on the order of operations, you might be missing parentheses around the test for 100 and 400.

tenebrousedge: Parentheses are not necessary for the correct order of operations.

SleeplessByte: Does that mean that if a year is NOT divisible by 4 (1023), it will NOT be checked against 400? As in:

- `false && !year.divisible_by?(100) || year.divisible_by?(400)`: shortcuts to false instantly
- and NOT: `false || year.divisible_by?(400)` -> `false || false`

tenebrousedge: Crystal does use short-circuit evaluation, yes.
```crystal
1023.divisible_by?(4) && true
#=> false
1024.divisible_by?(4) && false || "a"
#=> "a"
```

tenebrousedge: True enough. Adding that.

PR #981, opened 2019-04-28
Rust: add mentoring notes for bob.

TheDarkula:


NobbZ:
Matching is not specific to rust. It's very common in many languages, especially functional ones.

Also I'm not sure if disabling a default lint if the default linter should be the goal of mentoring.

In general I'm in favour of the lint and wouldn't use matching for booleans but a traditional if.

For this even as a "guard":

```
if trimmed.is_empty { return FINE; }

// do the other stuff
```

This will reduce necessary indentation.

tracks/rust/exercises/bob/mentoring.md
```diff
@@ -0,0 +1,140 @@
...
+A reasonable solution should:
+
+- Use constants
+- Use variables
+- Use `match` statements
+- Check if the input string is empty and return the "Sure" constant
+before doing any additional processing
+- Use simple methods from the standard library
+- Not use the regex crate
+- Not use separate functions
```
KillTheMule: I disagree here. Modularization is somewhat style-dependent, but there's not real disadvantage to putting things into functions, generally speaking. I'd really not like to push that onto people who're new to the language and try to organize things in a way that fits them, if it does not come with a definite disadvantage. Of course, people should not make useless separations that only add lines like `fn is_empty(s: &str) { s.is_empty() }`, but that's much less general than what you've written here.

tracks/rust/exercises/bob/mentoring.md
```diff
@@ -0,0 +1,140 @@
...
+const WHATEVER: &str = "Whatever.";
+
+#[allow(clippy::match_bool)]
+pub fn reply(message: &str) -> &str {
+    let trimmed: &str = message.trim();
+
+    match trimmed.is_empty() {
+        true => FINE,
+        false => {
+            let is_question: bool = trimmed.ends_with("?");
```
KillTheMule: This should maybe noted. People are often using `chars().last()`, which works as well of course, but `ends_with` is much clearer.

tracks/rust/exercises/bob/mentoring.md
```diff
@@ -0,0 +1,140 @@
...
+
+            let contains_alphabetic_characters: bool = trimmed.chars().any(|character| character.is_alphabetic());
+            let is_uppercase: bool = trimmed == trimmed.to_uppercase();
```
KillTheMule: This is pretty inefficient, since it's allocating a new `String` needlessly. I suggest making a `chars()` loop here to set both `contains_alphabetic_characters` and `is_uppercase` in one go.

tracks/rust/exercises/bob/mentoring.md
```diff
@@ -0,0 +1,140 @@
...
+```
+Well done you for seeing that we need to check if the string is empty with `== ""`. There's actually a method, `.is_empty()`, that we can use instead, which is much cleaner.
+```
+
+If they don't use `match`:
+```
+Check out using `match` instead of `if`. For some things, like destructuring, it's required. It's an idiom specific to Rust, so I always default to it unless the logic is comparing many different things and a match in that case makes it messy.
+```
+
+If they use an `if` on their `.is_empty()` check:
```
KillTheMule: I disagree quite a bit here. Doing an early return, and not having the rest in an arm is a win for readability here imho (subjective of course), and lessens the dreaded rightwards drift that sometimes plagues rust code. I'd say both variants are totally fine and you should remove this suggestion.

tracks/rust/exercises/bob/mentoring.md
```diff
@@ -0,0 +1,140 @@
...
+See about doing a  `match`, instead of `if`, on `your_trimmed_variable.is_empty()`. And, if it's `true`, instantly output our "Fine. Be that way!" constant. Then, everything else will be in the `false` arm.
+Like this:
+    ```rust
+    match your_trimmed_variable.is_empty() {
+        true => FINE,
+        false => {
+    ```
+Be aware that if you use Clippy, it will tell you that it is improper, it's purely a stylistic choice :)
+We can tell Clippy to allow it with `#[allow(clippy::match_bool)]`.
+```
```
KillTheMule: This line needs to be removed, doesn't it?

tracks/rust/exercises/bob/mentoring.md
```diff
@@ -0,0 +1,140 @@
...
+If they use regex:
+```
+This is good! Great idea using the regex crate! Let's look at how to do this with just the standard library :)
+```
+Then note the following:
+```
+See about finding a way to check if it's alphabetic without using regex.
+```
+
+If they use separate functions:
```
KillTheMule: As noted above, I'd generally allow this, very much depending on the function and context, of course.

tracks/rust/exercises/bob/mentoring.md
```diff
@@ -0,0 +1,140 @@
...
+To quote from [Rust By Example](https://doc.rust-lang.org/rust-by-example/scope/lifetime/static_lifetime.html): "When `static_string` goes out of scope, the reference can no longer be used, but the data remains in the binary."
+
+What I think the compiler is doing is recreating the pointer to the static data if we use a plain `let` line, whereas with `const`, the data and the pointer are always kept alive.
+
+Also, the use of `const` allows us to have it globally scopable.
+```
+
+If they use `_` instead of `false` in their match statement:
+```
+Just a stylistic note, the `_` in the true/false `match` can simply be `false` because there is no other option :)
```
KillTheMule: Imho that's not only stylistic. It suggests to the reader that there's something else to catch while there isn't, making it less readable. In other circumstances, it can even be a source of bugs, so the use of `_` should be limited to the cases where it's needed. I'd strengthen this.

KillTheMule:
@TheDarkula I've added some inline notes, I hope I don't come off as unfriendy or something, just trying to help make this good notes :)

> In general I'm in favour of the lint and wouldn't use matching for booleans but a traditional if.

For single booleans, I agree. If there are more than one, like in @TheDarkula's solution, it's pretty elegant to match imho, and it helps you in that you can't forget one possible combination (starting at 3 bools, I'd say that's definite possiblity).

tracks/rust/exercises/bob/mentoring.md
SleeplessByte: The issue is the nested code block inside the comment.

@TheDarkula  you can use 4 space identation for line 82-92 (remove the backticks) and then the backticked code block will remain.

NobbZ: Personally I prefer to use longer fences when nesting codeblocks instead of increasing indentation:

`````md
````md
This has another codeblock:

```rs
pub mod foo;
```
````
`````

Which will render like this:

````md
This has another codeblock:

```rs
pub mod foo;
```
````

Comment (SleeplessByte) [reply tracks/rust/exercises/bob/mentoring.md]: Ah yes @NobbZ . I like that!

File: tracks/rust/exercises/bob/mentoring.md (hunk 1 above)
Comment (TheDarkula) [reply tracks/rust/exercises/bob/mentoring.md]: This is a generalisation because, for the most part, when they do create separate functions, they are overcomplicating things. This guidance is so mentors can help direct them to a more concise and idiomatic way of thinking :)

Here's just one example: `fn is_question(message: &str) -> bool {
    message.ends_with("?")
}`

This can simply be `let is_question: bool = trimmed_message.ends_with("?");`

And, yes, I have seen `.is_empty()` in a separate function. Most, if not all, of the separate functions are "useless separations" :)

File: tracks/rust/exercises/bob/mentoring.md (hunk 2 above)
Comment (TheDarkula) [reply tracks/rust/exercises/bob/mentoring.md]: Thanks! We actually have it in the "Common Suggestions" section below, but I've clarified it a bit more :)

File: tracks/rust/exercises/bob/mentoring.md (hunk 3 above)
Comment (TheDarkula) [reply tracks/rust/exercises/bob/mentoring.md]: If you mean, `let is_yelled: bool = trimmed.chars().any(|character| character.is_alphabetic()) && trimmed == trimmed.to_uppercase();`, I prefer setting the separate variables to help explain what each piece is actually doing. I find it more readable. Now, I don't necessarily tell the students that they have to do that. If not, please elaborate more :)

File: tracks/rust/exercises/bob/mentoring.md (hunk 4 above)
Comment (TheDarkula) [reply tracks/rust/exercises/bob/mentoring.md]: I use `match` as much as possible as I find it more readable. So, I suggest it to my students and if they prefer the `if` statement, that's fine, but most of them didn't think about it as an option because of the pressures of taking on a new language and having clippy tell them that it is "wrong", and are grateful for the suggestion :)

Quoting from [clippy](https://rust-lang.github.io/rust-clippy/master/index.html#match_bool), "It makes the code less readable.". Being that it is purely a stylistic choice, this gives the student the power to make their own decisions about how they feel their code looks best.

File: tracks/rust/exercises/bob/mentoring.md (hunk 5 above)
Comment (TheDarkula) [reply tracks/rust/exercises/bob/mentoring.md]: Good catch, thanks so much! Got it changed here and for lines 116-121 :)

File: tracks/rust/exercises/bob/mentoring.md (hunk 6 above)
Comment (TheDarkula) [reply tracks/rust/exercises/bob/mentoring.md]: Addressed above :)

File: tracks/rust/exercises/bob/mentoring.md (hunk 7 above)
Comment (TheDarkula) [reply tracks/rust/exercises/bob/mentoring.md]: Great, thanks! I've added more to the suggestion :)

**Comment by TheDarkula** [comment]
@SleeplessByte

File: tracks/rust/exercises/bob/mentoring.md (hunk 1 above)
Comment (KillTheMule) [reply tracks/rust/exercises/bob/mentoring.md]: > This is a generalisation because, for the most part, when they do create separate functions, they are overcomplicating things.

I suggest avoiding generalisations in mentor notes. Some mentors are going to take it at face value and force students to do this, if you phrase it this way.

> And, yes, I have seen `.is_empty()` in a separate function. Most, if not all, of the separate functions are "useless separations" :)

That's not true. A separation like my example is useless because it doesn't actually modularize something, save lines or adds clarity for code. Look at the clock exercise, where people sometimes factor out the `modulo` function, which is only one line. This would not be "useless", although it adds more lines.

File: tracks/rust/exercises/bob/mentoring.md (hunk 3 above)
Comment (KillTheMule) [reply tracks/rust/exercises/bob/mentoring.md]: Yeah, being inefficient surely isn't a problem in exercises, but otoh doing it this way discards a bit of what a "low level" language can do, compared to a "higher level" language.

Again, if you write it this way in the mentoring notes, I'm afraid of mentors assuming this is the way it _has_ to be done. Maybe add a second, more efficient example solution, and add an inline comment referring to it?

File: tracks/rust/exercises/bob/mentoring.md (hunk 1 above)
Comment (SleeplessByte) [reply tracks/rust/exercises/bob/mentoring.md]: @TheDarkula My 2 cents here:

"Not use separate functions" is too strong of a wording but I understand what you're trying to accomplish. I would suggest you remove it here, but we can make subsequent edits to this note adding _examples_ why the separated out function is unnecessary. 

```suggestion
```

tracks/rust/exercises/bob/mentoring.md
SleeplessByte: The best way to go about this is to, below this example block, add a section with similar wording to:

```markdown
For `is_yelled` , variations include using a `chars()` loop to ensure no extra `String` is allocated.
```

I suggest **not** adding inline comments, as some mentors see that as a mandatory comment in the solution and it's much clearer to keep the same structure everywhere (having variations below the solution and not inline, or both).

tracks/rust/exercises/bob/mentoring.md
SleeplessByte: It should therefore be a talking point (tips, hints, brain farts) and not a suggestion (usually things that make a solution disprovable).

TheDarkula:
@SleeplessByte All done :)

SleeplessByte:
🚀

PR #983, opened 2019-05-01
Add me as a mentor for the rust track

KillTheMule:


mentors/rust/k.json
```diff
@@ -6,5 +6,13 @@
     "link_url": null,
     "avatar_url": null,
     "bio": "I love Rust and C and use them at my day job in automotive. I work with Embedded software but I'm also reasonably familiar with systems programming as well"
+  },
+    {
+    "github_username": "KillTheMule",
+    "name": "null",
```
rpottsoh: I think it would look better on the site if your name didn't literally appear as `null`.  I think if you remove the `"` marks the site will probably use `github_username` as your name.
```suggestion
    "name": null,
```

rpottsoh: If you don't want to use your GitHub username then make up a name for `"name"`.

PR #991, opened 2019-05-02
Add mentoring notes for java/reverse-string

michael-berger-FR:


tracks/java/exercises/reverse-string/mentoring.md
```diff
@@ -0,0 +1,41 @@
+### Reasonable solutions
+
+Loop-based solution:
+```java
+class ReverseString {
+    String reverse(String inputString) {
+        String output = "";
+        for (int i = inputString.length() - 1; i >= 0; i--) {
+          output += inputString.charAt(i);
```
sshine: Unless the Java compiler can somehow convert this to `StringBuffer` code underneath, shouldn't this example have a comment under it that says something like the following?

> Students may submit a solution that builds new strings. Since this has a memory overhead of copying the entire output string once for each character in it, a solution based on `StringBuilder` or `StringBuffer` is preferred.

SleeplessByte: The Java compiler does make optimisations here.

When you have less than ~1000 concatenations, the performance of `StringBuilder.append` is similar to string concatenation.

`StringBuffer` is worse from the get-go.

From the spec:
> To increase the performance of repeated string concatenation, a Java compiler _may_ use the StringBuffer class or a similar technique to reduce the number of intermediate String objects that are created by evaluation of an expression.

michael-berger-FR: Hello @sshine and @SleeplessByte ,
I could not agree more. I replaced the solution with StringBuffer by StringBuilder and added your two points in "common suggestions".
FYI, I have an issue under GitHub because of a deleted fork. That's why I'll close this PR and open a new one with the discussed modifications.
Thanks to both of you!

PR #968, opened 2019-04-28
Updated  swift / exercises / leap notes

alexito4:
Swift 5 includes `Int.isMultiple(of:)`

PR #992, opened 2019-05-03
Update to swift 5 isMultiple(of:)

asa-z:
The solutions should use the swift 5 isMultiple(of:) over a user-provided extension to Int.


asa-z:
This is duplicate of #968

PR #1003, opened 2019-05-09
[Java/Hamming] Update mentor notes to reflect the addition of new tests

michael-berger-FR:
New tests have been added to the test suite (exceptions thrown, "right strand must not be empty" and "left strand must not be empty"). This commit update the reasonnable solutions to take these new tests into account.

### DIFF
--- a/tracks/java/exercises/hamming/mentoring.md
+++ b/tracks/java/exercises/hamming/mentoring.md
@@ -6,8 +6,9 @@ class Hamming {
 
     Hamming(String leftStrand, String rightStrand) {
         if (leftStrand.length() != rightStrand.length())
-            throw new IllegalArgumentException("leftStrand and rightStrand must be of equal length.");
-        this.leftStrand = leftStrand;
+            throw new IllegalArgumentException(leftStrand.length() == 0 ? "left strand must not be empty."
+                    : (rightStrand.length() == 0 ? "right strand must not be empty."
+                    : "leftStrand and rightStrand must be of equal length."));        this.leftStrand = leftStrand;
         this.rightStrand = rightStrand;
     }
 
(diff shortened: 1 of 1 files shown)


tracks/java/exercises/hamming/mentoring.md
```diff
@@ -6,8 +6,9 @@ class Hamming {
 
     Hamming(String leftStrand, String rightStrand) {
         if (leftStrand.length() != rightStrand.length())
-            throw new IllegalArgumentException("leftStrand and rightStrand must be of equal length.");
-        this.leftStrand = leftStrand;
+            throw new IllegalArgumentException(leftStrand.length() == 0 ? "left strand must not be empty."
```
jmrunkle: Use isEmpty to check for empty strings

tracks/java/exercises/hamming/mentoring.md
```diff
@@ -6,8 +6,9 @@ class Hamming {
...
+                    : (rightStrand.length() == 0 ? "right strand must not be empty."
```
jmrunkle: Using a ternary operator inside of a ternary operator is generally considered bad style / practice in Java.

Since the validation is no longer so trivial, it may be best to have our mentoring notes suggest delegating to another method rather than trying to do everything straight in the constructor. For example, extracting a `validateStrands` method.

```java
private static void validateStrands(String leftStrand, String rightStrand) {
  if (leftStrand.length() == rightStrand.length()) {
    return;
  }
  if (leftStrand.isEmpty()) {
    throw new IllegalArgumentException("left ...");
 }
  if (rightStrand.isEmpty()) {
    throw new IllegalArgumentException("right ...");
 }
  throw new IllegalArgumentException("...");
}
```

sshine:
@michael-berger-FR: Would you like to address @jmrunkle's feedback so that your changes can be merged?

SleeplessByte:
@jmrunkle  if you have time, can you change your comments to suggestions, then add all the suggestions to a batch, apply. We can then merge this.

SleeplessByte:
@sanderploegsma can you have a look here? If you add suggestions we can batch apply them and merge this.

sanderploegsma:
Actually these test cases have been removed again recently, because of an update in the problem specifications. The [current version](https://github.com/exercism/java/blob/main/exercises/practice/hamming/.meta/src/reference/java/Hamming.java) looks pretty similar to the existing mentoring notes, the only difference is the message used in the exception.

So technically I think this PR is no longer applicable.

SleeplessByte:
Thank you for checking.

PR #1008, opened 2019-05-11
Python/Isogram: Add example of solution with handcrafted sets

jeffdparker:
def is_isogram(string):
    arr = [0] * 256

    for l in string.lower():
        if arr[ord(l)] == 0:
            arr[ord(l)] += 1
        elif (l != ' ') and (l != '-'):
            return False

    return True

### DIFF
--- a/tracks/python/exercises/isogram/mentoring.md
+++ b/tracks/python/exercises/isogram/mentoring.md
@@ -101,6 +101,30 @@ def is_isogram(string):
     return len(set(word)) == len(word)
 ```
 
+The solution might define their own sets:
+
+```python
+def is_isogram(string):
+    arr = [0] * 256
+
+    for l in string.lower():
+        if arr[ord(l)] == 0:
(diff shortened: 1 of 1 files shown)


NobbZ:
Why create an array of 256 elements if you only consider 228 possible
elements of it and should only consider 26?

Jeff Parker <notifications@github.com> schrieb am Sa., 11. Mai 2019, 15:00:

jeffdparker:
I see that I should have made it clearer that this wasn't a recommendation.  

As to your point that you only need 26 slots, I think we have blown past that limit in Python 3.

```python
    s = '\U00000394'
    print(s.isalpha())

    True
```

NobbZ:
> As to your point that you only need 26 slots, I think we have blown past that limit in Python 3, as pointed out in the commentary about the proposed solution.

This exercise is explicit about only considering the english alphabet of `a` to `z` and their uppercase variants.

Issue #948, opened 2019-04-21
Go-Analyzer comments


tehsphinx:
I added the comments to the `go-analyzer` repo in the `comments` branch and `/comments` folder. Feel free to move them to this repository whenever you are ready.

https://github.com/exercism/go-analyzer/tree/comments/comments

I used a directory structure that can be directly derived from the comment handle itself:
- `go.general.strings_trim_space_used` => `go/general/strings_trim_space_used.md`
- `go.two_fer.missing_share_with_function` => `go/two_fer/missing_share_with_function.md`
- `go.hamming.mixture_of_runes_and_bytes` => `go/hamming/mixture_of_runes_and_bytes.md`

iHiD:
@tehsphinx Thanks!

- Could you remove the single bullet point syntax pls (e.g. [here](https://github.com/exercism/go-analyzer/blob/comments/comments/go/two_fer/extra_name_variable_created.md))
- Would you mind putting a WIP PR up (Just prefix the title with `[WIP]`) and we'll get some discussion going on that :) 

No rush on this.

Thanks!

tehsphinx:
Closing in favour of the PR #1009

PR #1009, opened 2019-05-11
go-analyzer comments for twofer and hamming (approve_with_comment)

tehsphinx:
As suggested by @iHiD in #948 here the PR.


SleeplessByte:
@iHiD will this run through Ruby's formatting? In that case probably should be `%<name>s` instead of `%{name}`. We should probably document somewhere what the expected format is 🗡

tehsphinx:
- renamed folder to `automated-comments` as requested by @iHiD.
- moved comments for `disapprove_with_comment` into a separate PR: #1096

iHiD:
Just as a general FYI for analyzer maintainers, the plan for this PR is for @kytrinyx to review it in the next few days and give guidance over tone and language for the comments here. In parallel, I'll finalise any formatting things I care about (e.g. how to interpolate) and @tehsphinx will keep things updated as we go. Then once all that's agreed we can merge it and write up the guidelines for other tracks.

If other maintainers would like top open WIP PRs for this, so that we can consider different problems in different languages, that would be welcome too, but please bare in mind that our first-pass will be here, and you might then need to re-write your comments in line with mine and Katrina's recommendations.

We're aiming to launch approve_with_comment before I go on holiday in the second week of July, but that's a target, not a deadline :)

Any general questions not specifically relevant to this PR - please ping me in Slack :)

automated-comments/go/hamming/invert_if_for_happy_path_on_left.md
```diff
@@ -0,0 +1,4 @@
+You could make the function more readable by inverting the condition of the `if` statement handling the error case. 
+The rest of the function can then be outdented because an `else` branch is not needed. A Go proverb 
+says "The happy path is left-aligned", meaning the normal flow of control follows the left-hand margin, and only 
+exceptional or error cases are indented.
```
kytrinyx: I really like the Go proverb part in the second sentence. I think we can tighten up the first with something like:

> To simplify, invert the `if` condition handling the error case, then get rid of the `else` branch by returning within the `if`.

automated-comments/go/hamming/lossy_rune_to_byte_conversion.md
```diff
@@ -0,0 +1,5 @@
+You are converting a `rune` to a `byte`. Go `runes` can consist of multiple `bytes`, so this conversion can cause you problems if you have to deal with non-ASCII characters. However, in this case it's OK, because the input string contains only ASCII (single-byte) characters.
+
+To get a deeper understanding of these issues you might like to read [Strings, bytes, runes and characters in Go](https://blog.golang.org/strings) on the Go blog.
+    
+You can also add more tests to  this exercise containing non-ASCII characters and see what happens.
```
kytrinyx: Maybe:

> The `rune` to `byte` conversion is safe here, because the input string contains only ASCII characters. Go `runes` can consist of multiple `bytes`, though, which can cause trouble when dealing with non-ASCII characters.
> 
> For more on this, read [Strings, bytes, runes and characters in Go](https://blog.golang.org/strings) on the Go blog.
>
> To see this in practice, add a test containing non-ASCII characters.

automated-comments/go/hamming/minimize_slice_of_rune_conversions.md
```diff
@@ -0,0 +1,3 @@
+You are converting the variable `%{name}` multiple times to a `[]rune`. 
+Since this is a comparatively expensive operation it would be better to do the conversion once, save the result
+to a variable and reuse that.
```
kytrinyx: Do we know that they've used the name `name`?

To tighten it up, perhaps:

> Prefer to convert to a `[]rune` only once, saving the result in a variable and reusing it, as this conversion is a comparatively expensive operation.

automated-comments/go/hamming/mixture_of_runes_and_bytes.md
```diff
@@ -0,0 +1,3 @@
+Beware that you are using a mixture of `runes` and `bytes` here. This is working fine as the exercise only
+requires you to handle `ascii` characters. You can go ahead and add a few more test cases containing non-ascii characters.
+See what happens!
```
kytrinyx: We should be consistent about how we refer to ASCII. We've used:

- ascii
- ASCII
- `ascii`

To tighten this up, maybe:

> Using a mixture of `runes` and `bytes` is safe here, because the input only uses ASCII characters. To see what happens when it's not safe, try adding a test that uses non-ASCII characters.

automated-comments/go/hamming/return_on_error_case.md
```diff
@@ -0,0 +1,8 @@
+Go is a programming language where returning as soon as possible and multiple times from a function or method 
+is practiced a lot. It keeps code readable and maintainable.
+
+Usually one would handle edge cases first and error cases as soon as they occur by returning. This practice is
+nicely explained in the blog post [Align the happy path to the left edge](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88) by Mat Ryer.
+You can also watch the talk he gave on the toppic.
+  
+Can you apply this to your solution?
```
kytrinyx: I'd like to not use questions from the auto-mentoring comments, since questions presume answers, and if the person comments with an answer the auto-mentoring will not have any way of responding again.

To tighten this up, maybe:

> In Go, it's conventional to use multiple returns, returning as soon as possible from a function. Typically edge cases and error cases are handled as early as possible, leaving the happy path aligned to the left edge. For a nice explanation of this practice, see [Align the happy path to the left edge](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88) by Mat Ryer, or watch the talk that he gave on the same topic.

automated-comments/go/hamming/rune_index_used_with_bytes.md
```diff
@@ -0,0 +1,12 @@
...
+Try adding the following test to see what happens if `non-ascii` characters get involved:
+
+```go
+{
+	s1:          "aüa",
+	s2:          "aÃa",
+	want:        1,
+	expectError: false,
+}
+```
```
kytrinyx: > Iterating over `runes` and using the index to get single `bytes` is safe in this case, because the input is all ASCII characters. To see what happens when non-ASCII characters are used, add this test case:

automated-comments/go/hamming/rune_or_byte_to_string_conversion.md
```diff
@@ -0,0 +1,3 @@
+As you know, using `range` over a `string` gives you a `rune`, but indexing into a `string` returns a `byte` value. The difference is described in more detail here: [Strings, bytes, runes and characters in Go](https://blog.golang.org/strings).
+
+There is no need to convert runes and bytes to string to compare them. You should decide if you want to compare `runes` or `bytes`.
```
kytrinyx: As a general rule, I would avoid saying "as you know" since we can't know what they know.

To tighten up:

> For consistency, comparing either `runes` or `bytes`, but not both. Ranging over a `string` gives `rune` values, whereas indexing into a `string` gives `byte` values. The difference is described in detail in [Strings, bytes, runes and characters in Go](https://blog.golang.org/strings).

automated-comments/go/hamming/use_increase_operator.md
```diff
@@ -0,0 +1 @@
+Go has a special `++` operator for incrementing a variable, like this: `count++`.
```
kytrinyx: Here the context and recommendation is implied. I think I'd lean towards being explicit:

> In most cases, Go programmers would use Go's `++` operator to increment. E.g. `count++`.

automated-comments/go/hamming/use_of_naked_returns.md
```diff
@@ -0,0 +1,4 @@
+`Naked returns` (or `bare returns`) are considered evil by most of the Go community. Simply because they make code harder to read.
+The official [Go code review guide](https://github.com/golang/go/wiki/CodeReviewComments#naked-returns) suggests to avoid them if the function has more than a few lines of code.
+There is also a [proposal](https://github.com/golang/go/issues/21291) to remove bare returns altogether in Go 2.
+My suggestin would be to avoid them.
```
kytrinyx: > are considered evil by most of the Go community

That's a very strong statement. How about:

> The Go community recommends avoiding _naked returns_ (or _bare returns_) under normal circumstances, as they typically make code harder to read. See the [official Go code review guide](https://github.com/golang/go/wiki/CodeReviewComments#naked-returns) for details.
> There is a [proposal](https://github.com/golang/go/issues/21291) to remove bare returns altogether in Go 2.

automated-comments/go/hamming/use_zero_values_on_error_return.md
```diff
@@ -0,0 +1,3 @@
+When your function returns an `error`, this tells the caller that the other return value shouldn't be used. 
+Since it doesn't actually matter what you return in that case, the convention is to return the [zero value](https://golang.org/ref/spec#The_zero_value).
+In languages that don't allow multiple return values from functions, you often need to use an invalid 'out-of-band' value to signal an error, such as `-1`. That's not necessary in Go.
```
kytrinyx: > By convention, when returning an `error` value, other values are not used, and it's considered idiomatic to return the [zero value](https://golang.org/ref/spec#The_zero_value) in this situation. The `error` value makes using an invalid "out of band" value to signal an error redundant in Go.

automated-comments/go/two_fer/extra_name_variable_created.md
```diff
@@ -0,0 +1,4 @@
+There is no need to create an extra variable to hold the name. Go does not pass variables by reference. 
+You will therefore not change the `name` variable beyond the scope of the `ShareWith` function.
+
+Go does not have default values. Overwriting the `name` with `you` is setting a default value for `name` if it is empty.
```
kytrinyx: Maybe:

> Creating an extra variable to hold the name is unnecessary, as Go does not pass variables by reference. Re-assigning to `name` will not change the value beyond the scope of the `ShareWith` function.
> Since Go does not provide a way of providing default values to function parameters, it is a common idiom in Go to re-assign the value to provide a default.

automated-comments/go/two_fer/missing_function_comment.md
```diff
@@ -0,0 +1 @@
+Community style says that a function in Go should have a comment. For more details, have a look at the comments section below.
```
kytrinyx: > It's idiomatic in Go to provide a comment for each exported function. For more details, see the comments section below.

Question: do we only post the comment section if we pass back a comment about comments?

automated-comments/go/two_fer/missing_package_comment.md
```diff
@@ -0,0 +1 @@
+Go's community style says that package `twofer` should have a comment. For more details, have a look at the comments section below.
```
kytrinyx: Rather than "Go's community style" I would prefer _It's idiomatic in Go_. I don't really know what a 'community style' is.

Rather than "have a look" I would prefer "see", as this is shorter.

automated-comments/go/two_fer/plus_used_for_concatenation.md
```diff
@@ -0,0 +1,6 @@
+I see you are using `+` to concatenate the string. That is perfectly legit! 
+If you didn't know about the `fmt` package, check it out. It can do lots of 
+things related to string formatting, including concatenating strings.
+
+Using `+` is faster in simple concatenations and probably the right choice here. 
+I just wanted you to know about `fmt`, because it's awesome, and you'll be using it a lot.
```
kytrinyx: Maybe:

> In this case using `+` is almost certainly the right choice, as it's shorter in this case, and often faster for simple concatenations. If it's unfamiliar, check out the powerful `fmt` package. It's what people typically use for more complicated string concatenation, and for other string formatting needs.

automated-comments/go/two_fer/strings_trimspace_used.md
```diff
@@ -0,0 +1,5 @@
+Don't worry about cleaning up input in exercises on exercism. The platform does not require that.
+
+Even if this were a real application all the functions within your applications should not clean up
+input data. If you have potentially bad data coming from a source you should clean it up in the code reading
+from that source. Then you can consider the data clean within the rest of your code.
```
kytrinyx: This is a duplicate of automated-comments/go/general/strings_trim_space_used.md

automated-comments/go/two_fer/use_fmt_placeholder_for_string.md
```diff
@@ -0,0 +1,5 @@
+Tip: any type of value in Go can be printed using the format character `%v`. However, there's also a different format for each type (for example `%s` for strings). The type specific formats also offer more options for formatting that type. For example the format `%f`, for `float` values, can also specify the precision: `%.2f`.
+
+Additionally, if you use the type specific placeholders and you refactor your code you will get notified 
+if you changed the type of the variable but not the type of the placeholder. This can e.g. help 
+to avoid breaking a string based api without noticing it.
```
kytrinyx: I'm not sure what the actual recommendation is here.

automated-comments/go/two_fer/wrong_function_comment.md
```diff
@@ -0,0 +1 @@
+The comment describing the function `ShareWith` isn't quite in the standard format. The section on comments below may help you fix this.
```
kytrinyx: > The function comment for `ShareWith` does not follow the Go convention. See the section on comments below.

automated-comments/go/two_fer/wrong_package_comment.md
```diff
@@ -0,0 +1 @@
+The comment describing the package `twofer` isn't quite in the standard format. The section on comments below may help you fix this.
```
kytrinyx: The package comment for `twofer` does not follow the Go convention. See the section on comments below.

SleeplessByte:
I hope its fair game to respond to the general gist of @kytrinyx here, as I am in the process of creating a similar PR, ignoring the more detailed, specific, inline comments:

> use as few words as necessary

I think this will lead to _too_ concise wording. I don't see the point in doing that when most of the time _more_ words can lead to _better_ commentary.

> be neutrally observational, avoiding charged statements and blanket statements

:100: % agreed

> make the recommendation explicit

:100: % agreed

> put the recommendation first, then the explanation

I think this makes sense for _most_ but not necessarily for all commentary. Is there a reason you want an action-bullet pointed-list? I am worried that this will lead to students _just applying it_ without actually **learning** from it.

> avoid "me", "I", "we", etc, since the bot is not a person

:100: % agreed

> avoid "you" because it can sometimes come across as judging the person rather than the code

Meh. Does "Your code" really sound better than "you"? Sounds like a blanket tone policing statement that probably should be looked at case by case? Is probably a good general suggestion overall tho :)

> avoid words like "just" or "simply" which can come across as condescending

:100: % agreed

> avoid making assumptions about what people know and don't know

:100: % agreed, except that I think we can assume that people have _learned_ from the previous exercises in the track. So it might apply to `two-fer`, but not to the next exercise. If you don't agree, that means we should change the guidelines to _repeat commentary and lessons_, which is okay, but that needs to be coordinated :)

kytrinyx:
> I think this will lead to _too_ concise wording. I don't see the point in doing that when most of the time _more_ words can lead to _better_ commentary.

In that case I would want to evaluate this on a case by case basis. I found when reading the above comment suggestions that a lot of the unnecessary words made it harder to parse for meaning.

> > put the recommendation first, then the explanation
> 
> I think this makes sense for most but not necessarily for all commentary.

Agreed.

> Is there a reason you want an action-bullet pointed-list? I am worried that this will lead to students just applying it without actually learning from it.

I don't understand what you mean here. I didn't suggest an action-bullet-pointed list anywhere, did I?

My comments are about a general approach. I'm sure there will always be exceptions, and we should discuss those as they come up.

> Meh. Does "Your code" really sound better than "you"? Sounds like a blanket tone policing statement that probably should be looked at case by case? Is probably a good general suggestion overall tho :)

No I hate "your code". I'd like to leave off the "you" entirely, in general.

kytrinyx:
> I think we can assume that people have _learned_ from the previous exercises in the track.

I think that's fair. I still wouldn't want to say "As you know". I hate it when books use this as a transition after explaining something "now that we all understand X". Just because you said it, doesn't mean I understood it.

kytrinyx:
> > use as few words as necessary
> 
> I think this will lead to _too_ concise wording. I don't see the point in doing that when most of the time _more_ words can lead to _better_ commentary.

Maybe instead of "use as few words as necessary", something like "avoid unnecessarily wordy comments"

SleeplessByte:
I like all your suggestions here. I'll write up a commentary guideline 🎩 and add it to the automated mentoring support repo.

> No I hate "your code". I'd like to leave off the "you" entirely, in general.

Got it. I'll add both of these examples explicitly 🎆 

> I don't understand what you mean here. I didn't suggest an action-bullet-pointed list anywhere, did I?

Ah, no you didn't, but when you put the action in front, a simplified, abstract, example will look somewhat like this:

- Change X using Y because of Z
- Remove A, because when you have A, you don't account for B
- Add P, as Input O will fail with Q.

I'm worried that this will be interpreted like this:

- **Change X using Y** ~because of Z~
- **Remove A**, ~because when you have A, you don't account for B~
- **Add P**, ~as Input O will fail with Q.~

Then again, this is something I would like to try and see if my worry has merit or not.

tehsphinx:
> When we refer to actual concrete variable names in the comment, do we know that they used that variable name in their code?

The variable `%{name}` is a placeholder for whatever variable name they actually used in the code. The analyzer parses the actual name from the AST.

> Do we only post the "comments section" when there is explicit feedback about some comments that don't match the recommendations?

Correct. That is what the analyzer currently does.

automated-comments/go/general/remove_extra_variable.md
```diff
@@ -0,0 +1 @@
+You are declaring the variable `%{name}` and only using it once. You can inline it instead.
```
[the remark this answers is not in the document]
tehsphinx: Should I make it as general as you phrased it or mention the actual variable name they used. Maybe:

> The variable `%{name}` is only used once after declaration. In that case, it is common to inline it, using the definition directly.

automated-comments/go/general/use_variable_assignment.md
```diff
@@ -0,0 +1 @@
+You are declaring variable `%{name}` only to assign a value to it afterwards. You can declare the variable by assigning a value to it in one step.
```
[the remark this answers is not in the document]
tehsphinx: Same here with mentioning the variable this is about. Suggestion:

> Rather than declaring the variable `%{name}` first, and then assigning a value to it afterwards, it is more common to declare and assign in a single statement.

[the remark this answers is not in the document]
tehsphinx: Note: We are in `general` section here which applies to all exercises. In later exercises it might not be obvious which variable this refers to otherwise.

automated-comments/go/hamming/declare_variable_when_needed_not_start_of_function.md
```diff
@@ -0,0 +1 @@
+You are declaring `%{returnVar}` before you check if the length is equal, so the declaration may end up being unnecessary. It's better to declare a variable right before you use it.
```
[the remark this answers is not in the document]
tehsphinx: As mention below: `%{returnVar}` is a placeholder for the actual variable name used by the student.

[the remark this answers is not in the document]
tehsphinx: Suggestion with variable name:

> Variably `%returngVar}` is declared early. It's common when writing Go to declare variables as late as possible before they are used. This tends to reduce mental overhead when reading the code, and sometimes also leads to simpler code, since the code might return before the variable is needed.

Note: Don't like this one too much though. Maybe someone has a better suggestion how it integrate the variable name here.

automated-comments/go/hamming/invert_if_for_happy_path_on_left.md
tehsphinx: Suggesting to add this link on "[The happy path is left-aligned](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88)".

automated-comments/go/hamming/minimize_slice_of_rune_conversions.md
tehsphinx: With the used variable name maybe:

> Prefer to convert `%{name}` to a `[]rune` only once, saving the result in a variable and reusing it. This conversion is a comparatively expensive operation.

kytrinyx:
> I'm worried that this will be interpreted like this:

Ah, right. Yeah, I wouldn't want it to go there. I do think that if we're not too wordy, people will naturally read the explanations, but we'll have to keep an eye on that.

automated-comments/go/hamming/rune_or_byte_to_string_conversion.md
tehsphinx: > For consistency, comparing either runes or bytes, but not both.

This sound grammatically strange to me. Shouldn't it be `compare` instead of `comparing`?

kytrinyx:
> I'll write up a commentary guideline 🎩 and add it to the automated mentoring support repo.

@SleeplessByte ❤️ 💛 💚 💙 💜 🖤 Thank you so much for taking that one on!

automated-comments/go/two_fer/missing_function_comment.md
tehsphinx: Yes, that is currently the case.

automated-comments/go/two_fer/missing_package_comment.md
tehsphinx: > It's idiomatic in Go for package to have a comment. For more details see the comments section below.

automated-comments/go/two_fer/use_fmt_placeholder_for_string.md
tehsphinx: To rather use type specific identifiers than the general `%v`. If e.g. the type of the variable changes during some refactoring, `%v` might silently "fail" while a type specific placeholder will alarm you (or better the linters will).

automated-comments/go/general/strings_trim_space_used.md
```diff
@@ -0,0 +1,5 @@
+Don't worry about cleaning up input in exercises on exercism. The platform does not require that.
+
+Even if this were a real application all the functions within your applications should not clean up
+input data. If you have potentially bad data coming from a source you should clean it up in the code reading
+from that source. Then you can consider the data clean within the rest of your code.
```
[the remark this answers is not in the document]
tehsphinx: Made a small adjustment replacing the second `the input` with `it`. Hope that is fine.

automated-comments/go/hamming/error_msg_not_capitalized_nor_punctuated.md
```diff
@@ -0,0 +1 @@
+In Go, error strings are often chained together to form a kind of 'breadcrumb trail' reflecting the chain of function calls that led to the error. For this reason it is recommended they are [not capitalized or punctuated](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).
```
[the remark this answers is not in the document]
tehsphinx: I like `weird` 😄

automated-comments/go/hamming/use_of_naked_returns.md
tehsphinx: Was it intentional that you included the `official` in the link?

automated-comments/go/general/remove_extra_variable.md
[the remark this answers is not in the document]
iHiD: That's nice, @tehsphinx.

automated-comments/go/general/section_about_comments.md
```diff
@@ -0,0 +1,27 @@
+
+---
```
iHiD: Is there a reason for this whitespace and `---` here?

automated-comments/go/general/use_variable_assignment.md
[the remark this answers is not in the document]
iHiD: @tehsphinx Yes - interpolating is good.

automated-comments/go/hamming/extra_if_strings_empty.md
```diff
@@ -0,0 +1,2 @@
+The extra check if the `strings` are empty is not wrong but does not add value either. It would make the function
+more readable if it was removed.
```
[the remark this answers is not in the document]
iHiD: I prefer the passive voice to the imperative case here to make it more friendly. So "This can be simplified by removing..." rather than "Simplify by removing...".

I think it:
- Feels friendlier
- Makes someone think about what needs doing, rather than following an instruction.

If @kytrinyx agrees, can this get added to the guidance please, @SleeplessByte.

automated-comments/go/hamming/invert_if_for_happy_path_on_left.md
iHiD: Nice! :)

automated-comments/go/hamming/extra_if_strings_empty.md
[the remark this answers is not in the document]
SleeplessByte: Certainly. Personally I'm not a fan of the passive voice but I realise this is probably just cultural difference. In general, if you tell a Dutch person (in my experience) that they _can_ or _could_ do something, they'll probably respond with something like "ok thnx, I'll leave it".

I don't have a good solution 😓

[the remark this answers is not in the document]
iHiD: Yeah, I also agree. But as these comments are for approval, then we are giving people the choice of doing it or not. I don't think we should be giving people strong instructions when we're also saying "This is fundamentally fine". 

I would suggest "You can simplify..." but but I agree that we want to avoid "You".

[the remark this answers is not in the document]
SleeplessByte: Yeah. I think in _this particular case_ "This can be simplified by removing ..." works well. 

This is not easy 😅

iHiD:
@tehsphinx Can the folders match the [track exercise directory structure](https://github.com/exercism/go/tree/master/exercises/two-fer) pls (ie `two-fer` not `two_fer` in this case)

(cc @SleeplessByte for guidelines)

automated-comments/go/general/section_about_comments.md
tehsphinx: That was to separate it from the other comments in exalysis. Will remove it as this will probably be done differently with the automentoring.

tehsphinx:
@kytrinyx: I applied your suggested changes and resolved all conversations where there was nothing to add or reply. Can you have a look at the open conversations and close or comment on them?

Thanks a lot for the review!

automated-comments/go/general/remove_extra_variable.md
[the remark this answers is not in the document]
kytrinyx: @tehsphinx since we'll use whatever variable name they used, I think we should be more specific!

automated-comments/go/two_fer/use_fmt_placeholder_for_string.md
kytrinyx: Ah! That's great. Let me see if I can try rephrasing based on that...

> In Go people tend to prefer type-specific placeholders rather than the generic `%v`. There are two reasons for this. First, more specific placeholders often have additional options, for example the format `%f` for `float` values can also specify the precision: `%.2f`. The second reason is that `%v` accepts anything. If a variable type changes during a refactoring, `%v` will print the default for that type, which might not be the desired behavior. A type-specific placeholder will cause the linter to complain, catching the error before it gets deployed to production.

I don't know. Suggestions/improvements welcome :-)

kytrinyx:
> I was still too much thinking in terms of exalysis where a mentor actually takes and posts the comments.

That makes perfect sense! The conversational nature works really well in that context.

This all looks great to me ✨

automated-comments/go/general/remove_extra_variable.md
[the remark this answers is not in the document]
tehsphinx: @kytrinyx: This is still a `general` comment used for whenever there is a variable that can be inlined. Not sure how we can be more specific apart from mentioning the variable name they used. Can you clarify?

automated-comments/go/two_fer/use_fmt_placeholder_for_string.md
tehsphinx: I like it. Updated and replaced `anything` with `any value type`.

automated-comments/go/hamming/extra_if_strings_empty.md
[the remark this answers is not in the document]
tehsphinx: adjusted to `This can be simplified by removing`. I think it is more gentle.

automated-comments/go/general/section_about_comments.md
iHiD: Brilliant. Thanks :)

automated-comments/go/general/remove_extra_variable.md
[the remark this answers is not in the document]
SleeplessByte: I think Katrina means that the version _with_ the name is better than the _general_ comment.

In JS/TS I usually do give the code example in the earlier exercises, e.g. when inline-ing `export colorCode` I print that after telling them that `exports can be inlined`. Perhaps we want/not want that?

automated-comments/go/hamming/extra_if_strings_empty.md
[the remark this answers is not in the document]
kytrinyx: I agree! (Sorry, I completely missed this as it was in a collapsed section in the PR).

automated-comments/go/hamming/declare_variable_when_needed_not_start_of_function.md
```diff
@@ -0,0 +1,4 @@
+Variably %{returngVar} is earlier than necessary.
```
kytrinyx: I'm not sure what this means. Is it supposed to be something like this?

```
Variable %{returngVar} is defined earlier than necessary.
```

automated-comments/go/hamming/invert_if_for_happy_path_on_left.md
```diff
@@ -0,0 +1,6 @@
+To simplify, invert the `if` condition handling the error case, 
+then get rid of the `else` branch by returning within the `if`.
+A Go proverb says "[The happy path is left-aligned](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88)",
+meaning the normal flow of control follows the left-hand margin, 
+and only exceptional or error cases are indented.
+
```
kytrinyx: The other files don't have a trailing empty line. I doubt it's a problem, but if we want to be consistent we could ✂️ it.

automated-comments/go/hamming/mixture_of_runes_and_bytes.md
```diff
@@ -0,0 +1,3 @@
+Using a mixture of `runes` and `bytes` is safe here, because the input only uses ASCII characters. 
+To see what happens when it's not safe, try adding a test that uses non-ASCII characters.
+
```
kytrinyx: ✂️ ?

automated-comments/go/hamming/use_zero_values_on_error_return.md
```diff
@@ -0,0 +1,7 @@
...
+
+By convention, when returning an `error` value, other values are not used,
+and it's considered idiomatic to return the [zero value](https://golang.org/ref/spec#The_zero_value) in this situation.
+The `error` value makes using an invalid "out of band" value to signal an error redundant in Go.
```
kytrinyx: This one seems to have some duplication. Is that intentional?

automated-comments/go/two-fer/plus_used_for_concatenation.md
```diff
@@ -0,0 +1,3 @@
+In this case using `+` is almost certainly the right choice, as it's shorter in this case,
```
kytrinyx: Here we have "in this case" twice in the same phrase. Maybe use "here" for one of them?

automated-comments/go/hamming/declare_variable_when_needed_not_start_of_function.md
tehsphinx: Yes, thx!

automated-comments/go/hamming/use_zero_values_on_error_return.md
tehsphinx: Missed deleting the old one.

automated-comments/go/two-fer/plus_used_for_concatenation.md
tehsphinx: Changed it to
`In this case using `+` is almost certainly the right choice, as it's shorter,
and often faster for simple concatenations.`

Does that sound good?

tehsphinx: Not sure if the `,` before the `and` should stay. In German it would have to be removed.

iHiD: Grammatically it's fine with or without. I think I prefer it being there as it adds clarity that the shortness and fastness are two separate concepts worth taking a breath in between.

automated-comments/go/general/taking_length_of_string_to_check_empty.md
```diff
@@ -0,0 +1,2 @@
+Rather than declaring the variable %{name} first, and then assigning a value to it afterwards,
```
ErikSchierboom:
```suggestion
Rather than declaring the variable `%{name}` first, and then assigning a value to it afterwards, 
```

iHiD:
@tehsphinx how close is this to merging?

tehsphinx:
@iHiD: From my side this can be merged

tehsphinx:
@iHiD: `v0.4.0` incorporates the changes needed for comments to work: https://github.com/exercism/go-analyzer/releases

That means:
- rename `two_fer` to `two-fer` in comment handles so they map to the folder structure by simply replacing `.` with `/`
- switch to `approve` / `disapprove` states

PR #1013, opened 2019-05-12
Add dheater as a mentor for the C track

dheater:
Adding myself as a mentor


mentors/c/h.json
```diff
@@ -1 +1,10 @@
-[]
+[
+  {
+    "github_username": "dheater",
+    "name": "Daniel Heater",
+    "link_text": ,
```
rpottsoh:
```suggestion
    "link_text": null,
```

rpottsoh:
I missed this earlier.  Your user name starts with letter d so you need to be modifying d.json, not h.json.......

PR #1014, opened 2019-05-15
[Java/Allergies] Add mentoring notes

michael-berger-FR:
This short exercise should be solved with streams. Other approaches are more complex and less elegant.


SleeplessByte:
It's been almost a month. I'm merging this because having something is better than nothing. @exercism/website-copy-java please make subsequent edits in subsequent PRs.

PR #1023, opened 2019-05-17
[Java/Word Count] Add mentoring notes

michael-berger-FR:


tracks/java/exercises/word-count/mentoring.md
```diff
@@ -0,0 +1,51 @@
...
+            }
+            Integer count = countMap.get(aWord);
+            countMap.put(aWord, count == null ? 1 : count + 1);
+        }
+        return countMap;
+    }
+}
+```
+
+FP solution (Java 11):
```
SleeplessByte: Consider writing out FP as 
```suggestion
Functional programming (FP) solution (Java 11+):
```

jmrunkle: I actually think most (if not all) of this actually works starting in Java 8, but I could be mistaken.

tracks/java/exercises/word-count/mentoring.md
```diff
@@ -0,0 +1,51 @@
...
+
+    public Map<String, Integer> phrase(String input) {
+        Map<String, Integer> countMap = new HashMap<>();
+        input = input.trim().toLowerCase().replaceAll("([^A-Za-z0-9']|\\B'|'\\B)", " ");
+        String[] tokenizedInput = input.split("\\s+");
+        for (String aWord : tokenizedInput) {
+            if (aWord.isEmpty()) {
+                continue;
+            }
+            Integer count = countMap.get(aWord);
```
jmrunkle: Map has a merge method that is probably better for this use case.

tracks/java/exercises/word-count/mentoring.md
```diff
@@ -0,0 +1,51 @@
...
+
+public class WordCount {
+
+    public Map<String, Integer> phrase(String data) {
+        return Arrays.stream(data.split("[ \n\\p{Punct}&&[^\\']]"))
+                .filter(Predicate.not(String::isEmpty))
+                .map(this::stripQuotes)
+                .map(String::toLowerCase)
+                .map(word -> new AbstractMap.SimpleEntry<>(word, 1))
+                .collect(groupingBy(entry -> entry.getKey(), summingInt(entry -> entry.getValue())));
```
jmrunkle: Drop the previous line and make this change:

```suggestion
                .collect(groupingBy(identity(), summingInt(word -> 1)));
```

tracks/java/exercises/word-count/mentoring.md
michael-berger-FR: @jmrunkle, I agree that **most** of this working in Java 8. But `java.util.function.Predicate.not;` is Java 11... So what should I do? ;-)

tracks/java/exercises/word-count/mentoring.md
michael-berger-FR: Totally true!

tracks/java/exercises/word-count/mentoring.md
SleeplessByte: Perhaps you can add that as a helper with the comment "Use java.util.function.Predicate.not in Java 11+"

jmrunkle: I guess I had forgotten that they didn't add that until Java 11. My company uses Java 8, but uses the Guava method so I just did not really notice: https://google.github.io/guava/releases/23.0/api/docs/com/google/common/base/Predicates.html#not-com.google.common.base.Predicate-

PR #1029, opened 2019-05-22
Update "github_username" from astoehr to stoehraj

stoehraj:
I migrated all Github data from my old account (astoehr) to my new one (stoehraj) and have since deleted the old account.


stoehraj:
> 
> 
> Please move the data object to the correct file, `s.json`.

Whoops, my mistake!  Just fixed it.

PR #1031, opened 2019-05-22
Add Rob Kendal to mentor list

bpk68:


mentors/javascript/b.json
```diff
@@ -46,5 +46,13 @@
     "link_url": "https://michael.plotke.me/",
     "avatar_url": null,
     "bio": "I'm paid to write JavaScript. I mostly enjoy it. I also enjoy teaching. Exercism is a nice convergence."
-  }
+  },
+  {
+  "github_username": "bpk68",
+  "name": "Rob Kendal",
+  "link_text": "robkendal.co.uk",
```
rpottsoh: Link text should be equal to `null,` since it is the same as link_url.
```suggestion
  "link_text": null,
```

bpk68:
All updated for you :D

PR #1032, opened 2019-05-22
rust/bob: adds mentoring notes

davidwessman:
- Based on a few solutions and the feedback I got by my mentor.

### DIFF
--- a/tracks/rust/exercises/bob/mentoring.md
+++ b/tracks/rust/exercises/bob/mentoring.md
@@ -1,0 +1,37 @@
+### Reasonable solutions
+
+A reasonable solution should:
+
+* Use `.any()` or `.all()` to check for yelling and for alphabetic characters.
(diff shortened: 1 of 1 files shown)


SleeplessByte:
> I've approved the changes with the belief that if other Rust contributors should disagree, it is easier for them to notice with those notes posted on the website. And if there are other idiomatic Rust solutions that don't use a truth-table based approach, there is room for multiple reasonable solutions. My only point with the comment above was that I wouldn't find it as reasonable in an ML-like language (strict or non-strict) as various alternatives.

Yes, having some notes is better than none. If @exercism/website-copy-rust would like some changes lets do it in subsequent PRs.

davidwessman:
Have not been able to respond, on vacation 🌞

I will look at the feedback more closely when I'm at a computer. Thank you @sshine!

And further improvements sounds like a good way to move forward!

PR #1037, opened 2019-05-24
Add Semsudin's biography.

semsuddin:


semsuddin:
Sure, check it out.

rpottsoh:
One more knit pick.  Please change the commit comment on the second commit to just state that it is fixing the formatting.  Your first commit added your bio, essentially.  It is preferred that a commit have a single purpose, so no `ands`.

semsuddin:
Knit picky but a great suggestion nonetheless. Thanks. :)

PR #1044, opened 2019-05-28
csharp/clock - ensure fields are immutable

mikedamay:
https://exercism-team.slack.com/archives/CASDGG84B/p1558971847010000?thread_ts=1558935598.009300&cid=CASDGG84B

### DIFF
--- a/tracks/csharp/exercises/clock/mentoring.md
+++ b/tracks/csharp/exercises/clock/mentoring.md
@@ -83,7 +83,7 @@ public class Clock
 - The clock can be implemented as a struct or a class.  The main practical advantage of the struct
  in this exercise is that equality is handled automatically for the struct
 - To pad the string "00" and "D2" (where D stands for decimal) are equally valid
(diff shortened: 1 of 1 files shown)


tracks/csharp/exercises/clock/mentoring.md
```diff
@@ -83,6 +83,7 @@ public class Clock
 - The clock can be implemented as a struct or a class.  The main practical advantage of the struct
  in this exercise is that equality is handled automatically for the struct
 - To pad the string "00" and "D2" (where D stands for decimal) are equally valid
+- If the student has modified any of fields of the clock (as opposed to simply creating a new clock) then it is important to point out that this is not compatible with `GetHashCode()` which requires all fields used in its calculation to be immutable.
```
ErikSchierboom:
```suggestion
- If the student has modified any of the fields of the clock (as opposed to simply creating a new clock), then it is important to point out that this is not compatible with `GetHashCode()` which requires all fields used in its calculation to be immutable.
```

ErikSchierboom:
Thanks! 🎉

PR #1046, opened 2019-05-29
javascript/leap: Update mentoring.md

zeckdude:
When the sample response is copied to the text field on the site, it is adding a line break that needs to be removed. This PR fixes that.

### DIFF
--- a/tracks/javascript/exercises/two-fer/mentoring.md
+++ b/tracks/javascript/exercises/two-fer/mentoring.md
@@ -48,8 +48,7 @@ conditional _and_ use a template literal.
 
     Hi there {name},
 
-    Congrats on submitting your first exercise for the JavaScript
-    track. I have a few comments for you:
+    Congrats on submitting your first exercise for the JavaScript track. I have a few comments for you:
 
     - In Javascript, the result of an expression can be returned!
       ```javascript


SleeplessByte:
I don't think it does. Linebreaks are stripped. The reason is there is to keep the box from scrolling and limit the content to 80 chars width.

## Preview after copying

image

## Actual display in comment after submission

image

NobbZ:
In markdown, both of the following snippets should produce identical results when rendered:

```md
foo
bar
```

---

```md
foo bar
```

Your change is similar to these examples and wouldn't change anything in the rendered output after copying.

If though it does in fact change the rendered result on page, this is worth another bug report on exercism/exercism.

SleeplessByte:
I've confirmed the behaviour as reported to be correct.

zeckdude:
When you highlight the copy from the modal to the feedback field, you don’t see the unnecessary line break?  I’ll create an animated gif so you can see what I’m talking about.

SleeplessByte:
It's like the screenshot in my first response.

zeckdude:
@SleeplessByte @NobbZ This is what happens for me when I copy and paste from the sample text. You can see it copy a line break. I haven't left the line break in there to see if it actually keeps it when I write a comment to a student, but the OCD in me would like for it not to be there. Are you telling me to just ignore it?

unnecessary_linebreak

SleeplessByte:
Ah, smaller screen is what it is. I kinda want those code lines to stay at max 80 chars, which is why it's there in the first place (so you don't need to scroll). 

The problem is, when we _remove_ the line break, the mentor notes becomes scrollable (except for the link doing that already), but it means not all text is visible right away, which is what I was trying to avert.

If you can _live_ with removing the linebreak manually, I rather have that until we have a better system of providing copyable templates ❤️ as a site-feature.

zeckdude:
Yeah for sure. No problem. Makes sense.

NobbZ:
Yes, there is a linebreak in the textbox, but you can ignore it! The markdown renderer will treat is as a “regular” space and not a paragraph. The rendered HTML will **not** have that linebreak.

Please check the “preview” pane.

SleeplessByte:
> Please check the “preview” pane.

Yes that's like my screenshot :)

PR #1047, opened 2019-05-29
erlang/leap: Fix sub-heading level

juhlig:
The sub-heading "Using a single expression" was on the wrong level, missing one `#`.


SleeplessByte:
Thank you :heart:

PR #1049, opened 2019-05-29
javascript/resistor-color: mention time complexity and premature-optimisation

jocelo:
Adding a reasonable solution when the student is taking into consideration the time complexity of the function, as the student is capable of handling the problem proposed and they are starting to think in more advanced topics.

### DIFF
--- a/tracks/javascript/exercises/resistor-color/mentoring.md
+++ b/tracks/javascript/exercises/resistor-color/mentoring.md
@@ -15,32 +15,6 @@ export const colorCode = color => COLORS.indexOf(color)
 A student may use a function declaration instead of an arrow function expression, as well as using an explicit `return`
 instead of an implict return.
 
-A student may use an object to store color names instead of an array because they might be taking into consideration time complexity. Reach out to the student to ask if this is the case.
-A pretty solid guess about this is when the student is using [`Object#keys`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys) to return the list of colors as an array.
-
-The solution should look something similar to this:
-
-```javascript
-const colorsCode = {
-    'black': 0,
-    'brown': 1,
-    'red': 2,
-    'orange': 3,
-    'yellow': 4,
-    'green': 5,
-    'blue': 6,
-    'violet': 7,
-    'grey': 8,
-    'white': 9
-}
-
-export const COLORS = Object.keys(colorsCode)
-
-export function colorCode(color) {
-    return colorsCode[color]
-}
-```
-
 ### Common suggestions
 
 If the solution does not look like the reasonable solution, do _not_ approve it. Here are a few common suggestions based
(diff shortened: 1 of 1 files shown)


jocelo:
The more I get involved in this project the more I'm enjoying it... thanks for your feedback @SleeplessByte, I understand and agree with your comments. Indeed on the performance testing I only saw a difference of 5% tops... so you think is worth it to leave a transcribed version of this discussion somewhere in the mentor notes?

I think it might be of value for mentors, since I personally have seen a couple of students using an object, and when I asked the reason behind this proposal, the student gave me the time complexity explanation.

If you think we should keep some mention to this in mentor notes, what would it be the best section for it (Reasonable solutions, Common suggestions, Talking points)? ...maybe as a talking point ?

Either way, I know now how to approach a student like this in the future.

Note: This is the fist of the students I saw doing this:
https://exercism.io/mentor/solutions/58d387fd60dc48a68d06530f6547b731

SleeplessByte:
https://github.com/exercism/problem-specifications/issues/1524 was actually opened a few hours ago due to a similar comment (yet different) on the ruby specific notes. We're aiming to **also** change the README.md so it's less inclined to point to an object.

That said:

> [...] so you think is worth it to leave a transcribed version of this discussion somewhere in the mentor notes?

> If you think we should keep some mention to this in mentor notes, what would it be the best section for it (Reasonable solutions, Common suggestions, Talking points)? ...maybe as a talking point ?

Yes please! This is a great suggestion. It should actually be a talking point; you can use the perf link I added. You **could** choose to put in under a sub-heading under `### Reasonable solutions` similar to [these notes](https://github.com/exercism/website-copy/blob/master/tracks/javascript/exercises/resistor-color-duo/mentoring.md#approvability), but at the minimum mention the variation using an `Object` straight-away and mentioning that it's _not_ preferred over the `Array` version.

> I think it might be of value for mentors, since I personally have seen a couple of students using an object, and when I asked the reason behind this proposal, the student gave me the time complexity explanation.

Yes, I've seen this as well and it's a _great_ opportunity to teach them to write code in context. A lot of people, especially juniors and mediors tend to either abstract or optimise way too early or try to nail everything with their collection of hammers. In this case, you can objectively **show** that this is a premature optimisation (we're talking about 1.9 million vs 2.1 million ops **a second**).

jocelo:
Applied change proposed..., let me know if you see something else worth discussing!

tracks/javascript/exercises/resistor-color/mentoring.md
```diff
@@ -64,3 +64,35 @@ export function colorCode(color) {
 They may not know of the _concept_ of `Array`. You may gently ask how experienced they are with programming. **Note**
 that there _are_ languages that do **not** have arrays. So them not knowing about it does not mean they don't have any
 experience.
+
+#### Time Complexity
+
+A student may use an `Object` instead of an `Array` to store color names with the intention to optimize the execution time.
+This is **not** cause for dissaproval but it's a good opportunity to discuss with the student about not to prematurely optimize code.
```
tejasbubane: Typo `disapproval`

jocelo: Typo fixed. thanks!

SleeplessByte:
Thank you @jocelo for this additions. I think it definitely makes sense to have it and I appreciate the work you did :sparkles:

jocelo:
Happy to help @SleeplessByte ! 

For sure! I'll create another PR for those changes

PR #1051, opened 2019-05-30
ruby/matrix: Do not encourage bloated `initialize` methods

yyyc514:
Could we have a discussion about this?  I think typically this is bad behavior and it's much better design to do the work when methods are called.  This pattern can lead to all sorts of performance problems latter (doing too much work). It's also a pretty serious violation of SRP.

Thoughts?

### DIFF
--- a/tracks/ruby/exercises/matrix/mentoring.md
+++ b/tracks/ruby/exercises/matrix/mentoring.md
@@ -2,10 +2,16 @@
 
 ```ruby
 class Matrix
-  attr_reader :rows, :columns
   def initialize(matrix_as_string)
-    @rows = extract_rows(matrix_as_string)
-    @columns = rows.transpose
+    @matrix_as_string = matrix_as_string
+  end
+  
+  def rows
+    extract_rows(@matrix_as_string)
+  end
+  
+  def columns
+    rows.transpose
   end
 
   private


yyyc514:
Obviously we'd update the notes also if we decided to change the best practice example, I just wanted to start the discussion and see where it goes first.

yyyc514:
Ugh, whatever but it seems confusing to me that the repository that has the Ruby exercise notes isn't the best place to discuss changing the Ruby exercise notes.

Seems perhaps another argument for putting the mentor notes in the track repositories, no?

SleeplessByte:
> Uh, whatever but it seems confusing to me that the repository that has the Ruby exercise notes isn't the best place to discuss changing the Ruby exercise notes.
> 
> Seems perhaps another argument for putting the mentor notes in the track repositories, no?

To me these are two separate things: @F3PiX is asking to have the discussion about the `initialize` methods in the ruby respository so that people with good understanding of Ruby can all discuss. It doesn't really have anything to do with the mentor notes, as it's more about what we consider an optimal solution.

Once a consensus is reached, a PR should be opened on the website-copy repo to reflect that consensus in terms of writing.

SleeplessByte:
> But on the other hand, we moved them here to relieve track maintainers of also becoming note maintainers.

And _most_ maintainers are not mentors and therefore are also not the best candidates to maintain those notes.

The website-copy-language teams should have solved it, but most of them have one or no members.

yyyc514:
> Yes. But on the other hand, we moved them here to relieve track maintainers of also becoming note maintainers. I'm personally a proponent of keeping the notes in the track repositories; as a track maintainer who wants to keep the mentor notes for the tracks I participate in up-to-date, subscribing to all track updates in website-copy led me to have to create some email filters to avoid feeling spammed.

Completely agree, though I don't know/understand the politics for the previous move.  It's just frustrating to open an issue for notes and then be told that despite the notes being here (very intentionally it seems) the best place to actually open an issue to discuss real changes to notes is in the track, etc...  feels like something is broken somewhere.

sshine:
>> Additionally, what @rpottsoh and I do will basically become impossible if they move back to the track repos.
>
> What do you all do that would become impossible?

Having the mentor notes in a central places makes it possible for a small team to provide quality control. @rpottsoh and @SleeplessByte both do a lot of this for a lot of tracks they're not otherwise affiliated with. I'm not as heroic, but my efforts are also widespread among languages. Writing teaching material does require knowledge of the programming language, but as you approve a bunch, you start to see common patterns of best practice.

As long as this repo doesn't drown in mentor notes PRs, there are definite pros to this.

@yyyc514, @F3PiX : Now that the discussion in exercism/ruby#962 has been had, and #1149 has been submitted, should this PR be merged or closed?

SleeplessByte:
Right, to elaborate on @sshine is that I currently get between 20 and 250 emails _a day_ from `exercism` related repositories, which are from the ones I maintain/contribute to. A lot of the quality control is gramma, spell check, format, things a copy team usually takes care of. If this were living in each _language_ repository, Those 250 emails can easily become 2500 (as we have ~50 language tracks). At that point I'll say no.

yyyc514:
Yeah, I think #1149 closes this.

PR #1059, opened 2019-06-01
javascript/resistor-color: Update mentoring.md

zeckdude:
Add acceptable solution for resistor-color exercise

### DIFF
--- a/tracks/javascript/exercises/resistor-color/mentoring.md
+++ b/tracks/javascript/exercises/resistor-color/mentoring.md
@@ -15,16 +15,6 @@ export const colorCode = color => COLORS.indexOf(color);
 A student may use a function declaration instead of an arrow function expression, as well as using an explicit `return`
 instead of an implict return.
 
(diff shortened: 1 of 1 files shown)


tracks/javascript/exercises/resistor-color/mentoring.md
```diff
@@ -9,18 +9,28 @@ export const COLORS = [
...
 ];
 
-export const colorCode = color => COLORS.indexOf(color)
+export const colorCode = color => COLORS.indexOf(color);
 ```
 
 A student may use a function declaration instead of an arrow function expression, as well as using an explicit `return`
 instead of an implict return.
 
+Using the [`Array#findIndex`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) method is also acceptable for the `colorCode()` method but remind the student that [`Array#indexOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf) only requires a simple string as the argument whereas `Array#findIndex` requires a function to determine the index.
```
SleeplessByte: We really want people to use the indexOf. It's not reasonable to use findIndex in this exercise. I thought it was clear that it felt under manually doing indexOf, but I think your addition makes sense to add!

Can you move this to a common suggestion heading?

PR #1065, opened 2019-06-04
Haskell/Leap: Add Data.Time.Calendar.isLeapYear solution

sshine:
Since there's a popular library, `time`, that exposes exactly this function, an idiomatic solution is to re-export this function.


SleeplessByte:
🚀

PR #1066, opened 2019-06-04
Mentoring guideline for erlang/bob

juhlig:
As there is a multitude of ways this can be solved, each rather large, so I just put the one that I think is the best, give or take...

### DIFF
--- a/tracks/erlang/exercises/bob/mentoring.md
+++ b/tracks/erlang/exercises/bob/mentoring.md
@@ -1,0 +1,46 @@
+### Reasonable solutions
+
+```erl
+response(Str) ->
+  response(Str, false, false, false, false).
+
+response([], false, false, false, false) ->
+  "Fine. Be that way!";
+response([], true, false, _HasOther, false) ->
+  "Whoa, chill out!";
+response([], true, false, _HasOther, true) ->
+  "Calm down, I know what I'm doing!";
+response([], _HasUpper, _HasLower, _HasOther, false) ->
+  "Whatever.";
+response([], _HasUpper, _HasLower, _HasOther, true) ->
+  "Sure.";
+response([C|Str], HasUpper, HasLower, HasOther, MaybeQuestion) when C=<$\s ->
+  response(Str, HasUpper, HasLower, HasOther, MaybeQuestion);
+response([$?|Str], HasUpper, HasLower, HasOther, _MaybeQuestion) ->
+  response(Str, HasUpper, HasLower, HasOther, true);
+response([C|Str], _HasUpper, HasLower, HasOther, _MaybeQuestion) when C>=$A, C=<$Z ->
+  response(Str, true, HasLower, HasOther, false);
+response([C|Str], HasUpper, _HasLower, HasOther, _MaybeQuestion) when C>=$a, C=<$z ->
+  response(Str, HasUpper, true, HasOther, false);
+response([_|Str], HasUpper, HasLower, _HasOther, _MaybeQuestion) ->
+  response(Str, HasUpper, HasLower, true, false).
+```
+
+### Common suggestions
+
+* If the student is using regular expressions, ask for a solution that works
+  without them.
+* If the student is using the `string` module, ask for a solution that works
+  without it. An exception could be if the student is trying to accomodate for
+  unicode input.
+* If the student's solution requires traversing the input string multiple
+  times (to check for uppercase letters, lowercase letters, ...), ask for
+  a solution that requires only one pass.
+
+### Talking points
+
+* Regular expressions vs pattern matching
+
+<!-- Local Variables: -->
+<!-- mode: gfm -->
+<!-- End: -->


tracks/erlang/exercises/bob/mentoring.md
```diff
@@ -0,0 +1,46 @@
...
+  response(Str, HasUpper, HasLower, HasOther, MaybeQuestion);
+response([$?|Str], HasUpper, HasLower, HasOther, _MaybeQuestion) ->
+  response(Str, HasUpper, HasLower, HasOther, true);
+response([C|Str], _HasUpper, HasLower, HasOther, _MaybeQuestion) when C>=$A, C=<$Z ->
+  response(Str, true, HasLower, HasOther, false);
+response([C|Str], HasUpper, _HasLower, HasOther, _MaybeQuestion) when C>=$a, C=<$z ->
+  response(Str, HasUpper, true, HasOther, false);
+response([_|Str], HasUpper, HasLower, _HasOther, _MaybeQuestion) ->
+  response(Str, HasUpper, HasLower, true, false).
+```
```
sshine: You say that there are many reasonable solutions. If this is the average size of a reasonable solution, I wouldn't mind if there were 1-3 of them. But I understand your incentive.

As for this particular solution, I believe that you can make it more abstract and thus more readable if you apply meaningful names in either helper functions or variables. The code doesn't read well from top to bottom with the first many default cases being nonsensical until you reach the bottom.

I see that the [example solution](https://github.com/exercism/erlang/blob/master/exercises/bob/src/example.erl) does this.

Perhaps, if you'd like a single-pass solution that does no trimming, it could export the particular properties that `response` needs; something like...

```erlang
response(Str) ->
  {IsSilent, IsShouting, IsQuestion} = classify(Str),
  if IsSilent -> "Fine. Be that way!";
     IsShouting andalso IsQuestion -> ...;
     ...
  end.
```

or an equivalent case-of.

tracks/erlang/exercises/bob/mentoring.md
```diff
@@ -0,0 +1,46 @@
...
+### Common suggestions
+
+* If the student is using regular expressions, ask for a solution that works
+  without them.
+* If the student is using the `string` module, ask for a solution that works
+  without it. An exception could be if the student is trying to accomodate for
+  unicode input.
+* If the student's solution requires traversing the input string multiple
+  times (to check for uppercase letters, lowercase letters, ...), ask for
+  a solution that requires only one pass.
```
sshine: But that's not what's in the [example solution](https://github.com/exercism/erlang/blob/master/exercises/bob/src/example.erl); are you sure that Erlang mentors agree on this?

I don't know if you can say that a multiple-pass solution is worse on all accounts.

Because each individual predicate's definition is isolated into its own function, such solutions tend to be very readable.

So it may conversely just as well be a case of premature optimization to seek the "fused" solution.

I remember from taking a course in Standard ML at university that after having completed an exercise, I'd realize that I was performing a lot more traversals than I needed, after which point I sought to meld every single logical step into one big fold. The code became unreadable, but used the fewest amount of traversals. When taking the Haskell course, I realized that I could get such fusion for free without having to think about it. Since Erlang does not provide such fusion, one is left with a pragmatic choice.

tracks/erlang/exercises/bob/mentoring.md
sshine: Or... one that used atoms rather than booleans for the classification, I suppose.

tracks/erlang/exercises/bob/mentoring.md
```diff
@@ -0,0 +1,46 @@
...
+  response(Str, HasUpper, HasLower, true, false).
+```
+
+### Common suggestions
+
+* If the student is using regular expressions, ask for a solution that works
+  without them.
+* If the student is using the `string` module, ask for a solution that works
+  without it. An exception could be if the student is trying to accomodate for
+  unicode input.
```
sshine: Also, why should the student seek to avoid the `string` module?

One purpose of Exercism is to gain language fluency, and a part of that is using the standard library.

Is this entirely because there aren't any `string` functions that allow for a single-pass solution?

tracks/erlang/exercises/bob/mentoring.md
```diff
@@ -0,0 +1,46 @@
...
+  times (to check for uppercase letters, lowercase letters, ...), ask for
+  a solution that requires only one pass.
+
+### Talking points
+
+* Regular expressions vs pattern matching
+
+<!-- Local Variables: -->
+<!-- mode: gfm -->
+<!-- End: -->
```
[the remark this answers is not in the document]
NobbZ: They allow Emacs to choose the correct flavor of markdown.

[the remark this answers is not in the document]
sshine: Should we perhaps not have editor-specific comments?

sshine:
@juhlig: I did not mean to block this PR with my comments. Since @NobbZ has already approved it, I'll do the same and the notes can be improved subsequently.

juhlig:
@sshine and sorry I didn't reply to them =^^= Didn't see the notifications in my Mailbox, I guess...

PR #1070, opened 2019-06-06
Update z.json

zeckdude:
Add mentor profile for Chris Seckler

### DIFF
--- a/mentors/javascript/z.json
+++ b/mentors/javascript/z.json
@@ -2,7 +2,7 @@
   {
     "github_username": "zeckdude",
     "name": "Chris Seckler",
-    "link_text": "https://github.com/zeckdude",
+    "link_text": null,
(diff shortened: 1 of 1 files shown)


mentors/javascript/z.json
```diff
@@ -1 +1,10 @@
-[]
+[
+  {
+    "github_username": "zeckdude",
+    "name": "Chris Seckler",
+    "link_text": "https://github.com/zeckdude",
```
NobbZ:
```suggestion
    "link_text": null,
```

This uses the `link_url` as label for the link automatically and will avoid forgetting updating the label when you need to change the URL later on.

zeckdude: Ok sounds good. Btw, how did you do the commit suggestion in Github? Looks cool!

Screen Shot 2019-06-10 at 10 13 37 AM

PR #1073, opened 2019-06-10
Add mentoring notes for Scala two-fer

ricemery:


tracks/scala/exercises/two-fer/mentoring.md
```diff
@@ -0,0 +1,12 @@
...
+```scala
+object Twofer {
+  def twofer(name: String = "you"): String =
+    s"One for $name, one for me."
+}
+```
+
+### Common suggestions
+
+- Use default parameters instead of multipe `twofer` functions.
```
ErikSchierboom: I don't know if this happens a lot, but for C# a lot of people construct the dynamic string without using interpolation (using `+` or `String.format`). If this is common in Scala, it might be useful to also add a comment about this too.

ricemery: Good point! I have stole some text from the C# notes, and tweaked it for the Scala nodes.
Thanks

ErikSchierboom:
Great to have mentoring notes!

ricemery:
Thanks for your help!

PR #1110, opened 2019-06-21
java/matrix: add mentoring notes (Fixes #1102)

### LINKED ISSUE #1102: java/matrix: create mentor notes
Create some mentor notes for the java/matrix exercise.

Most importantly it should include one or more reasonable solutions (at least the general structure of a reasonable solution).

jmrunkle:
Add mentoring notes for java/matrix. Fixes #1102

### DIFF
--- a/tracks/java/exercises/matrix/mentoring.md
+++ b/tracks/java/exercises/matrix/mentoring.md
@@ -1,0 +1,54 @@
+### Reasonable solutions
+
+```java
+class Matrix {
+    private int[][] matrix;
+
+    Matrix(String matrixAsString) {
+        String[] rows = matrixAsString.split("\n");
+        matrix = new int[rows.length][];
+        for (int i = 0; i < rows.length; i++) {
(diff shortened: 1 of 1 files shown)


sshine:
There was a pretty heavy discussion in exercism/ruby#962 on bloated constructors.

Perhaps these thoughts are also fitting for Java.

jmrunkle:
It is a good point. The constructors can also tend to get bloated, but I think it depends if the constructor is doing only one thing. If so, then doing it directly in the constructor is neither better nor worse than passing it to another method. Once the constructor is trying to do more than one thing... I would strongly suggest it should be using methods.

In this particular class, the constructor is doing only one thing - creating the 2D int array / matrix. There are other breakdowns of the problem where you could end up cluttering the constructor with extra work (and I think there are other problems with some of those solutions).

If we ever added validation of the input matrix as part of the exercise, then that logic in the constructor may actually violate the SRP and be worth extracting two methods (and perhaps using an extra iteration through the data).

PR #1117, opened 2019-06-25
Add Ruby Two-Fer approval comments

iHiD:
Would appreciate any comments on this.

~I'm playing with the idea of using unicode character as our internal delimiter.~

I've now been overruled by @NobbZ and instead switched to this being ERB.

### DIFF
--- a/automated-comments/ruby/two-fer/kernel_format.md
+++ b/automated-comments/ruby/two-fer/kernel_format.md
@@ -1,0 +1,1 @@
+As well as the format method, a common way to create strings in Ruby is to use interpolation. For example, rather than `"format("One for %s, one for me.", <%= variable %>)` you might consider `"One for #{<%= variable %>}, one for me."`.
--- a/automated-comments/ruby/two-fer/kernel_format.md.erb
+++ b/automated-comments/ruby/two-fer/kernel_format.md.erb
@@ -1,1 +1,0 @@
-In Ruby, it's more common to use interpolation than the format method. So instead of `"format("One for %s, one for me.", <%= variable %>)`, consider `"One for #{<%= variable %>}, one for me."`.
--- a/automated-comments/ruby/two-fer/kernel_format.md.erb
+++ b/automated-comments/ruby/two-fer/kernel_format.md.erb
@@ -1,0 +1,1 @@
+In Ruby, it's more common to use interpolation than the format method. So instead of `"format("One for %s, one for me.", <%= variable %>)`, consider `"One for #{<%= variable %>}, one for me."`.
--- a/automated-comments/ruby/two-fer/string_building.md
+++ b/automated-comments/ruby/two-fer/string_building.md
@@ -1,0 +1,1 @@
+In Ruby, it is unusual to use string addition. People tend to use `sprintf`, `format, `%`, or string interpolation. So instead of `"One for " + <%= variable %> + ", one for me."`, consider `"One for #{<%= variable %>}, one for me."`.
--- a/automated-comments/ruby/two-fer/string_building.md.erb
+++ b/automated-comments/ruby/two-fer/string_building.md.erb
@@ -1,1 +1,0 @@
-In Ruby, it's more common to use interpolation than string addition. So instead of `"One for " + <%= variable %> + ", one for me."`, consider `"One for #{<%= variable %>}, one for me."`.
--- a/automated-comments/ruby/two-fer/string_building.md.erb
+++ b/automated-comments/ruby/two-fer/string_building.md.erb
@@ -1,0 +1,1 @@
+In Ruby, it's more common to use interpolation than string addition. So instead of `"One for " + <%= variable %> + ", one for me."`, consider `"One for #{<%= variable %>}, one for me."`.
--- a/automated-comments/ruby/two-fer/string_format.md
+++ b/automated-comments/ruby/two-fer/string_format.md
@@ -1,0 +1,1 @@
+As well as the `String#%` method, a common way to create strings in Ruby is to use interpolation. For example, rather than `"format("One for %s, one for me.", <%= variable %>)` you might consider `"One for #{<%= variable %>}, one for me."`.
--- a/automated-comments/ruby/two-fer/string_format.md.erb
+++ b/automated-comments/ruby/two-fer/string_format.md.erb
@@ -1,1 +1,0 @@
-In Ruby, it's more common to use interpolation than the `String#%` method. So instead of `"One for %s, one for me." % <%= variable %>`, consider `"One for #{<%= variable %>}, one for me."`.
--- a/automated-comments/ruby/two-fer/string_format.md.erb
+++ b/automated-comments/ruby/two-fer/string_format.md.erb
@@ -1,0 +1,1 @@
+In Ruby, it's more common to use interpolation than the `String#%` method. So instead of `"One for %s, one for me." % <%= variable %>`, consider `"One for #{<%= variable %>}, one for me."`.


SleeplessByte:
FWIW, I don't think the comment in https://github.com/exercism/website-copy/pull/1117/files#diff-af721afcdadb1d218ddcf55438f846d4 is factually _true_. Both the mentor notes and almost every single code-base I've worked with preferred `format('...', named: 'value', other: 'value')` over interpolation.

iHiD:
@SleeplessByte I've honestly never seen `sprintf` used in the wild in Ruby. And I can't really remember seeing `format` or `%` often either. Whereas interpolation is everywhere. Even 95% of the solutions approvable solutions that people submit use interpolation, which makes me think that's what they're finding when they Google.

If you can show me some data pointing the other way, I'm happy to be proven wrong.

SleeplessByte:
You can also barely find people using `.codepoints` and will find them use `.bytes` all the time. Just because it's plentiful, doesn't mean it's idiomatic or _right_, in this case.

1. The fact that rubocop [has this as a setting](https://github.com/rubocop-hq/rubocop/blob/ca01b3e18816b5a326d18ef3904128acc2891c52/spec/rubocop/cop/style/format_string_spec.rb#L191), and its default is to prefer [`format` and `sprintf` over `%`](https://github.com/rubocop-hq/ruby-style-guide#sprintf)
1. [batsov](https://batsov.com/articles/2013/06/27/the-elements-of-style-in-ruby-number-2-favor-sprintf-format-over-string-number-percent/) also added [`String#format`](https://www.rubydoc.info/gems/powerpack/String#format-instance_method) to his [powerpack]((https://github.com/bbatsov/powerpack).
1. [This](https://idiosyncratic-ruby.com/49-what-the-format.html) guide on idiosyncratic Ruby.
1. [This styleguide](https://relishapp.com/womply/ruby-style-guide/docs/strings) by cucumber on their relish project which favours interpolation and `format`.
1. The postgresql adapter in rails [uses](https://github.com/rails/rails/blob/4dcb46182a4aaa57f44f3eb722c1db54fa0ff843/activerecord/lib/active_record/connection_adapters/postgresql/quoting.rb#L53) [format](https://github.com/rails/rails/blob/b2eb1d1c55a59fee1e6c4cba7030d8ceb524267c/activerecord/lib/active_record/connection_adapters/postgresql/oid/date.rb#L13) a few [times](https://github.com/rails/rails/blob/b2eb1d1c55a59fee1e6c4cba7030d8ceb524267c/activerecord/lib/active_record/connection_adapters/postgresql/oid/date_time.rb#L13), granted for actual formatting (adding leading zero's etc). The rest of rails uses `sprintf` [for](https://github.com/rails/rails/blob/98a57aa5f610bc66af31af409c72173cdeeb3c9e/activesupport/lib/active_support/duration/iso8601_serializer.rb#L28) [messages](https://github.com/rails/rails/blob/db4b77aca147ec3c7376d803fc6ccb14c9195352/actionpack/lib/action_dispatch/testing/assertions/routing.rb#L60) [here](https://github.com/rails/rails/blob/db4b77aca147ec3c7376d803fc6ccb14c9195352/actionpack/lib/action_dispatch/testing/assertions/routing.rb#L100) and [there](https://github.com/rails/rails/blob/db4b77aca147ec3c7376d803fc6ccb14c9195352/actionpack/lib/action_dispatch/testing/assertions/routing.rb#L102) but also [again](https://github.com/rails/rails/blob/aeba121a83965d242ed6d7fd46e9c166079a3230/activerecord/lib/active_record/connection_adapters/abstract/quoting.rb#L126) for [number](https://github.com/rails/rails/blob/98a57aa5f610bc66af31af409c72173cdeeb3c9e/actionview/lib/action_view/helpers/date_helper.rb#L1003) [formatting](https://github.com/rails/rails/blob/b2eb1d1c55a59fee1e6c4cba7030d8ceb524267c/tools/line_statistics#L30).
1. The five people in [this recent feature request](https://bugs.ruby-lang.org/issues/15927) definitely use it.
1. [This book](https://books.google.nl/books?id=jcUbTcr5XWwC&pg=PA48&lpg=PA48&source=bl&ots=fKGhw92AhA&sig=ACfU3U2fYS57JiGDbR-QfLmUXT2eNSNSkA&hl=en&sa=X&ved=2ahUKEwiks5y21IXjAhXN-6QKHTxBAr44ChDoATAFegQICRAB#v=onepage&q=ruby%20format%20vs%20interpolation&f=false) talks about it in detail
1. It is used in [these](https://github.com/exercism/ruby/blob/5f131e59358227e313f15ce2c2313a0526409380/exercises/clock/.meta/solutions/clock.rb#L24) [examples](https://github.com/exercism/ruby/blob/5f131e59358227e313f15ce2c2313a0526409380/exercises/phone-number/.meta/solutions/phone_number.rb#L14) in `exercism/ruby`.
1. It's very hard to search on GH for `%` syntax, but [here](https://github.com/parasquid/upcloudify/blob/master/lib/upcloudify.rb#L26) is an example
1. Finally, the `ruby/ruby` mri implementation codebase uses it in the following libs: [`pp`](https://github.com/ruby/ruby/blob/fc9f19abe853bcef2b348e2d6d7c555b677ab28d/lib/pp.rb#L365), [`rdoc/generator`](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/lib/rdoc/generator/markup.rb#L52) [`libexec/racc`](https://github.com/ruby/ruby/blob/b93508b32c8f11bd468a885745338eab259bd2cd/libexec/racc#L102)  ([again](https://github.com/ruby/ruby/blob/b93508b32c8f11bd468a885745338eab259bd2cd/lib/racc/state.rb#L145)) [`drb/unix`](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/lib/drb/unix.rb#L78) [`webrick`](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/lib/webrick/httpserver.rb#L156) [`e2mmap`](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/lib/e2mmap.rb#L150) [`irb`](https://github.com/ruby/ruby/blob/70e87d9660af24fb93cac1312b6e1990602c1396/lib/irb/workspace.rb#L153) [`set`](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/lib/set.rb#L636) [`sync`](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/lib/sync.rb#L53) [`time`](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/lib/time.rb#L651) [`ipaddr`](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/lib/ipaddr.rb#L226) [`fileutils`](https://github.com/ruby/ruby/blob/746812ee9645ff821a039136f9576b7c8f18b920/lib/fileutils.rb#L1042) [`help_command`](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/lib/rubygems/commands/help_command.rb#L344) [`uri`](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/lib/uri/rfc2396_parser.rb#L309), the following tools: [🔧](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/tool/ruby_vm/models/attribute.rb#L41) [🔧](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/tool/ruby_vm/models/trace_instructions.rb#L27) [🔧](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/tool/ruby_vm/models/bare_instructions.rb#L52) [🔧](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/tool/ruby_vm/models/c_expr.rb#L39) [🔧](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/tool/ruby_vm/models/operands_unifications.rb#L90) and these samples: [🏷](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/sample/list2.rb#L9) [🏷](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/sample/list3.rb#L11) [🏷](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/sample/list.rb#L64) [🏷](https://github.com/ruby/ruby/blob/d48783bb0236db505fe1205d1d9822309de53a36/sample/openssl/c_rehash.rb#L160). In these cases, sometimes it's actual formatting, and sometimes it's using unnamed parameters and sometimes it's using named paramters to build a string.

Obvi, I won't give you the examples on which I've worked, because I'm biased, but this should show that it ís actually used in the wild. There are plenty of reasons you _don't_ want string interpolation, because you don't want to destructure a hash you already have, because you want to format the variable, or because you _want to store the format string in a file_/_use templating without a lib_/_want translatable strings_. It's _far_ easier to move from `sprintf` / `%` / `format` to something like `I18n`, then from `#{}`.

kotp:
Rubocop setting for point 1 is one of the first things I change in preference of `String#%`.

The blog incompletely, and perhaps incorrectly, states the first reason to prefer `sprintf` over `String#%` with "% takes either a single element or an array of elements as its sole argument".  It takes a single argument.  I think that part is important.  The argument can be an array or hash as well as the author refers to as a "single element".

And I used to avoid using it, when it first came out... but it has grown on me.  The argument about the `a % b` being unclear without context can be said for other "operator" methods as well... So I don't think the argument against it is that strong.

My main reason for using `Kernel#sprintf` and it's alias `/format` or using `String#%` vs interpolation is the lack of formatting for interpolation.  The formatting codes make it clear in the format and friends what is intended to be used as arguments (or at least what the formatting is expected to be) when reading the code.

But another difficulty is when you use string interpolation and try to format... then we end up often seeing a mix of both being used, which makes it look and read even worse.

I agree, it may be common to use interpolation rather than substitution... but that doesn't make it right.  Between using `sprintf`, `format` or `%` I leave it to the student to form their own opinion, while exposing the ones they did not use (and mentioning the aliasing that is going on, as well as the full documentation being available for `String#%` in the `Kernel` documentation.

iHiD:
Thank you all. I've changed the wording slightly to try and be less opinionated. I'm going to merge this because I need something on master to develop against to get this feature live. 

However, I would love PRs to improve the language, including links to articles on Ruby strings etc. With reference to the debate on the "normality" of interpolation, I'm not precious on pushing people towards interpolation, but I would like to make people aware of its existence.

PR #1121, opened 2019-06-28
Automentor Ruby/TwoFer: New comment 

F3PiX:
For my considerations, see https://github.com/exercism/automated-mentoring-support/issues/59


iHiD:
Thanks @F3PiX.

A couple of thoughts. This "has no problem with" reminded me of Katrina's comment [here](https://github.com/exercism/website-copy/pull/1009#discussion_r296794986).

We've said we won't use the word "you".

Maybe below?

---

String concatenation is perfectly valid within Ruby, but a more idiomatic way to compose a string is with [string interpolation](http://ruby-for-beginners.rubymonstas.org/bonus/string_interpolation.html). 

For example, rather than 

```
One for " + %<name_variable>s + ", one for me
```

you could do:

```
One for #{%<name_variable>s}, one for me
```

kytrinyx:
I've never been under the impression that string concatenation is less idiomatic than string interpolation. String concatenation seems to be used with shorter expressions a lot, whereas string interpolation is used for more complex examples.

iHiD:
I've changed "idiomatic" to "common".

iHiD:
I'm going to close this PR as it's now out of date with master, which incorporated these changes. I know this is still something you want to discuss @F3PiX, so please feel free to open a new PR against what's on master for discussion if you want to :)

PR #1126, opened 2019-06-29
csharp/leap: add automated mentoring notes

ErikSchierboom:


automated-comments/csharp/leap/do_not_use_if_statement.md
```diff
@@ -0,0 +1 @@
+Try converting the `if`-statement(s) to a single expression using the `&&` and `||` operators.
```
iHiD:
```suggestion
Consider converting the `if`-statement(s) into a single expression using the `&&` and `||` operators.
```

"Consider" is an instruction to think about. "Try" is an instruction to do something. I prefer the former when we're comment with approval. If these are comment with disapprove then ignore this for now.

ErikSchierboom: This is an approval comment, so I agree with your suggestion and have applied it.

PR #1128, opened 2019-06-29
csharp/general: add automated mentoring notes

ErikSchierboom:


mikedamay:
4.In [automated-comments/csharp/general/convert_variable_to_const.md.]( https://github.com/exercism/website-copy/pull/1128/files#diff-26fe5bebb2a23b52f325c11e879eeab3)  “make it a class-only constant, which is the more idiomatic solution”.  I had never thought of a class level constant being more idiomatic than method level.  I would say that the decision depends on the role of the value in the code – a class level constant where it reflects business rules or some real-world value or method level where a number of fixed values are used in different combinations for some reason.

5.In [automated-comments/csharp/general/do_not_use_nested_if_statement.md](https://github.com/exercism/website-copy/pull/1128/files#diff-2bb0f0aceca14ce9fe55115bc25c869f) I have never understood the preference for early returns or the obsession with cyclomatic complexity although I concede I am in the minority.  I was slightly heartened to see that the Stack Exchange Software Engineering question  https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement was closed as opinion based so there is hope for my tribe.

6.In [automated-comments/csharp/general/return_immediately.md](https://github.com/exercism/website-copy/pull/1128/files#diff-c2760094aa83f9b8bc0a7f0f33e61e05) I have slightly milder reservations about “Instead of storing the return value into an intermediate value, you could just return it directly” than I do about the “nested if” but I don’t think I would make the point in a review.  There is even a school of thought that a single return statement is a good thing although I think that’s going out of fashion.  My own position is that it depends on the specific case – length of method, readability, for starters.

7.In [automated-comments/csharp/general/use_expression_bodied_member.md](https://github.com/exercism/website-copy/pull/1128/files#diff-161d6d3094803e98f77f982e7ce2563d) “method only has a single line of code”.  Could we say “method only has a single statement”?  And would this be an appropriate place to mention that the expression can span more than one line if this aids readability?

8.In [automated-comments/csharp/general/use_private_visibility.md](https://github.com/exercism/website-copy/pull/1128/files#diff-2a0779290463836777674ff043f64328) “field is only used within its class, its visibility can be set to `private`.”  Maybe “its visibility can and, almost always should, be set to `private`.”?

9.In [automated-comments/csharp/general/use_string_interpolation_not_string_format.md](https://github.com/exercism/website-copy/pull/1128/files#diff-332137f6e2774b72742059be63201625) “Note that string interpolation is just a compiler trick”.  Maybe “Note that string interpolation is just a compiler trick, also known as syntactic sugar”?

automated-comments/csharp/general/convert_variable_to_const.md
```diff
@@ -0,0 +1 @@
+You can convert the `%{name}` variable to a `const` as the value is intended never to change. You can either declare the `const` a method-only constant, by defining it within the method (similar to how the variable is defined), or make it a class-only constant, which is the more idiomatic solution.
```
ErikSchierboom: @mikedamay said:

“make it a class-only constant, which is the more idiomatic solution”. I had never thought of a class level constant being more idiomatic than method level. I would say that the decision depends on the role of the value in the code – a class level constant where it reflects business rules or some real-world value or method level where a number of fixed values are used in different combinations for some reason."

automated-comments/csharp/general/do_not_use_nested_if_statement.md
```diff
@@ -0,0 +1,35 @@
+You are currently using nested if-statements. In general, it is usually best to avoid deep nesting, as it makes for harder reading. Consider the following two pieces of code:
```
ErikSchierboom: @mikedamay said:

"I have never understood the preference for early returns or the obsession with cyclomatic complexity although I concede I am in the minority. I was slightly heartened to see that the Stack Exchange Software Engineering question https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement was closed as opinion based so there is hope for my tribe."

automated-comments/csharp/general/return_immediately.md
```diff
@@ -0,0 +1 @@
+Instead of storing the return value into an intermediate value, you could just return it directly. This usually makes code a bit easier to understand, as seeing that a method returns will allow you to skip reading the rest of the code, which you would still have to do when a variable is used, as that variable could be modified in another line of code.
```
ErikSchierboom: @mikedamay said:

"I have slightly milder reservations about “Instead of storing the return value into an intermediate value, you could just return it directly” than I do about the “nested if” but I don’t think I would make the point in a review. There is even a school of thought that a single return statement is a good thing although I think that’s going out of fashion. My own position is that it depends on the specific case – length of method, readability, for starters."

automated-comments/csharp/general/use_expression_bodied_member.md
```diff
@@ -0,0 +1 @@
+As the `%{name}` method only has a single line of code, you could write the method as an [expression-bodied method](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/expression-bodied-members#methods).
```
ErikSchierboom: @mikedamay said:

“method only has a single line of code”. Could we say “method only has a single statement”? And would this be an appropriate place to mention that the expression can span more than one line if this aids readability?

automated-comments/csharp/general/use_private_visibility.md
```diff
@@ -0,0 +1 @@
+As the `%{name}` field is only used within its class, its visibility can be set to `private`.
```
ErikSchierboom: @mikedamay said:

“field is only used within its class, its visibility can be set to private.” Maybe “its visibility can and, almost always should, be set to private.”?

automated-comments/csharp/general/use_string_interpolation_not_string_format.md
```diff
@@ -0,0 +1 @@
+You are currently using `string.Format` to dynamically build a string. This is perfectly fine, but the more idiomatic C# solution to create such a string is to use [string interpolation](https://csharp.net-tutorials.com/operators/the-string-interpolation-operator/). Note that string interpolation is just a compiler trick, as under the hood any string interpolation expression is converted to a `string.Format` call. The benefit of using string interpolation is thus purely visual.
```
ErikSchierboom: @mikedamay said:

“Note that string interpolation is just a compiler trick”. Maybe “Note that string interpolation is just a compiler trick, also known as syntactic sugar”?

ErikSchierboom:
@mikedamay I've added your comments inline to the appropriate files, which makes it slightly easier to have a discussion :) Thanks for the feedback, I'll post my response soon.

automated-comments/csharp/general/use_string_interpolation_not_string_format.md
ErikSchierboom: I've added the syntactic sugar hint, and did some rephrasing about the compiler bit.

automated-comments/csharp/general/use_private_visibility.md
ErikSchierboom: Good point! I've updated.

automated-comments/csharp/general/use_expression_bodied_member.md
ErikSchierboom: Good point!

> And would this be an appropriate place to mention that the expression can span more than one line if this aids readability?

Sure! I've added an example.

Absol

automated-comments/csharp/general/convert_variable_to_const.md
ErikSchierboom: @mikedamay These additions are great. Do you perhaps have a suggestion for a new comment?

automated-comments/csharp/general/do_not_use_nested_if_statement.md
ErikSchierboom: Hmmm, let's ask some other people's opinions then. @robkeim @SleeplessByte what do you think about this comment and the guideline itself?

automated-comments/csharp/general/return_immediately.md
ErikSchierboom: Just to be sure, this comment will only be applied when the code looks like this:

```csharp
public int Test()
{
    ....
    var ret = CalculateValue();
    return ret;
}
```

So there are two conditions that have to be met:

- The value that is returned is assigned to in the previous line.
- The return statement is the last statement in the method.

Does it make more sense now to have this comment? I think I may have accidentally copied the wrong comment here, hence the confusion. To reduce the confusion, I've renamed the comment to "do_not_assign_and_return". Does that make things more clear?

Regarding the direct return that I think you are referring to, I _am_ in favor of doing that, but I agree that that can be a very personal thing.

ErikSchierboom:
@mikedamay Thanks for the review! I've updated several things and left some questions for you.

automated-comments/csharp/general/do_not_use_nested_if_statement.md
mikedamay: Sounds good - as I said in my original reply on Slack - there are no show stoppers.  I would certainly be comfortable with your original wording.

automated-comments/csharp/general/return_immediately.md
mikedamay: I totally agree with your position.

automated-comments/csharp/general/use_expression_bodied_member.md
mikedamay: Great

automated-comments/csharp/general/use_string_interpolation_not_string_format.md
mikedamay: Sounds good

automated-comments/csharp/general/convert_variable_to_const.md
mikedamay: You can convert the `%{name}` variable to a `const` as the value is intended never to change.  You will most often see `const`s defined at the class level as they are frequently used in multiple methods.  You have the option of defining them at the method level when their use is restricted to a single method.  This minimises the scope within which they are visible.

FYI but not part of the comment
https://stackoverflow.com/questions/6373072/defining-local-variable-const-vs-class-const

automated-comments/csharp/general/do_not_use_nested_if_statement.md
SleeplessByte: Nested ifs in generals aren't bad, but I do care about cognitive complexity (and therefor also about cyclomatic complexity). Nested if statements often _hide_ other smells or issues. I think it's fine to point this out.

I would suggest:

```suggestion
You are currently using nested if-statements. When possible, try to avoid deep nesting, as it can lead to code that is harder to follow. Consider the following two pieces of code:
```

automated-comments/csharp/general/return_immediately.md
ErikSchierboom: Good to hear!

automated-comments/csharp/general/do_not_use_nested_if_statement.md
ErikSchierboom: Great suggestion @SleeplessByte !

automated-comments/csharp/general/convert_variable_to_const.md
ErikSchierboom: I like it! I've updated the comment.

automated-comments/csharp/general/convert_field_to_const.md
```diff
@@ -0,0 +1 @@
+Consider converting the `%{name}` field to a `const`, as the value is intended never to change. Using constants is not only more common, there are also some subtle differences between a `const` and a `field`, which are explained in [this StackOverflow post](https://stackoverflow.com/questions/755685/static-readonly-vs-const#755693).
```
SleeplessByte: Will the analyser still _block_ if this change is not made? Or is this always a "tip"? If it still blocks:

```suggestion
Convert the `%{name}` field to a `const`, as the value is intended never to change. Using constants is not only more common, there are also some subtle differences between a `const` and a `field`, which are explained in [this StackOverflow post](https://stackoverflow.com/questions/755685/static-readonly-vs-const#755693).
```

automated-comments/csharp/general/convert_variable_to_const.md
```diff
@@ -0,0 +1 @@
+Consider converting the `%{name}` variable to a `const`, as the value is intended never to change. Usually, `const` values are defined at the class level, as they are frequently used in multiple methods. In this case, the `const` value can be defined within the method itself, which means that other methods cannot use its value.
```
SleeplessByte: Will the analyser still block if this change is not made? Or is this always a "tip"? If it still blocks:


```suggestion
Convert the `%{name}` variable to a `const`, as the value is intended never to change. Usually, `const` values are defined at the class level, as they are frequently used in multiple methods. In this case, the `const` value can be defined within the method itself, which means that other methods cannot use its value.
```

automated-comments/csharp/general/do_not_assign_and_return.md
```diff
@@ -0,0 +1 @@
+Consider returning the value directly instead of storing it into an intermediate value and then returning it.
```
SleeplessByte: Will the analyser still block if this change is not made? Or is this always a "tip"? If it still blocks:

```suggestion
Return the value directly, instead of storing it into an intermediate value and then returning it.
```

automated-comments/csharp/general/has_main_method.md
```diff
@@ -0,0 +1,7 @@
+Try removing the `Main` method from code, as the tests don't require them and they make the code slightly harder to read.
+
+However, a better approach is to debug the code while running one or more unit tests. This has the added advantage that one can focus on debugging a specific test case. Here are some links that explain how to debug C# code while running unit tests in various IDE's:
```
SleeplessByte: Is this supposed to be here?

automated-comments/csharp/general/use_null_coalescing_operator_not_null_check.md
```diff
@@ -0,0 +1 @@
+Rather than explicitly checking for a value to equal `null` and then doing something with it, consider using the [null-coalescing operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator) to simplify your code.
```
SleeplessByte: I would invert this per https://github.com/exercism/automated-mentoring-support/pull/61.

```suggestion
Use the [null-coalescing operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator) to simplify your code, rather than explicitly checking for a value to equal `null` and then doing something with it.
```

automated-comments/csharp/general/use_string_interpolation_not_string_concatenation.md
```diff
@@ -0,0 +1 @@
+Rather than using string concatenation, consider using [string interpolation](https://csharp.net-tutorials.com/operators/the-string-interpolation-operator/) to dynamically build a string.
```
SleeplessByte: I would invert this per exercism/automated-mentoring-support#61.

```suggestion
Use [string interpolation](https://csharp.net-tutorials.com/operators/the-string-interpolation-operator/) to dynamically build a string, rather than using string concatenation.
```

Perhaps you want to point out _why_?

automated-comments/csharp/general/use_string_interpolation_not_string_format.md
```diff
@@ -0,0 +1,3 @@
+Rather than using `string.Format` to dynamically build a string, the more common C# approach is to use [string interpolation](https://csharp.net-tutorials.com/operators/the-string-interpolation-operator/).
```
SleeplessByte: Perhaps also invert this?

```suggestion
A more common approach in C# is to use [string interpolation](https://csharp.net-tutorials.com/operators/the-string-interpolation-operator/), rather than using `string.Format` to dynamically build a string
```

automated-comments/csharp/general/use_string_interpolation_not_string_format.md
```diff
@@ -0,0 +1,3 @@
...
+
+Note that string interpolation is just a compiler trick, also known as syntactic sugar. This means that when a string interpolation expression is compiled, the compiler will just convert it to a `string.Format` call. The benefit of using string interpolation is thus purely visual.
```
SleeplessByte: TIL 🎈

automated-comments/csharp/general/do_not_use_nested_if_statement.md
```diff
@@ -0,0 +1,35 @@
...
+    }
+}
+```
+
+Compare that to the following code:
+
+```csharp
+void PrintName(Person p)
+{
+  if (p == null)
```
robkeim: Is there a reason we're not suggesting:
```
if (p == null || p.Name == null)
```

automated-comments/csharp/general/has_compile_errors.md
```diff
@@ -0,0 +1 @@
+The code has one or more compile errors, which prevents the test suite from running. A good first step in finding out how to fix the error(s), is to examine the error messages displayed when trying to compile the code. These error messages contain an error number and description, which help pinpoint what the exact problem is.
```
robkeim:
```suggestion
The code has one or more compile errors, which prevents the test suite from running. A good first step in finding out how to fix the error(s) is to examine the error messages displayed when trying to compile the code. These error messages contain an error number and description, which helps pinpoint what the exact problem is.
```

automated-comments/csharp/general/do_not_use_nested_if_statement.md
robkeim: I think it's hard to find a strict black and white rule that applies in this case, but I also think it should be pointed out as a potential area of improvement.

automated-comments/csharp/general/do_not_use_nested_if_statement.md
ErikSchierboom: Nope. I'll apply the suggestion.

automated-comments/csharp/general/convert_field_to_const.md
ErikSchierboom: The analyser will not block (this is used in the approve_with_comment) path.

automated-comments/csharp/general/do_not_assign_and_return.md
ErikSchierboom: The analyser will not block (this is used in the approve_with_comment) path.

automated-comments/csharp/general/use_string_interpolation_not_string_format.md
ErikSchierboom: Nice! :)

automated-comments/csharp/general/has_main_method.md
ErikSchierboom: Whoops, I missed a bit. Will add the missing text.

ErikSchierboom:
@robkeim @SleeplessByte I've updated the comments!

automated-comments/csharp/general/use_string_interpolation_not_string_format.md
```diff
@@ -0,0 +1,3 @@
+A more common approach in C# is to use [string interpolation](https://csharp.net-tutorials.com/operators/the-string-interpolation-operator/), rather than using `string.Format` to dynamically build a string.
+
+Note that string interpolation is just a compiler trick, also known as syntactic sugar. This means that when a string interpolation expression is compiled, the compiler will actually convert it to a `string.Format` call. The benefit of using string interpolation is thus purely visual.
```
iHiD: I really like this additional note.

ErikSchierboom:
Thanks @robkeim!

PR #1132, opened 2019-07-02
Fix incorrect wording under "reasonable solutions"

jgilray:


TheDarkula:
Great catch! Thanks :)

PR #1138, opened 2019-07-03
csharp/general: remove confusing example

ErikSchierboom:
Temporarily removing the example, as that is only really useful when the code itself can benefit from it.


iHiD:
This might be a good example of where not using "general" but having it specific is good too.

ErikSchierboom:
> This might be a good example of where not using "general" but having it specific is good too.

Well, I think this should still be general, only that we should have two separate cases:

1. For very simple code (this case).
1. For more complex code that uses several chained method calls.

PR #1144, opened 2019-07-03
go/twofer: escape % properly where it's not a placeholder

tehsphinx:
Sample case to fix: https://exercism.io/mentor/analyses/6860


ErikSchierboom:
Boy, the CI is slow as hell today!

PR #1145, opened 2019-07-03
csharp/general: generalize console write comment

ErikSchierboom:
Just saw a solution that uses `Console.Error.WriteLine`, which is also wrong, but did not match `Console.WriteLine` or `Console.Write`. Hence, this generalization.


ErikSchierboom:
Thanks for the review!

PR #1156, opened 2019-07-08
javascript/gigasecond: Fix export_inline template

SleeplessByte:


automated-comments/javascript/gigasecond/export_inline.md
```diff
@@ -1,7 +1,5 @@
 Functions, classes and constants can be `export`ed inline.
 
 ```javascript
-const %{const.name} = (...)
-
 export %{method-signature}
```
ErikSchierboom:
```suggestion
export %{method-signature}
```

The comments talk about exporting functions, classes and constants, but the example mentions a method-signature. Should that maybe be someting like `%{exported-identifier}` or something like that?

ccare:
Just to clarify: my :+1: refers to fixing the missing template, rather than the semantics of the messaging.

SleeplessByte:
@ErikSchierboom yeah the naming is sub-optimal, but makes sense in the `gigasecond` context. There's only a function in this case. I can't change the parameter name here because it's already deployed :P

SleeplessByte:
FWIW, it's called like this:

```json
{
  "params": {
    "method.signature": "let gigasecond = (date) => ..."
  },
  "comment": "javascript.gigasecond.export_inline"
}
```

So it becomes:

```
Functions, classes and constants can be exported inline.

export let gigasecond = (date) => ...
```

ErikSchierboom:
Alright, thanks for explaining!

PR #1171, opened 2019-07-17
Add comments when solving two-fer with string interpolation

kntsoriano:
This adds comments when solving two-fer on Ruby with string interpolation.

Related issue: https://github.com/exercism/ruby-analyzer/issues/9

### DIFF
--- a/automated-comments/ruby/two-fer/string_interpolation.md
+++ b/automated-comments/ruby/two-fer/string_interpolation.md
@@ -1,0 +1,5 @@
+As well as string interpolation, another common way to create strings in Ruby is to use [String#format](https://www.rubyguides.com/2012/01/ruby-string-formatting/). For example:
+
+```ruby
+"One for %%s, one for you" %% %{name_variable}"
+```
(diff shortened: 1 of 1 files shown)


automated-comments/ruby/two-fer/string_interpolation.md
```diff
@@ -0,0 +1,5 @@
+As well as string interpolation, another common way to create strings in Ruby is to use [String#format](https://www.rubyguides.com/2012/01/ruby-string-formatting/). For example:
```
kotp:
```suggestion
As well as string interpolation, another common way to create strings in Ruby is to use [String#%](https://www.rubyguides.com/2012/01/ruby-string-formatting/) (perhaps read as "String format").
For example:
```
There is no `String#format`, but there is `String#%`.

kytrinyx: I think it's `Kernel#format` https://ruby-doc.org/core-2.6.3/Kernel.html#method-i-format

kotp: Because of the content of the given link, I believe the focus was `String` class, rather than the `Kernel` class, though, yes, of course, `Kernel#format` exists as well.  I think the other wording to indicate that it should instead be `Kernel` is because of the "another common way" wording.  Perhaps `Kernel` is used more often for `String` formatting, which may be non-intuitive.

kntsoriano: Thanks for this. I'm adding the suggestion above as part of this PR.

automated-comments/ruby/two-fer/string_interpolation.md
```diff
@@ -0,0 +1,5 @@
...
+
+```ruby
+"One for %%s, one for you" %% %{name_variable}"
```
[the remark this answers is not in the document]
kntsoriano: Agreed.

PR #1174, opened 2019-07-18
Corrects the example Hamming Java solutions.

SteveNewson:
The current solutions for the Hamming Java problem does not actually
pass the test. When either value is empty, the exception message
should be different. This change corrects that.


SteveNewson:
This is a duplicate of: https://github.com/exercism/website-copy/pull/1003

PR #1188, opened 2019-07-25
rust/bob: Make code simpler and more concise.

Aloso:
It's encouraged to shadow a variable name, if the name is appropriate and the old value isn't needed anymore.

`match` shouldn't be used for single booleans, `if`/`else` is equally readable, but shorter and requires fewer indents and braces.

It's recommended to use functions paths instead of closures where possible.

Rust can infer types, so it's OK to omit the type, if it's obvious from the variable name. Variables starting with `is_` or `contains_` are usually booleans.

### DIFF
--- a/tracks/rust/exercises/bob/mentoring.md
+++ b/tracks/rust/exercises/bob/mentoring.md
@@ -64,7 +64,7 @@ If they don't use a constant:
 We can improve this by setting our output strings to be [constants](https://doc.rust-lang.org/std/keyword.const.html)
 ```
 
-If they don't create a trimmed variable or use something other than `.trim()` or `.trim_right()`:
+If they don't create a trimmed variable or use something other than `.trim()` or `.trim_end()`:
 ```
 Have a look at the `.trim()` method.
 ```
--- a/tracks/rust/exercises/bob/mentoring.md
+++ b/tracks/rust/exercises/bob/mentoring.md
@@ -55,7 +55,7 @@ pub fn reply(message: &str) -> &str {
 }
 ```
 
-Note that `message.chars().any(..)` is equivalent to `message.contains(..)`, and `message.chars().all(..)` is equivalent to `message.matches(..)`.
+Note that `message.chars().any(..)` is equivalent to `message.contains(..)`.
 
 ### Common Suggestions
 
--- a/tracks/rust/exercises/bob/mentoring.md
+++ b/tracks/rust/exercises/bob/mentoring.md
@@ -31,30 +31,27 @@ const CALM: &str = "Calm down, I know what I'm doing!";
 const FINE: &str = "Fine. Be that way!";
 const WHATEVER: &str = "Whatever.";
 
-#[allow(clippy::match_bool)]
 pub fn reply(message: &str) -> &str {
-    let trimmed: &str = message.trim();
+    let message: &str = message.trim();
 
-    match trimmed.is_empty() {
-        true => FINE,
-        false => {
-            let is_question: bool = trimmed.ends_with("?");
+    if message.is_empty() {
+        FINE
+    } else {
+        let is_question = message.ends_with('?');
 
-            let contains_alphabetic_characters: bool = trimmed.chars().any(|character| character.is_alphabetic());
-            let is_uppercase: bool = trimmed == trimmed.to_uppercase();
-            let is_yelled: bool = contains_alphabetic_characters && is_uppercase;
+        let contains_alphabetic = message.chars().any(char::is_alphabetic);
+        let is_uppercase = !message.chars().any(char::is_lowercase);
+        let is_yelled = contains_alphabetic && is_uppercase;
 
-            match (is_question, is_yelled) {
-                (true, false) => SURE,
-                (false, true) => WHOA,
-                (true, true) => CALM,
-                (false, false) => WHATEVER,
-            }
+        match (is_question, is_yelled) {
+            (true, false) => SURE,
+            (false, true) => WHOA,
+            (true, true) => CALM,
+            (false, false) => WHATEVER,
         }
     }
 }
 ```
-Variations of `is_yelled` can also include using a `chars()` loop to ensure no extra `String` is allocated.
 
 ### Common Suggestions
 
--- a/tracks/rust/exercises/bob/mentoring.md
+++ b/tracks/rust/exercises/bob/mentoring.md
@@ -32,7 +32,7 @@ const FINE: &str = "Fine. Be that way!";
 const WHATEVER: &str = "Whatever.";
 
 pub fn reply(message: &str) -> &str {
-    let message: &str = message.trim();
+    let message = message.trim();
 
     if message.is_empty() {
         FINE
--- a/tracks/rust/exercises/bob/mentoring.md
+++ b/tracks/rust/exercises/bob/mentoring.md
@@ -131,7 +131,7 @@ If they ask about `&str` having a static lifetime:
 ```
 Lifetimes prevent dangling pointers; they are needed when a value points to some data, so the value doesn't outlive its data.
 
-String literals are [slices](https://doc.rust-lang.org/std/primitive.slice.html) that point to UTF-8 string data. This data is never freed, so string slices are alive for the entire duration of the application, so they have the `'static` lifetime.
+String literals are [slices](https://doc.rust-lang.org/std/primitive.slice.html) that point to a section in the executable, where all strings are stored. Because this data is never freed, string slices are alive for the entire duration of the application, which is denoted with the special `'static` lifetime.
 ```
 
 If they use `_` instead of `false` in their match statement:
--- a/tracks/rust/exercises/bob/mentoring.md
+++ b/tracks/rust/exercises/bob/mentoring.md
@@ -40,6 +40,8 @@ pub fn reply(message: &str) -> &str {
         let is_question = message.ends_with('?');
 
         let contains_alphabetic = message.chars().any(char::is_alphabetic);
+        // A message is uppercase if it doesn't contain lowercase.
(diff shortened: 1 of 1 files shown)


tracks/rust/exercises/bob/mentoring.md
```diff
@@ -31,30 +31,27 @@ const CALM: &str = "Calm down, I know what I'm doing!";
...
-            let is_uppercase: bool = trimmed == trimmed.to_uppercase();
-            let is_yelled: bool = contains_alphabetic_characters && is_uppercase;
-
-            match (is_question, is_yelled) {
-                (true, false) => SURE,
-                (false, true) => WHOA,
-                (true, true) => CALM,
-                (false, false) => WHATEVER,
-            }
+    let message: &str = message.trim();
```
coriolinus: Is the type annotation here necessary?

Aloso: No, it isn't I can remove it as well.

tracks/rust/exercises/bob/mentoring.md
```diff
@@ -31,30 +31,31 @@ const CALM: &str = "Calm down, I know what I'm doing!";
...
-                (true, false) => SURE,
-                (false, true) => WHOA,
-                (true, true) => CALM,
-                (false, false) => WHATEVER,
-            }
+    let message = message.trim();
+
+    if message.is_empty() {
+        FINE
+    } else {
```
L0uisc: Why the `else` block here? I think I prefer the following:
```rust
if message.is_empty() {
    return FINE
}
let is_question = message.ends_with('?');
let contains_alphabetic = message.chars().any(char::is_alphabetic);
// A message is uppercase if it doesn't contain lowercase.
// `.all(char::is_uppercase)` doesn't work because of whitespace/punctuation
let is_uppercase = !message.chars().any(char::is_lowercase);
let is_yelled = contains_alphabetic && is_uppercase;
match (is_question, is_yelled) {
    (true, false) => SURE,
    (false, true) => WHOA,
    (true, true) => CALM,
    (false, false) => WHATEVER,
}
```

tracks/rust/exercises/bob/mentoring.md
```diff
@@ -31,30 +31,31 @@ const CALM: &str = "Calm down, I know what I'm doing!";
...
+        match (is_question, is_yelled) {
+            (true, false) => SURE,
+            (false, true) => WHOA,
+            (true, true) => CALM,
+            (false, false) => WHATEVER,
         }
     }
 }
 ```
-Variations of `is_yelled` can also include using a `chars()` loop to ensure no extra `String` is allocated.
```
SleeplessByte: Is there something wrong with this advice? When something like this is removed, please give sufficient reasoning _why_.

tracks/rust/exercises/bob/mentoring.md
```diff
@@ -106,11 +93,6 @@ Then note the following:
 See about finding a way to check if it's alphabetic without using regex.
 ```
 
-If they use separate functions:
```
SleeplessByte: I see you say 

> It's recommended to use functions paths instead of closures where possible.

Is this a _language recommendation_? Or a personal one? From what I've seen from my own fiddling with rust is that in production-projects closures are used heavily and I can't seem to match this recommendation. Some science here will probably really help with accepting this change, given that it doesn't clash with `track-anatomy`.

tracks/rust/exercises/bob/mentoring.md
```diff
@@ -126,13 +108,11 @@ If they put their constants inside the function:
 Note that it's common convention to put constants at the beginning of a file like I've done above. If it's a public constant, which is typical, we put it following our `use` lines, along with `struct`s and `enum`s. If it's a private constant, we put it at the beginning of the scope of the function. But, either way doesn't affect the performance or functionality!
 ```
 
-If they ask about `&str` having a static lifetime in comparison to using constants:
+If they ask about `&str` having a static lifetime:
 ```
-To quote from [Rust By Example](https://doc.rust-lang.org/rust-by-example/scope/lifetime/static_lifetime.html): "When `static_string` goes out of scope, the reference can no longer be used, but the data remains in the binary."
```
SleeplessByte: This seems like great advice (with the link I mean). Can you explain what's wrong with it? Or is this part of the rest of the "change" for lifetimes?

Can we keep the quote + link in, but use your description?

tracks/rust/exercises/bob/mentoring.md
Aloso: This advice is no longer relevant because of the changes I made. My sample solution uses

```rust
let is_uppercase = !message.chars().any(char::is_lowercase);
```

tracks/rust/exercises/bob/mentoring.md
Aloso: `foo` is simply shorter than `|x| x.foo()`. Clippy suggests to remove redundant closures as well.

tracks/rust/exercises/bob/mentoring.md
Aloso: I'l add the link again.

I removed this section, because

1. The question in the heading is too vague.
2. The answer is in part incorrect.
3. The answer goes off-topic. I feel like it doesn't really answer the question(s).

tracks/rust/exercises/bob/mentoring.md
Aloso: I don't have a strong opinion here. However, I like the solution with `else` because it demonstrates that everything is an expression.

Note that the sample solution isn't supposed to be perfect/ideal, it's only one of many good solutions. Which is “best” is a matter of taste.

tracks/rust/exercises/bob/mentoring.md
Aloso: Oh, I see this is about something else.

The mentoring notes originally said that you should _not_ use helper functions, but I disagree. This is a subject on which there are many different opinions. We shouldn't force students to use a certain pattern, unless there's a _clear_ advantage.

> It's recommended to use functions paths instead of closures where possible.

I should have phrased this better :)

This is about closures like `|c: char| c.is_alphabetic()`, which can be replaced with `char::is_alphabetic`.

The [clippy source](https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/eta_reduction.rs#L18) says about this: “Needlessly creating a closure adds code for no benefit and gives the optimizer more work.”

tracks/rust/exercises/bob/mentoring.md
```diff
@@ -31,30 +31,31 @@ const CALM: &str = "Calm down, I know what I'm doing!";
...
+
+    if message.is_empty() {
+        FINE
+    } else {
+        let is_question = message.ends_with('?');
+
+        let contains_alphabetic = message.chars().any(char::is_alphabetic);
+        // A message is uppercase if it doesn't contain lowercase.
+        // `.all(char::is_uppercase)` doesn't work because of whitespace/punctuation
+        let is_uppercase = !message.chars().any(char::is_lowercase);
```
L0uisc: I just saw this in a student's work:
```rust
let is_uppercase = message.chars().any(char::is_uppercase)
    && !message.chars().any(char::is_lowercase);
```
Should be a bit faster even, since it will short-circuit in more cases than just testing for at least one alphabetic char.

Aloso: You're right. I will change it.

I believe the most efficient (but not the most beautiful) solution is this:

```rust
let message = message.trim_start_matches(|c: char| !c.is_alphabetic());
!message.is_empty() && !message.chars().any(char::is_lowercase)
```

tracks/rust/exercises/bob/mentoring.md
Aloso: @SleeplessByte I added the link again.

Aloso:
@SleeplessByte are there any issues remaining that I should address?

SleeplessByte:
@Aloso I will get back to you and ping you if so 🥇

tracks/rust/exercises/bob/mentoring.md
```diff
@@ -31,30 +31,30 @@ const CALM: &str = "Calm down, I know what I'm doing!";
 const FINE: &str = "Fine. Be that way!";
 const WHATEVER: &str = "Whatever.";
 
-#[allow(clippy::match_bool)]
 pub fn reply(message: &str) -> &str {
```
L0uisc: While this has not been merged yet: a student added a comment that he changed the return type to `&'static str` since there is no sense in tying the output lifetime to the input. What are your thoughts? I don't really know, but I think it is both true and unnecessary. It is true that the output isn't tied to the input, but it is less clutter and it doesn't matter that there is no `'static` currently.

NobbZ: And this comment makes me wonder, why a `&str` is returned at all, and not a `String` for the result.

Aloso: @NobbZ because the string doesn't need to be allocated on the heap; returning a `&str` is cheaper.

@L0uisc I agree that `&'static str` would make more sense. I'll ask it on slack, so we can discuss it there.

efx:
Anything holding this back from being merged @SleeplessByte? 
I think these are helpful changes as it makes the canonical solution cleaner.

SleeplessByte:
Unfortunately this has not been resolved internally yet. I'll see if a subset of the changes _can_ be merged!

coriolinus:
@SleeplessByte Given that we're more or less punting on track anatomy in favor of v3, are you willing to rescind your change request to allow this to move forward now?

angelikatyborska:
@SleeplessByte ping :)

SleeplessByte (approved):

Aloso:
Wow, a 5 year old PR merged! 🎉

SleeplessByte:
@Aloso apologies and celebrations ❤️

PR #1195, opened 2019-07-29
Clarify mutation is allowed in high-scores

adamzev:
Adding a test to check against mutating the list has been suggested and rejected because this is an introductory core exercise and immutability has not yet been introduced. However, all of the reasonable examples provided do not mutate the list so it is not clear to new mentors whether preserving the original list is required when deciding if they should accept a solution. I added clarification to the notes that mutating is acceptable but should be commented on.

See [this thread](https://github.com/exercism/python/pull/1868#issuecomment-515838536) for further discussion that led to this pull request.

### DIFF
--- a/tracks/python/exercises/high-scores/mentoring.md
+++ b/tracks/python/exercises/high-scores/mentoring.md
@@ -48,6 +48,8 @@ Another variation of this solution includes the start at -1 for slicing the resu
         return sorted(scores)[-1:-4:-1]
 ```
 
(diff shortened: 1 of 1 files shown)


adamzev:
> I feel like this note is a bit out of place without an example of the "reasonable and acceptable" solution that uses mutation.

I was trying to keep it succinct but I see your point. I added examples of the most common mutations I've seen in two of the functions.

yawpitch:
@adamzev thanks for your work on this.

PR #1197, opened 2019-07-29
rust/reverse-string: Remove advice about turbofish operator

Tohnmeister:
Discussion on the #track-rust Slack channel shows that usage of the turbofish operator, in cases where it's not needed by the compiler, is a controversial subject. Hence, it makes sense to not put any advice about it in the mentoring notes.

### DIFF
--- a/tracks/rust/exercises/reverse-string/mentoring.md
+++ b/tracks/rust/exercises/reverse-string/mentoring.md
@@ -12,7 +12,6 @@ A reasonable solution should:
 - Use [.chars()](https://doc.rust-lang.org/std/primitive.str.html#method.chars),
 [.rev()](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.rev), and
 [.collect()](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect)
-- Be explicit with `.collect()` by using the turbofish operator `::<>`
 - Not use other iterators or iterator methods like `.into_iter()` or
 `.flat_map()`
 - Not use variables or loops
@@ -44,12 +43,6 @@ pub fn reverse(input: &str) -> String {
 ```
 
 ### Common Suggestions
-
-If they don't use the turbofish operator:
-```
-It's best to be explicit with generics such as `.collect()` by using the turbofish operator `::<>` to specify what type we want to build. With
-more code, it makes it easier to simply look at the collect instead of having to find the function's signature.
-```
 
 If they use other iterators:
 ```


TheDarkula:
Yes, we are an opionated bunch :)

The thing to remind ourselves of is that the goal of exercism is to provide options to students so that they may make their own decisions about what is best. Obviously this comes with our own individual opinions as well. My opinion on the matter, being that I wrote the mentor notes, is that the turbofish operator, just like explicit typing, helps readibilty of code. If we see `variable_name.collect::<Vec<i32>>()`, we know exactly what we are ending up with. There are a lot of individuals that feel that type ellision is a great thing. I am the opposite. I would much rather be presented with precisely what I am working with at any given point.

Another troubling matter is that even an internet search yields very little on the subject, even the "official" Rust book.
This gives us the added benefit of teaching students very early on about the turbofish operator in a very simple example, which most all of my students have thanked me for.

coriolinus:
>  internet search yields very little on the subject, even the "official" Rust book.

This suggests to me that the syntax was designed to be used in extremis, not routinely.

Still, we agree that the point of exercism is to ensure students have all the knowledge necessary to form their own opinions. I'm still concerned that, as these mentor notes are written, mentors will present the turbofish operator as mandatory; students will learn the lesson "it is good practice to always add explicit types even when the compiler can figure out the right thing on its own." I don't want them to learn that lesson, because I believe it is wrong.

@TheDarkula would you be satisfied with a compromise which:

- removes the turbofish note from the section "a reasonable exercise should"
- softens the wording in "common suggestions" to 

    ```
    If they don't use the turbofish operator:

    While the type inferencer can frequently do the right thing with generics such as `.collect()`, it is sometimes helpful to use the turbofish operator `::<>` anyway to specify what type we want to build. With more code, it makes it easier to simply look at the collect instead of having to find the function's signature. When precisely one should insert an explicit type annotation is a matter of judgment. 
    ```

Tohnmeister:
Makes sense to me! :-)

efx:
@TheDarkula anything holding this back?

> . I'm still concerned that, as these mentor notes are written, mentors will present the turbofish operator as mandatory;

I totally agree  @coriolinus.


How about we summarize the main viewpoints from the slack thread and then include the reasoning in the mentoring notes? As a new mentor I would like to see expanded reasoning so I can better know the context of solutions and coach various students and their solutions.
For many solutions to reverse string the one line functions using the turbofish seems to go against the style practice of letting the compiler do its type inferencing.

coriolinus:
@SleeplessByte I'd like to request that you remove your change request. IMO this exercise is not intended to teach the use of the turbofish:

- It does not appear in [the exercise's topics](https://github.com/exercism/rust/blob/767963fcebb04f814b7d58f19f9323472fcb8b4e/config.json#L136-L147).
- Turbofish is not among the v3 concepts currently under consideration [1](https://exercism-team.slack.com/archives/CC48Y3H1B/p1573744795000100) [2](https://doodle.com/poll/sefgidb36tscgnm2).

PR #1200, opened 2019-07-31
rust: remove smileys from advice

Tohnmeister:
Smileys, i.m.o., are often over-used to soften messages. While giving serious advice, there's no need to use smileys. It will give the reader the idea that the message isn't too serious, while in fact, it is.

### DIFF
--- a/tracks/rust/exercises/bob/mentoring.md
+++ b/tracks/rust/exercises/bob/mentoring.md
@@ -88,8 +88,7 @@ match your_trimmed_variable.is_empty() {
     true => FINE,
     false => {
 ```
-Be aware that if you use Clippy, it will tell you that it is improper, it's purely a stylistic choice :)
-We can tell Clippy to allow it with `#[allow(clippy::match_bool)]`.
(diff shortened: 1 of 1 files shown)


kotp:
I agree with the sentiment, but also the changes to be made includes punctuation.  Emoticons are (in my opinion) not intended to be used as punctuation.  So even if the emoticons are left in, please at least, let us use punctuation.

tracks/rust/exercises/bob/mentoring.md
```diff
@@ -88,8 +88,7 @@ match your_trimmed_variable.is_empty() {
     true => FINE,
     false => {
 ```
-Be aware that if you use Clippy, it will tell you that it is improper, it's purely a stylistic choice :)
-We can tell Clippy to allow it with `#[allow(clippy::match_bool)]`.
+Be aware that if you use Clippy, it will tell you that it is improper, it's purely a stylistic choice. We can tell Clippy to allow it with `#[allow(clippy::match_bool)]`.
```
kotp: Not positive, but one line per sentence seems to work well, if we have adopted that sentiment in this repository, that would be the suggested change here.

Tohnmeister:
Thanks for reviewing.

PR #1208, opened 2019-08-06
Add Josh Goebel to JS mentors; fix avatar_url elsewhere

yyyc514:


yyyc514:
What is the diff between "" and null? Do you know?  Someone told me before "" was broken?  Is that correct?

rpottsoh:
Null will allow whatever has been designated as a default to be used instead of what is in the field.  I don't know that "" is broken, it just isn't desired.  In some places it could be broken I suppose.  I don't know about here though.

PR #1214, opened 2019-08-09
Add rna-transcription mentoring to typescript

micahwood:
Converting the rna-transcription notes from the Javascript track to work with Typescript as a starting point


tracks/typescript/exercises/rna-transcription/mentoring.md
```diff
@@ -0,0 +1,56 @@
...
+  A: 'U',
+  T: 'A',
+};
+
+function invalidInput() {
+  throw new Error('Invalid input DNA.')
+}
+
+export class Transcriptor {
+  toRna(sequence: string): string {
```
SleeplessByte:
```suggestion
  public toRna(sequence: string): string {
```

micahwood:
Yup that's exactly what I did. Thanks for the JS notes! I figured that would be the best place to start from as there currently aren't any notes available right now and it's the biggest part of the current queue. I have added the public modifier.

SleeplessByte:
Thank you!

PR #1222, opened 2019-08-15
go/twofer: comments for disapprove

tehsphinx:


automated-comments/go/two-fer/find_minimal_conditional.md
```diff
@@ -0,0 +1,5 @@
+Unless it makes the code more readable, repetition should be avoided.
+Try to find the _minimal conditional_ to make the code DRY (Don't Repeat Yourself).
+
+Note: DRY isn't always better. It may sometimes be better to duplicate code snippets rather than to introduce
```
bitfield: All of this is absolutely correct and wise advice, but maybe this is a little too much nuance for absolute beginners on the first exercise. (More or less any amount of nuance is usually too much in that situation.)

If you say to an experienced Gopher "Sometimes it's good to repeat yourself, and other times it isn't, depending on the situation," she will nod wisely and know exactly what you mean. But saying that to a beginner just leaves them mystified.

I'm not sure exactly what triggers this comment, so I can't necessarily suggest a better phrasing. But it might be something like "Short and simple code is more readable. Try to find the _minimal conditional_ here which passes the test."

automated-comments/go/two-fer/missing_share_with_function.md
```diff
@@ -0,0 +1 @@
+The `ShareWith` function is missing or couldn't be found.
```
bitfield: Maybe just "is missing"?

automated-comments/go/two-fer/replace_stub_comments.md
```diff
@@ -0,0 +1,5 @@
+The stub comments should be replaced with proper comments on package and function.
```
bitfield: How about "The stub comments are placeholders to show where the real documentation comments should go. Replace these with a brief, but helpful, description of the package, and a sentence or two explaining what the function does, what parameters it takes, and what it returns."

automated-comments/go/two-fer/strings_builder_used_for_concatenation.md
```diff
@@ -0,0 +1,6 @@
+`strings.Builder` is a highly optimized way to build complex strings.
```
bitfield: We've discussed this a few times before, and it's another classic problem of how much information to give the student at what stage. This exercise is perfectly well-served by using `fmt.Sprintf`, and it's an extremely valuable piece of the standard library to know.

Using concatenation is basically a performance optimization, and I'm very cautious about introducing students to this idea at such an early stage (especially because about 99% of students tend to err on the side of performance over readability. If anything, they need to be encouraged to make their code slower, but more readable!)

That said, if you _do_ opt to go the concatenation route, `strings.Builder` is absolutely the best general tool for this. It's only in certain weird, unusual, extreme cases where you have _very_ few strings that `+` is a good option. I think that's _way_ out of scope for `two-fer`.

So I'm not even sure that we should be _disapproving_ the use of `strings.Builder`. I'd be more inclined to make it a suggestion. What do you think?

automated-comments/go/two-fer/work_with_any_provided_name.md
```diff
@@ -0,0 +1,2 @@
+This solution references the names `Alice` and `Bob` directly in the code. This is a good first step!
+The funcion should be able to work with any name provided to it.
```
bitfield: 'function' not 'funcion'.

We might make this hint a bit clearer. "This is a good first step, but one way to improve this function would be to make it handle any name provided to it, not just Alice and Bob specifically."

automated-comments/go/two-fer/find_minimal_conditional.md
SleeplessByte: I agree. This is a targeted message to `two-fer`.

Expect the student to know programming, and thus be at least somewhat familiar with DRY and YAGNI. Perhaps:

```md
It is unnecessary to duplicate code in this instance. Find the _minimal condition_ to make the 
code DRY (Don't Repeat Yourself).
```

automated-comments/go/two-fer/missing_share_with_function.md
SleeplessByte: Actually, per the guidelines, I think this needs an action:

```suggestion
 Add the function `ShareWith`, as it is required and currently missing.
```

If you want action - cause;

```suggestion
The `ShareWith` function is required and missing. Add the function `ShareWith`.
```

If you want the reverse (ok in this instance, IMO).

Or something like that.

automated-comments/go/two-fer/replace_stub_comments.md
SleeplessByte: Make it an action! I took @bitfield 's advice and applied "actionifier"

```suggestion
Replace the stub comments with a brief, but helpful, description of the package, and a sentence or two explaining what the function does, what parameters it takes, and what it returns. `
```

I don't think `stub` needs to be explained.

automated-comments/go/two-fer/replace_stub_comments.md
```diff
@@ -0,0 +1,5 @@
...
+Accurate, helpful comments on how to use the code, are very important to Go developers. 
+This is a great guide to writing good comments: [Effective Go: Commentary](https://golang.org/doc/effective_go.html#commentary).
+The `godoc` tool will produce documentation from the code comments automatically.
```
SleeplessByte: I suggest removing the "but how" question from this comment:

```suggestion
The `godoc` tool produces documentation from the code comments automatically, when executed.
```

Or something similar.

automated-comments/go/two-fer/sharewith_signature_changed.md
```diff
@@ -0,0 +1,4 @@
+The signature of the `ShareWith` function should be `ShareWith(name string) string`.
```
SleeplessByte:
```suggestion
Fix the signature of `the ShareWith` function. The expected the signature is `ShareWith(name string) string`.
```

We reallly want to push them not to change it. I know that MUST is incorrect as they can _change_ the tests, but I don't think SHOULD in this case is _actionable_ or has enough gravitas. We can be stern here.

automated-comments/go/two-fer/sharewith_signature_changed.md
```diff
@@ -0,0 +1,4 @@
...
+Exercism is built on the TDD (Test Driven Development) principle. 
+The tests already exist and define the outward facing interface as well as the behaviour of the code to be implemented.
+Changing the signature will break the tests.
```
SleeplessByte:
```suggestion
Changing the signature breaks the tests.
```

automated-comments/go/two-fer/sharewith_signature_changed.md
```diff
@@ -0,0 +1,4 @@
+The signature of the `ShareWith` function should be `ShareWith(name string) string`.
+Exercism is built on the TDD (Test Driven Development) principle. 
+The tests already exist and define the outward facing interface as well as the behaviour of the code to be implemented.
```
SleeplessByte:
```suggestion
The tests already exist and define the outward facing interface, as well as the behaviour of the code to be implemented.
```

automated-comments/go/two-fer/strings_builder_used_for_concatenation.md
SleeplessByte: Given what @bitfield says and discussions I've had with @iHiD about this in JS/TS and Ruby, I don't think this should be a disapproval comment either. 

You might even want to _leave it out alltogether_. This is the first exercise. If both `sprintf` and `Stringbuilder` are _reasonable_ solutions, accept both. There are plenty exercises later in the track with strinbuilding where this comment would probably make more sense. 

If I'm completely wrong, just ignore me :heart:

automated-comments/go/two-fer/strings_builder_used_for_concatenation.md
```diff
@@ -0,0 +1,6 @@
...
+In simple cases like this using `fmt.Sprintf` is the preferred way. It has a lot
+of formatting options making it very flexible far beyond simple concatenation.
+
+The `+` operator is a very fast alternative for concatenations.
```
SleeplessByte: This is giving too many options.

I don't think _speed_ in `two-fer` is a concern at all. Another suggestion or idea:

- accept stringbuilder, you MAY add a "did you know about sprintf"
- accept sprintf, you MAY add "in simple use cases, + is a great alternative for concatenations"
- accept +, don't say anything. Instead, next time if they use + when it wasn't a good choice, tell them, then.

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
```diff
@@ -0,0 +1,4 @@
+`strings.Join` is perfect if the data already comes in the form of a `[]string`.
```
SleeplessByte:
```suggestion
Don't use `strings.Join` if the data is not of type `[]string`. 
```

I suggest being more stern here. Right now it's a statement where I can respond with "okay, so is it wrong now??". If you phrase it like the suggestion here, it's instantly clear(er).

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
```diff
@@ -0,0 +1,4 @@
...
+Take a look at `fmt.Sprintf` for this use case; it's a very powerful way of building up strings from multiple pieces.
```
SleeplessByte:
```suggestion
Use `fmt.Sprintf` instead it's a very powerful way of building up strings from multiple pieces.
```

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
```diff
@@ -0,0 +1,4 @@
...
+
+In case of simple concatenations, the `+` operator is a very fast alternative.
```
SleeplessByte: I suggest removing this. When the student uses Sprintf, it will be given as a comment on line 5 of [automated-comments/go/two-fer/strings_builder_used_for_concatenation.md](https://github.com/exercism/website-copy/pull/1222/files#diff-da6983c41277b728eff789945e1250e4).

```suggestion
```

automated-comments/go/two-fer/work_with_any_provided_name.md
```diff
@@ -0,0 +1,2 @@
+This solution references the names `Alice` and `Bob` directly in the code. This is a good first step!
```
SleeplessByte:
```suggestion
Change the function to work with any name provided. Currently the solution only works with `Alice` and `Bob`, which is a first good step.
```

I would prefer to drop line 2, and have this be actionable.

automated-comments/go/two-fer/work_with_any_provided_name.md
SleeplessByte: Instead of my suggestion above, I like this too. However, there is no "second" way to improve _this_ issue, so I would be it imperative.

```suggestion
This is a good first step. Improve this function to make it handle any name provided to it, not just Alice and Bob specifically.
```

automated-comments/go/two-fer/find_minimal_conditional.md
ErikSchierboom: I also agree that while DRY indeed isn't always better, perhaps we should not at this point bother the student with this subtle nuance?

automated-comments/go/two-fer/strings_builder_used_for_concatenation.md
ErikSchierboom: I don't know anything about Go, but in the C# track we have a similar situation where there are several different ways in which strings can be concatenated. The C# track differentiates between string concatenation that an experienced C# programmer would never use when solving two-fer (e.g. using `string.Join`), and string concatenation options that are less idiomatic (e.g. `string.Format`). In the former case, we disapprove. In the latter case, we approve with a comment. I'm not sure which scenario applies here though, as I don't know Go.

automated-comments/go/two-fer/find_minimal_conditional.md
kytrinyx: Agree that we should keep this short, concise, and specific to the two-fer case. 👍

automated-comments/go/two-fer/strings_builder_used_for_concatenation.md
kytrinyx: I don't think experienced Go developers would _ever_ use `strings.Builder` in the context of something like TwoFer. They'd use `fmt` or `+`, and I'm pretty sure their concern would be readability, not performance. At least the Go folks I've worked with.

Go has a culture of going as simple as possible, but no simpler. My gut instinct with Builder is to tell them that it's overkill for this exercise (no elaboration), and that experienced Go programmers would typically use `fmt` or simple concatenation with `+`.

automated-comments/go/two-fer/strings_builder_used_for_concatenation.md
kytrinyx: Yeah, I agree that too many options is not great at the start of the process. I would agree with mentioning `fmt` and accepting `+` without comment.

automated-comments/go/two-fer/replace_stub_comments.md
bitfield: You'd be surprised what does and doesn't need to be explained! I come across this all the time when writing technical books. One of the key things I do is try out drafts on readers of various experience levels, and ask them to point out things that didn't make sense to them. That always results in changes I'd never have guessed needed to be made.

automated-comments/go/two-fer/work_with_any_provided_name.md
bitfield: To native English speakers, the imperative "Improve this..." strikes just _fractionally_ the wrong note. It sounds like a parent telling a child to wipe their face or tie their shoelaces. I like to try to set a collaborative tone, where the student and I explore the problem together. A tiny bit of extra politeness can really grease the wheels here.

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
bitfield: Notwithstanding the specific advice here, which I think is right, I'm not sure 'stern' is exactly the right word for the tone we want. Clear, concise, and easy to act on, yes. But can't we be all those things and still be gentle, polite, encouraging, and collegial?

automated-comments/go/two-fer/work_with_any_provided_name.md
SleeplessByte: That's probably true for _some_ native English speakers, but this is not a mentor doing collaboration here. It's a machine spitting out comments.  Personally this kind of politeness to me is actually _rude_. 

Alas, if you can make it actionable without sounding impolite, I'm super happy :heart:

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
SleeplessByte: No. The analyzer should not spit out _emotive_ content. It should sound neutral (hence me using "stern").

automated-comments/go/two-fer/replace_stub_comments.md
SleeplessByte: In that case leave it in 😅. We're probably targetting experienced developers (in a different language), but even then, you're probably right 😁

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
bitfield: Understood. I think this probably exposes a pretty fundamental difference of approach; I'm not sure there's anything more I can usefully contribute to the automated analyzer as a project, if that's the case. We'd just end up arguing over the same points for every comment.

SleeplessByte: Does [this](https://github.com/exercism/automated-mentoring-support/blob/5b01bae1faced4c771308bab10a4fd6277e2a4d1/docs/comment_copy_guidelines.md) perhaps help?

SleeplessByte: (But hey, I'm not the authority here! Just following what we came up with earlier 😅).

Let's also ask @iHiD and @kytrinyx from a product/platform point of view.

automated-comments/go/two-fer/sharewith_signature_changed.md
tehsphinx:
```suggestion
Fix the signature of the `ShareWith` function. The expected signature is `ShareWith(name string) string`.
```

automated-comments/go/two-fer/strings_builder_used_for_concatenation.md
tehsphinx: This comment is only shown if they used the `strings.Builder`.

So they either knew about it already as they are not new to Go. In this case they probably thought their solution is super fast, when it is not. They can handle this comment.

Or they searched for how to build a string in Go and found this (comparably) very complicated and highly optimized piece of software, and now think that string concatenation in Go is really complicated, because you have to use this weird `strings.Builder`.

In my opinion they should know either way, that `strings.Builder` is rarely used in production code because most cases when a strings needs building, `fmt` or `+` is simpler and even faster in case of `+`.

tehsphinx: I'm happy to make this an approval comment though, if that is fine with @kytrinyx

automated-comments/go/two-fer/strings_builder_used_for_concatenation.md
tehsphinx: Probably a good idea to provide less options, yes. I'll remove the `+` part.

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
tehsphinx:
```suggestion
Don't use `strings.Join` if the data is not of type `[]string`.
```

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
tehsphinx:
```suggestion
Use `fmt.Sprintf` instead. It's a very powerful way of building up strings from multiple pieces.
```

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
```diff
@@ -0,0 +1,4 @@
+`strings.Join` is perfect if the data already comes in the form of a `[]string`.
+Take a look at `fmt.Sprintf` for this use case; it's a very powerful way of building up strings from multiple pieces.
+
```
tehsphinx:
```suggestion
```

automated-comments/go/two-fer/work_with_any_provided_name.md
tehsphinx:
```suggestion
```

automated-comments/go/two-fer/strings_builder_used_for_concatenation.md
tehsphinx: @kytrinyx: Did not see your message before I answered this (still had the tab open). I simplified it to your suggestion but left out the `Experienced Go programmers` so students don't feel like we are labelling them inexperienced.

automated-comments/go/two-fer/strings_builder_used_for_concatenation.md
```diff
@@ -0,0 +1,2 @@
+The `strings.Builder` is overkill for this simple use case.
```
ErikSchierboom:
```suggestion
Using `strings.Builder` is overkill for this simple use case.
```

tehsphinx: Thx! Seems a common mistake I do 😄

tehsphinx:
Thanks again to all for your ideas, suggestions and corrections, @bitfield, @SleeplessByte, @ErikSchierboom and @kytrinyx !!!

automated-comments/go/two-fer/strings_builder_used_for_concatenation.md
kytrinyx: > avoid students possibly feel like we are labelling them "inexperienced".

Good call!

SleeplessByte:
I think this is good to go. Let's make additional changes in a subsequent PR?

PR #1223, opened 2019-08-15
go/general: comments for disapprove

tehsphinx:


automated-comments/go/general/avoid_init_function.md
```diff
@@ -0,0 +1,4 @@
+Using the `init` function should be avoided:
```
SleeplessByte:
```suggestion
Avoid using the `init` function:
```

automated-comments/go/general/avoid_init_function.md
```diff
@@ -0,0 +1,4 @@
...
+1) It slows down startup time and the developer using the package cannot do anything about it apart from removing the `init` function.
```
SleeplessByte:
```suggestion
1) It slows down startup time and the developer using the package cannot do anything about that slow-down, apart from removing the `init` function.
```

automated-comments/go/general/avoid_printing_and_logging.md
```diff
@@ -0,0 +1,8 @@
+Usually it is not a good idea to output something to `stdout` or `stderr` in lower level packages.
```
SleeplessByte:
```suggestion
Avoid printing and logging via `stdout` and `stderr`.

Usually it is not a good idea to output something to `stdout` or `stderr` in lower level packages.
```

automated-comments/go/general/avoid_printing_and_logging.md
```diff
@@ -0,0 +1,8 @@
...
+That is done in higher level or the `main` package, that have the necessary context to for example know what to do with an error.
+Using `log.Println` or `fmt.Println` in a lower level package can not be overwritten from the outside.
+Best avoid them in the exercises here. They are useful for debugging but should be removed before submitting.
```
SleeplessByte:
```suggestion
Printing and logging are useful for debugging, but should be removed before submitting an exercise, or releasing a package.
```

Or something like that. The overal advice here is _great_, and you probably want to make sure it's clear that you basically never want it, unless you're writing an application (if I understood correctly).

automated-comments/go/general/avoid_printing_and_logging.md
```diff
@@ -0,0 +1,8 @@
...
+
+It might be that you feel like an error should be returned by a function but there is no error return value in the function signature.
```
SleeplessByte: Can this be rephrased so there is no "you" or "feel"?

automated-comments/go/general/avoid_printing_and_logging.md
```diff
@@ -0,0 +1,8 @@
...
+In the real world you could change this, but in these exercises the function signature is pre-determined.
```
SleeplessByte: Can this be rephrased so there is no "you"?

automated-comments/go/general/avoid_printing_and_logging.md
```diff
@@ -0,0 +1,8 @@
...
+Best discuss this with your mentor. An exception might not be necessary after all or the exercise might need to be adjusted.
```
SleeplessByte: The mentor most likely will get this comment, not the student. 

image

So perhaps flip the wording?

automated-comments/go/general/gofmt_not_used.md
```diff
@@ -0,0 +1,22 @@
+Code formatting in Go is very important to the community. Take a look at `gofmt` and run it on the solution.
```
SleeplessByte:
```suggestion
Run `gofmt` on the solution, which helps formatting code in Go.

Code formatting in Go is very important to the community.
```

automated-comments/go/general/gofmt_not_used.md
```diff
@@ -0,0 +1,22 @@
...
+Most IDEs can be set up to auto-format the code on every save. For example:
+
+- Visual Studio Code has [great Go support](https://code.visualstudio.com/docs/languages/go)
+
```
SleeplessByte:
```suggestion
```

I suggest removing these to make sure it won't be double line-height list.

automated-comments/go/general/gofmt_not_used.md
```diff
@@ -0,0 +1,22 @@
...
+- Vim also has a [fantastic Go extension](https://github.com/fatih/vim-go)
+
```
SleeplessByte:
```suggestion
```

I suggest removing these to make sure it won't be double line-height list.

automated-comments/go/general/golint_not_satisfied.md
```diff
@@ -0,0 +1,14 @@
+There's a great tool called [`golint`](https://github.com/golang/lint) which will examine the code for common problems and style issues.
+Try running `golint` on the solution, e.g.: `golint two_fer.go`; it will make some useful suggestions.
```
SleeplessByte:
```suggestion
Run `golint` on the solution, e.g.: `golint two_fer.go`; it will make some useful suggestions.
```

The problem with `try` is that, if it fails, a student might leave it. With an imperative it's more likely that they'll complain if it doesn't work. (I have no proof for this 😅, only anecdotal evidence).

automated-comments/go/general/golint_not_satisfied.md
```diff
@@ -0,0 +1,14 @@
...
+
+It's a good idea to always check ones code with `golint`.
+When writing Go software for production, many build pipelines will automatically fail if the code doesn't pass `golint`.
+
+If you are keen on getting much more feedback from code linters there is the [golangci-lint](https://github.com/golangci/golangci-lint) project
+allowing you to run a lot of linters with one command.
+You can create a config file for your project to adjust the behaviour to your needs.
```
SleeplessByte:
```suggestion
A project config file can adjust the behaviour to a project's needs.
```

automated-comments/go/general/remove_extra_function.md
```diff
@@ -0,0 +1 @@
+One or more extra functions have been created in this solution. They make the code harder to read and can be removed.
```
SleeplessByte:
```suggestion
One or more extra functions have been created in this solution. Remove them as they make the code harder to read.
```

Perhaps?

automated-comments/go/general/remove_main_function_and_correct_package_name.md
```diff
@@ -0,0 +1,5 @@
+A `main` function is used as an entry point to a program. Exercism is using TDD (Test Driven Development).
+That means the code is executed by running the tests to make them pass. To run the tests use `go test` from the terminal.
+The tests will pass if the functionality is implemented correctly.
+
+Remove the `main` function and rename the package name back to `twofer`. Then run the tests.
```
SleeplessByte: I suggest moving this line to line 1. And then having the rest of this great comment.

automated-comments/go/general/strings_compare_used.md
```diff
@@ -0,0 +1,4 @@
+`strings.Compare` is an interesting choice to use here. It's [official documentation](https://golang.org/pkg/strings/#Compare) states:
+
+> Compare is included only for symmetry with package bytes.
+> It is usually clearer and always faster to use the built-in string comparison operators ==, <, >, and so on.
```
SleeplessByte: Since this is probably a disapproval comment, add an action

automated-comments/go/general/strings_trim_space_used.md
```diff
@@ -1,3 +1,5 @@
-For all Exercism exercises, assume that the code you write is part of an internal library, 
+There is no need to clean up the code for example with `strings.TrimSpace`.
```
SleeplessByte:
```suggestion
There is no need to clean up the code, for example using `strings.TrimSpace`.
```

automated-comments/go/general/syntax_error.md
```diff
@@ -0,0 +1 @@
+Unfortunately there is a syntax error in this solution. Try running the tests in a terminal with `go test`.
```
SleeplessByte:
```suggestion
Unfortunately there is a syntax error in this solution. Run `go test` in a terminal.
```

automated-comments/go/general/avoid_printing_and_logging.md
ErikSchierboom: I like this suggestion.

automated-comments/go/general/gofmt_not_used.md
```diff
@@ -0,0 +1,22 @@
...
+- [Jetbrains Goland](https://www.jetbrains.com/go/) is an entire IDE for Go
+
+To run `gofmt` manually:
```
ErikSchierboom: I like this very clear example!

automated-comments/go/general/golint_not_satisfied.md
```diff
@@ -0,0 +1,14 @@
+There's a great tool called [`golint`](https://github.com/golang/lint) which will examine the code for common problems and style issues.
+Try running `golint` on the solution, e.g.: `golint two_fer.go`; it will make some useful suggestions.
+
+It's a good idea to always check ones code with `golint`.
```
ErikSchierboom:
```suggestion
It's a good idea to always check one's code with `golint`.
```

Or perhaps just:

```suggestion
It's a good idea to always check the code with `golint`.
```

automated-comments/go/general/syntax_error.md
ErikSchierboom: To me, it looks a little bit like the student might think that running `go test` might solve the syntax error :D

Maybe something like:

```suggestion
Unfortunately there is a syntax error in this solution. This causes the tests to fail, which can be verified by running the tests in a terminal with `go test`.
```

I'm not sure how to elegantly word the action a student should take (fix the syntax error).

automated-comments/go/general/avoid_printing_and_logging.md
kytrinyx: I don't think we should use the word `exception` here, as it is ambiguous. Do we mean raising an exception (of which there are none in Go?) or do we mean an exception that proves the rule?

automated-comments/go/general/remove_extra_function.md
kytrinyx: This is tricky. What is the context of this comment? Is it an exercise that is supposed to have exactly one function? Or is it an exercise that might be solved with three but definitely not four or five functions?

I think I'd like this to be more imperative and use active voice. Something like:

> Simplify this code to use fewer functions. Unnecessary functions make the code harder to read.

Except that I don't like this either (so I didn't make it a suggestion).

automated-comments/go/general/avoid_printing_and_logging.md
tehsphinx:
```suggestion
Sometimes it looks like the function signature should be changed to return an error.
```

automated-comments/go/general/avoid_printing_and_logging.md
tehsphinx:
```suggestion
In contrast to the real world, the function signature is pre-determined in these exercises.
```

automated-comments/go/general/avoid_printing_and_logging.md
tehsphinx:
```suggestion
There's room for discussion: to return an error might not be necessary after all or the exercise might need to be adjusted.
```

automated-comments/go/general/remove_extra_function.md
tehsphinx: It is supposed to be applied only on a few of the early exercises: `twofer`, `hamming`, maybe `raindrops`, `scrabble` and `isogram` (open for discussion!) where it makes it harder to read when it is split into 2 functions. The condition to trigger this is `len(funcs) > 1`.

automated-comments/go/general/syntax_error.md
tehsphinx: Maybe in this casy it is obvious ;)

automated-comments/go/general/avoid_printing_and_logging.md
```diff
@@ -0,0 +1,10 @@
+Avoid printing and logging via `stdout` and `stderr`.
+
+Usually it is not a good idea to output something to `stdout` or `stderr` in lower level packages.
+That is done in higher level or the `main` package, that have the necessary context to for example know what to do with an error.
```
ErikSchierboom:
```suggestion
That is done in higher level or the `main` package, that have the necessary context to know what to do with an error.
```

automated-comments/go/general/remove_extra_function.md
kytrinyx: In that case, can we name the file "single_function" or something explicit like that?

```suggestion
Solve this exercise with a single function. Additional functions in this case make the code harder to read.
```

tehsphinx:
Added general comments about `missing functions` and `function signature changed` already, so I can switch to it in future versions of the analyzer.

tehsphinx:
@iHiD: Can we get this merged? Even if we might rephrase things again? Currently mentors are missing these comments and then wonder why it says not to approve.

iHiD:
Merging. We can move forward on new PRs.

PR #1224, opened 2019-08-15
go/hamming: disapprove comments

tehsphinx:


automated-comments/go/hamming/comparing_bytes_disguised_as_runes.md
```diff
@@ -0,0 +1,4 @@
+Converting the `bytes` to `runes` to compare disguises the problem.
+It works perfectly in the context of this exercise as there are only ascii letters in DNA strands.
```
bitfield: I think we would usually write 'ASCII', not 'ascii'—it's an acronym.

automated-comments/go/hamming/comparing_bytes_disguised_as_runes.md
```diff
@@ -0,0 +1,4 @@
...
+
+Try creating more tests with non-ascii characters and see what happens.
```
bitfield: I think this is likely to be a bit beyond the capabilities of students so early in the track, and unless they do that, they won't get the point. Do we need a more detailed explanation of the problem here?

I'm not completely clear what triggers this comment, but I'm guessing it's something like indexing the string by byte, and then converting that byte to a rune before comparing it. The right thing to do (I usually propose) is to convert both strings to rune sequences _first_, before doing anything with them. Then it's meaningful to compare their lengths, and safe to iterate over them comparing element with element.

automated-comments/go/hamming/distance_signature_changed.md
```diff
@@ -0,0 +1,4 @@
+The signature of the `Distance` function should be `Distance(a, b string) (int, error)`.
+Exercism is built on the TDD (Test Driven Development) principle.
```
bitfield: This is absolutely right, but a key part of TDD is also updating the tests to take account of changes to the code under test!

A simple example is in `grains`, when students often implement Total() by calling Square repeatedly and ignoring the error value returned. I sometimes guide them towards checking the error and returning it if non-nil, which of course changes the function signature and thus requires updating the tests.

I'd be cautious about discouraging students from changing the function signature _when they have a good reason_ for doing so. Perhaps say rather "The test-driven development (TDD) method which Exercism uses means that the function signature must always match what the tests expect. If the function signature is updated, the tests which call it must also be updated to match, or they won't work."

automated-comments/go/hamming/missing_distance_function.md
```diff
@@ -0,0 +1 @@
+The `Distance` function is missing or couldn't be found.
```
bitfield: Maybe just 'is missing'? Missing and can't be found are really more or less the same thing, aren't they? Possibly the student forgot to upload the solution file, and just submitted the tests instead (this happens surprisingly often). We might give them a hint if that's the case.

automated-comments/go/hamming/strings_split_used.md
```diff
@@ -0,0 +1,2 @@
+The type in Go to hold a single character is `rune`. A string can be transformed into a `[]rune` (slice of runes) via type conversion: `var aRunes = []rune(a)`.
```
bitfield: This is good advice, but I wonder if we can reframe it in a slightly more positive way. "This exercise involves examining a string character by character (in Go characters are referred to as _runes_, by the way). Splitting the string into a slice of strings, one for each rune, is one way to do this, but it's not the most efficient. Go has a data type specifically for storing individual runes: `rune`. Try converting the string to a slice of runes like this: `ar := []rune(a)`."

automated-comments/go/hamming/comparing_bytes_disguised_as_runes.md
```diff
@@ -0,0 +1,4 @@
+Converting the `bytes` to `runes` to compare disguises the problem.
```
SleeplessByte: What problem?

automated-comments/go/hamming/comparing_bytes_disguised_as_runes.md
SleeplessByte:
```suggestion
It works perfectly in the context of this exercise as there are only ASCII letters in DNA strands.
```

automated-comments/go/hamming/comparing_bytes_disguised_as_runes.md
SleeplessByte:
```suggestion
Try creating more tests with non-ASCII characters and see what happens.
```

(same fix as above).

For the rest I agree with @bitfield .

automated-comments/go/hamming/comparing_bytes_disguised_as_runes.md
SleeplessByte:
```suggestion
Such conversion works perfectly in the context of this exercise as there are only ASCII letters in DNA strands.
```

Grammarly, `it` refers to "the problem"

automated-comments/go/hamming/distance_signature_changed.md
SleeplessByte: @bitfield not to disregard your comment, which I think is valid, I also agree very much with @tehsphinx for the simple reason that we have one more constraint: we need to be able to make some assumptions about the code when analyzing it, for now. 

Discouraging changing tests is one way to be able to do that 😅.

Is it common here to change the tests? If it's not common and not _expected_, I would discourage it, like Gabriel currently does.

automated-comments/go/hamming/distance_signature_changed.md
```diff
@@ -0,0 +1,4 @@
...
+The tests already exist and define the outward facing interface as well as the behaviour of the code to be implemented.
+Changing the signature will break the tests.
```
SleeplessByte:
```suggestion
Changing the signature breaks the tests.
```

automated-comments/go/hamming/distance_signature_changed.md
```diff
@@ -0,0 +1,4 @@
+The signature of the `Distance` function should be `Distance(a, b string) (int, error)`.
+Exercism is built on the TDD (Test Driven Development) principle.
+The tests already exist and define the outward facing interface as well as the behaviour of the code to be implemented.
```
SleeplessByte:
```suggestion
The tests already exist and define the outward facing interface, as well as the behaviour of the code to be implemented.
```

automated-comments/go/hamming/missing_distance_function.md
SleeplessByte:
```suggestion
Add the function `Distance`, as it is required and currently missing.
```

Perhaps same comment as my suggestion to two-fer

automated-comments/go/hamming/strings_split_used.md
SleeplessByte: Per the guidelines, I think we should put the _action first_, but right now there isn't any. This automated mentoring comment are three statements without actionable.

Perhaps:

````suggestion
Improve the efficiency by converting the string to a slice of `rune`s like this:

```go
ar := []rune(a)
```

The type in Go to hold a single character is `rune`. Splitting the string into a slice of single character strings and working with them is a lot slower than ...
````

I want to incorporate everything @bitfield suggested butttt I don't feel very articulate right now 😭

File: automated-comments/go/hamming/comparing_bytes_disguised_as_runes.md (hunk 2 above)
Comment (tehsphinx) [reply automated-comments/go/hamming/comparing_bytes_disguised_as_runes.md]: Yes, this is triggered when doing something like this:

```go
for i, r := range a {
    if r == rune(b[i])
}
```

I'll look into explaining the "problem" better. I think I would stick with letting the user find out what to do about it as this is more of a challenge rather than a "must". That should probably be clearer as well.

automated-comments/go/hamming/distance_signature_changed.md
tehsphinx: Generally I think this is a valid point but as @SleeplessByte said: We are not really able to incorporate that into the analysers.

For now this comment will only be visible to mentors anyway as it is a `disapprove` comment. It is up to the mentor to decide if he ignores it and works with the student on a changed function signature or not.

I'd also argue, that the function signatures in the tests are pretty well thought through and if a change is regularly required we should change that in the exercises.

ErikSchierboom: Personally, I usually discourage changing the tests, as it makes communicating with the student harder for the mentor, which can no longer assume he knows how the tests are structured. It also makes it harder for the analyzer to work :)

automated-comments/go/hamming/distance_signature_changed.md
ErikSchierboom: Same comment here about potentially converting this to a general comment with parameters.

automated-comments/go/hamming/missing_distance_function.md
ErikSchierboom: Same comment here about potentially converting this to a general comment with parameters.

automated-comments/go/hamming/distance_signature_changed.md
tehsphinx: Will do that in the future. The current analyzer that is online already provides these comments on disapprove, so I am trying to get them in there without a new release for now.

tehsphinx:
Thanks all for your review! I incorporated the changes and rephrased some comments completely.

Feel free to have another look!

automated-comments/go/hamming/comparing_bytes_disguised_as_runes.md
tehsphinx: Rephrased

tehsphinx:
Let's get this one perfect first, then I'll deal with the others as there are some commonalities.

automated-comments/go/hamming/strings_split_used.md
```diff
@@ -0,0 +1,8 @@
+Use the built in type `rune` to work with individual characters in Go.
```
kytrinyx:
```suggestion
Use the built-in type `rune` to work with individual characters in Go.
```

automated-comments/go/hamming/strings_split_used.md
```diff
@@ -0,0 +1,8 @@
...
+A `string` can be converted into a `[]rune` (slice of runes) with:
+
+```go
+aRunes := []rune(a)
+```
+
+Splitting the string into a lot of small strings (one for each rune) works, but is not the most efficient.
```
kytrinyx:
```suggestion
Splitting the string into a lot of small strings (one for each rune) works, but is inefficient.
```

automated-comments/go/hamming/comparing_bytes_disguised_as_runes.md
```diff
@@ -0,0 +1,5 @@
+Converting a `byte` to a `rune` works perfectly in the context of this exercise as there are only ASCII letters in DNA strands.
+ASCII characters consist of one `byte` each. As soon as non-ASCII characters get involved this will break as they consist of 
+up to 4 bytes.
```
kytrinyx: Should this line be combined with the previous one? (Not sure how whitespace is treated in the UI)

SleeplessByte: New lines are mark-down new lines, so a treated as a space if there's only one.

tehsphinx:
Thanks @kytrinyx !

automated-comments/go/hamming/comparing_bytes_disguised_as_runes.md
```diff
@@ -0,0 +1,5 @@
+Converting a `byte` to a `rune` works perfectly in the context of this exercise as there are only ASCII letters in DNA strands.
+ASCII characters consist of one `byte` each. As soon as non-ASCII characters get involved this will break as they consist of
```
SleeplessByte:
```suggestion
ASCII characters consist of one `byte` each. As soon as non-ASCII characters get involved this method of comparison breaks as those characters consist of 
```

Just solidifying those references, as this sorta refers back to "this exercise" when "this method of comparing" (converting a byte to a rune and then doing the comparison) is meant.

tehsphinx:
@iHiD: Can we get this merged? Even if we might rephrase things again? Currently mentors are missing these comments and then wonder why it says not to approve.

tehsphinx:
@SleeplessByte: are your requested changes implemented?

iHiD:
Merging. We can move forward on new PRs.

PR #1227, opened 2019-08-16
Add leap analyzer disapprove comments

ricemery:


automated-comments/scala/leap/no_loops.md
```diff
@@ -0,0 +1 @@
+Loop operations are not needed.
```
ErikSchierboom: Is is maybe useful to suggest what the student could look at?

ricemery: @ErikSchierboom  thanks for reviewing.

On 'looping' I am not sure what else I would say. I'm not sure why anyone would even use a loop for this exercise.  Maybe say

'Loop operations are not needed. Use a single `Boolean` expression instead of loop operators.'

ErikSchierboom: I really like that suggestion! It makes it very actionable for the user.

ricemery: Done. 
Thanks @ErikSchierboom

ricemery:
Thanks Erik!

PR #1230, opened 2019-08-16
csharp/two-fer: make single string automated comment more generic

ErikSchierboom:


automated-comments/csharp/two-fer/use_single_formatted_string_not_multiple.md
```diff
@@ -1 +1 @@
-There are two `return` statements returning very similar strings. How could this code be structured so that this duplication is eliminated? This will involve having just one `return` statement that returns a single `string`.
+There are two very similar `string` instances being returned. How could this code be structured such that this duplication is eliminated? This will involve removing any branching and returning just a single `string`.
```
SleeplessByte: This only thing I don't like is that there is no _clear_ action. Apart from that this change is good!

ErikSchierboom: Yeah I agree. The previous version was better in that, but this also works for ternary operator returns. Thanks for the review!

ErikSchierboom: So one option is to split this comment into two.

SleeplessByte: Example?

ErikSchierboom: Well, the current comment notes that only one `return` should be returned, but this also fires when someone uses a ternary operator in an expression bodied member. In the latter case, no return keyword can be removed, as there is none. I could create a new comment for the ternary version, but it seems a little bit like overkill.

SleeplessByte: Ah. I understand now. Yes I also have a more general message to remove the condition / duplication.

ErikSchierboom: So what do you think about the current wording?

SleeplessByte: Let's just do this and change it later if we come up with something better?

ErikSchierboom: Yep! I like that.

ErikSchierboom:
Thanks for the review!

PR #1233, opened 2019-08-16
Rust: Make automatic suggestion for reverse-string more precise

Aloso:
A `char` in Rust is a Unicode scalar value:

```rust
assert_eq!("猫".len(), 3);
assert_eq!("猫".chars().count(), 1);
```

So we don't need a library to support Unicode. However, some characters consist of multiple `char`s:

```rust
assert_eq!("ü".chars().count(), 2);
assert_eq!("\r\n".chars().count(), 2);
```

These are called "grapheme clusters", and should not be confused with "non-ASCII characters".


ZapAnton:
I do agree that the comment should be more precise.

Just a small note - perhaps a link that explains what grapheme clusters are could be added to the comment? I am  just afraid  that the grapheme concept may be a little more confusing for the students than the concept of  `non-ASCII`.

SleeplessByte:
@Aloso mentioned that students figure that out by themselves if this comment is given. Let's merge this and if we then find out it's _not_ clear enough, we'll add additional data?

SleeplessByte:
Thank you for doing this @Aloso !

PR #1244, opened 2019-08-18
Add two-fer disapproval comments

iHiD:
I'm going to improve this later by being explicit about the variable names, but the analyzer isn't outputting those yet, so this can't use them, so we need to have this merged first.

### DIFF
--- a/automated-comments/ruby/general/explicit_return.md
+++ b/automated-comments/ruby/general/explicit_return.md
@@ -2,7 +2,7 @@ Remove the explict return to simplify the method.
 
 In Ruby, a method will return the result of the last evaluated statement. Explicitly using the `return` keyword is therefore unnecessary, except when returning early in a method.
 
-For example, this will print `16` as the `multiply` method will return the result of `a * b`, it's final expression.
+For example, this will print `16` as the `multiply` method returns the result of `a * b`, its final expression.
 
 ```ruby
 def multiple(a,b)
--- a/automated-comments/ruby/general/explicit_return.md
+++ b/automated-comments/ruby/general/explicit_return.md
@@ -1,0 +1,13 @@
+Remove the explict return to simplify the method.
+
+In Ruby, a method will return the object that was returned from the last evaluated statement. Explicitly using the `return` keyword is therefore unnecessary unless we are returning early in a method.
+
+For example, this will print `16` as the `multiply` method will return the result of `a * b`, it's final expression.
+
+```ruby
+def multiple(a,b)
+  a * b
+end
+
+print multiple(2, 8)
+```
--- a/automated-comments/ruby/general/explicit_return.md
+++ b/automated-comments/ruby/general/explicit_return.md
@@ -1,6 +1,6 @@
 Remove the explict return to simplify the method.
 
-In Ruby, a method will return the object that was returned from the last evaluated statement. Explicitly using the `return` keyword is therefore unnecessary unless we are returning early in a method.
+In Ruby, a method will return the result of the last evaluated statement. Explicitly using the `return` keyword is therefore unnecessary, except when returning early in a method.
 
 For example, this will print `16` as the `multiply` method will return the result of `a * b`, it's final expression.
 
--- a/automated-comments/ruby/general/incorrect_indentation.md
+++ b/automated-comments/ruby/general/incorrect_indentation.md
@@ -1,0 +1,1 @@
+Use two spaces to indent this code. In Ruby, using two spaces is a strong community standard - we never use tabs, and always only use two spaces.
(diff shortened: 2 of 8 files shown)


automated-comments/ruby/two-fer/incorrect_default_param.md
```diff
@@ -0,0 +1 @@
+What could the default value of the paramater be set to in order to avoid having to use a conditional?
```
kytrinyx: At first glance I was unsure about the question, thinking that it would be posted to the student automatically, and I don't feel like automated comments should encourage an answer.

However, since this is only given to the mentor as a suggestion, I think it's probably fine.

automated-comments/ruby/two-fer/reassigning_param.md
```diff
@@ -0,0 +1 @@
+What could the default value of the paramater be set to in order to avoid having to reassign the variable?
```
ErikSchierboom:
```suggestion
What could the default value of the parameter be set to in order to avoid having to reassign the variable?
```

automated-comments/ruby/two-fer/missing_default_param.md
```diff
@@ -0,0 +1 @@
+There is not a paramater with a default value so the tests will fail.
```
ErikSchierboom:
```suggestion
There is not a parameter with a default value so the tests will fail.
```

automated-comments/ruby/two-fer/incorrect_default_param.md
ErikSchierboom:
```suggestion
What could the default value of the parameter be set to in order to avoid having to use a conditional?
```

iHiD:
> Looking good, just some tiny nits.

@ErikSchierboom And this is the reason I exclusively use the abbreviation ("param") in code ;)

automated-comments/ruby/two-fer/incorrect_default_param.md
iHiD: @kytrinyx Yeah - I wonder if we should change the rules for disapprove comments now they come from a human. The rules we wrote for approve are to come from computers.

automated-comments/ruby/general/explicit_return.md
```diff
@@ -0,0 +1,13 @@
+Remove the explict return to simplify the method.
+
+In Ruby, a method will return the object that was returned from the last evaluated statement. Explicitly using the `return` keyword is therefore unnecessary unless we are returning early in a method.
```
SleeplessByte:
```suggestion
In Ruby, a method will return the result of the last evaluated statement. Explicitly using the `return` keyword is therefore unnecessary, unless when returning early in a method.
```

A bit more concise and without personas?

automated-comments/ruby/general/explicit_return.md
```diff
@@ -0,0 +1,13 @@
...
+
+For example, this will print `16` as the `multiply` method will return the result of `a * b`, it's final expression.
```
SleeplessByte:
```suggestion
For example, this will print `16` as the `multiply` method returns the result of `a * b`, its final expression.
```

I think that's meant to be just possessive?

automated-comments/ruby/general/incorrect_indentation.md
```diff
@@ -0,0 +1 @@
+Use two spaces to indent this code. In Ruby, using two spaces is a strong community standard - we never use tabs, and always only use two spaces.
```
SleeplessByte:
```suggestion
Use two spaces to indent this code. In Ruby, using two spaces is a strong community standard - never use tabs, always use two spaces.
```

Without the persona this is more of a  "never, always" rule instead of "we, the community". Idk. ignore if you hate 🗡

automated-comments/ruby/two-fer/splat_args.md
```diff
@@ -0,0 +1 @@
+Define an explicit `name` variable rather than using the splat argument keyword (`*`). This will make the code clearer to someone else reading it, and involves less manipulation to access the value of the argument.
```
SleeplessByte:
```suggestion
Define an explicit `name` variable rather than using the splat argument keyword (`*`). This makes the code clearer to read, and involves less manipulation to access the value of the argument.
```

It's not just for other people, but also for yourself in an arbitrary timeframe.

automated-comments/ruby/general/explicit_return.md
iHiD: Updated `unless` for `except`

automated-comments/ruby/general/explicit_return.md
iHiD: Ridiculous language ;)

iHiD:
@SleeplessByte @ErikSchierboom @kytrinyx Thanks all! Merging :)

PR #1253, opened 2019-08-23
Java: two-fer mentoring notes to discourage String.format

jmrunkle:
Basically, we found out that `String.format` has distinctly bad performance characteristics and we want to steer our students away from relying heavily on this when a plain old string concatenation will do the trick.

We already updated the automated comments about this, but now we are updating the mentoring notes to match.

### DIFF
--- a/tracks/java/exercises/two-fer/mentoring.md
+++ b/tracks/java/exercises/two-fer/mentoring.md
@@ -3,30 +3,45 @@
 ```java
 class Twofer {
   String twofer(String name) {
+    return "One for " + name == null ? "you" : name + ", one for me.";
+  }
+}
+```
+
+```java
+import java.util.Optional;
+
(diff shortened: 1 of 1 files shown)


tracks/java/exercises/two-fer/mentoring.md
```diff
@@ -23,11 +23,26 @@ import java.util.Optional;
 
 class Twofer {
   String twofer(String name) {
-    return String.format("One for %s, one for me.", Objects.toString(name, "you"));
+    return "One for " + Objects.toString(name, "you") + ", one for me.";
+  }
+}
+```
+
+### Not recommended
```
SleeplessByte:
```suggestion
#### Not recommended
```

SleeplessByte:
Great! ❤️

PR #1264, opened 2019-08-28
Change some sample disapprove comments to feel less 'stern'.

iHiD:
This PR tries to make some disapproval comments feel more like they're coming from a human. It is aimed to give some discussion points. If, in the comments below, people suggest some other examples that they'd like seen worked on, I will add extra commits refactoring those ones too.

### Background

We put together some guidelines for comments given with approval. We wanted them to be empathetic but not personal as these comments are posted by a "bot" not a human. Also, in these situations we're not asking for a student to do anything - we're just informing them of something, so the aim is not to be conversational.

We inadvertently applied the same guidelines to comments give as part of disapproval. As these are being sent by a mentor, the unpersonal tone doesn't work as well, so they can be both conversational and much more personable. The are also requesting changes, which adds another dynamic.

This PR aims to fix that with some samples that we can discuss. Please go wild :)

### DIFF
--- a/automated-comments/csharp/general/invalid_method_signature.md
+++ b/automated-comments/csharp/general/invalid_method_signature.md
@@ -1,1 +1,1 @@
-Change the signature of the `%{name}` method to be `%{signature}`.
+In order to make the tests pass, you need to change the signature of the `%{name}` method to be `%{signature}`.
--- a/automated-comments/csharp/general/invalid_method_signature.md
+++ b/automated-comments/csharp/general/invalid_method_signature.md
@@ -1,1 +1,1 @@
-In order to make the tests pass, you need to change the signature of the `%{name}` method to be `%{signature}`.
+In order to make the tests pass, the signature of the `%{name}` method needs to be changed to `%{signature}`.
--- a/automated-comments/csharp/general/missing_method.md
+++ b/automated-comments/csharp/general/missing_method.md
@@ -1,1 +1,1 @@
-Add a method named `%{name}`. The tests won't pass without it. 
+In order to make the tests pass, you need to add a method named `%{name}`.
--- a/automated-comments/go/two-fer/strings_join_used_for_concatenation.md
+++ b/automated-comments/go/two-fer/strings_join_used_for_concatenation.md
@@ -1,3 +1,3 @@
 `strings.Join` is only normally used if the data is of type `[]string`.
-In other situations, it is normal to use `fmt.Sprintf`, which is a very powerful way of building up strings from multiple pieces.
+In other situations, it is more common to use `fmt.Sprintf`, which is a very powerful way of building up strings from multiple pieces.
 Could you try updating your solution to use that?
--- a/automated-comments/go/two-fer/strings_join_used_for_concatenation.md
+++ b/automated-comments/go/two-fer/strings_join_used_for_concatenation.md
@@ -1,2 +1,3 @@
-Don't use `strings.Join` if the data is not of type `[]string`.
-Use `fmt.Sprintf` instead. It's a very powerful way of building up strings from multiple pieces.
+`strings.Join` is only normally used if the data is of type `[]string`.
+In other situations, it is normal to use `fmt.Sprintf`, which is a very powerful way of building up strings from multiple pieces.
+Could you try updating your solution to use that?
--- a/automated-comments/javascript/general/no_method.md
+++ b/automated-comments/javascript/general/no_method.md
@@ -1,1 +1,1 @@
-No method called `%{method.name}` was found. The tests won't pass without it.
+In order to make the tests pass, you need to create a method called `%{method.name}`.
--- a/automated-comments/javascript/general/no_named_export.md
+++ b/automated-comments/javascript/general/no_named_export.md
@@ -1,6 +1,6 @@
 The tests won't pass without an [export](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) called `%{export.name}`.
 
-You can export constants or functions like this:
+Constants and functions can be exported like this:
 ```javascript
 // exporting a "const"
 export const %{export.name} = /*...*/
--- a/automated-comments/javascript/general/no_named_export.md
+++ b/automated-comments/javascript/general/no_named_export.md
@@ -1,7 +1,6 @@
-No [export](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export)
-called `%{export.name}` was found. The tests won't pass without it.
+The tests won't pass without an [export](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) called `%{export.name}`.
 
-Ensure `%{export.name}` exists and add `export` in front of it:
+You can export constants or functions like this:
 ```javascript
 // exporting a "const"
 export const %{export.name} = /*...*/
@@ -9,3 +8,5 @@ export const %{export.name} = /*...*/
 // exporting a named function declaration
 export function %{export.name}() { /* ... */ }
 ```
+
+Please ensure `%{export.name}` exists and that it has `export` in front of it.
--- a/automated-comments/javascript/general/no_parameter.md
+++ b/automated-comments/javascript/general/no_parameter.md
@@ -1,2 +1,1 @@
-The function `%{function.name}` requires a parameter.
-The tests won't pass without it.
+In order to make the tests pass, you need to change the function `%{function.name}` to require a parameter.
--- a/automated-comments/javascript/gigasecond/dont_use_date_parse.md
+++ b/automated-comments/javascript/gigasecond/dont_use_date_parse.md
@@ -1,7 +1,3 @@
-Use [`Date#getTime`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime),
-as `Date.parse(%<parameter.name>)` is not a good candidate.
+Your solution used `Date.parse(%<parameter.name>)`. This works, but isn't _intended_ to be used with a `Date` object. In fact, in order to work it first coerces the input value to a `string` and then parses it back into a date, which is unnecessary in this case.
 
-`Date.parse` is supposed to be called with a `string`, and therefore not
-_intended_ to be used like this. The reason it does work is because `Date.parse`
-first coerces the input value to a `string` and then parses it back into a date.
-These extra conversions are unnecessary.
+It would be better to use `%<parameter.name>.getTime()` as [`Date#getTime`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime) doesn't coerce the input but works on the Date directly.


automated-comments/csharp/general/invalid_method_signature.md
```diff
@@ -1 +1 @@
-Change the signature of the `%{name}` method to be `%{signature}`.
+In order to make the tests pass, you need to change the signature of the `%{name}` method to be `%{signature}`.
```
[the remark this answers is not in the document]
ErikSchierboom: I like the suggested change! Using "you" can sometimes be a little like finger-pointing.

automated-comments/csharp/general/missing_method.md
```diff
@@ -1 +1 @@
-Add a method named `%{name}`. The tests won't pass without it. 
+In order to make the tests pass, you need to add a method named `%{name}`.
```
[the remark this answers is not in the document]
ErikSchierboom: 👍

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
```diff
@@ -1,2 +1,3 @@
-Don't use `strings.Join` if the data is not of type `[]string`.
-Use `fmt.Sprintf` instead. It's a very powerful way of building up strings from multiple pieces.
+`strings.Join` is only normally used if the data is of type `[]string`.
+In other situations, it is normal to use `fmt.Sprintf`, which is a very powerful way of building up strings from multiple pieces.
```
ErikSchierboom: Maybe this is me being Dutch, but "normal" felt a bit odd to me. It felt a bit like saying that not doing it is abnormal. Maybe "customary" is better? Feel free to ignore if this really _is_ me being Dutch :D

```suggestion
In other situations, it is customary to use `fmt.Sprintf`, which is a very powerful way of building up strings from multiple pieces.
```

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
```diff
@@ -1,2 +1,3 @@
-Don't use `strings.Join` if the data is not of type `[]string`.
-Use `fmt.Sprintf` instead. It's a very powerful way of building up strings from multiple pieces.
+`strings.Join` is only normally used if the data is of type `[]string`.
```
ErikSchierboom: See comment below about being Dutch, but I like the below:

```suggestion
`strings.Join` is usually only used if the data is of type `[]string`.
```

automated-comments/javascript/general/no_method.md
```diff
@@ -1 +1 @@
-No method called `%{method.name}` was found. The tests won't pass without it.
+In order to make the tests pass, you need to create a method called `%{method.name}`.
```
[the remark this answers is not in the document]
ErikSchierboom: @kotp Any reason not to use the same phrasing a suggested earlier? So something like this:

```suggestion
In order to make the tests pass, there must be a method named `%{name}`.
```

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
```diff
@@ -1,2 +1,3 @@
...
+In other situations, it is normal to use `fmt.Sprintf`, which is a very powerful way of building up strings from multiple pieces.
+Could you try updating your solution to use that?
```
ErikSchierboom: Maybe be explicit about what to use here:

```suggestion
Could you try updating your solution to use `fmt.Sprintf`?
```

@kotp In other comments, you've suggested to remove the "you" part, which I like. Do you have any suggestion how to apply it here?

automated-comments/javascript/general/no_named_export.md
```diff
@@ -1,11 +1,12 @@
-No [export](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export)
-called `%{export.name}` was found. The tests won't pass without it.
+The tests won't pass without an [export](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) called `%{export.name}`.
 
-Ensure `%{export.name}` exists and add `export` in front of it:
+You can export constants or functions like this:
```
ErikSchierboom:
```suggestion
Constants and functions can be exported like this:
```

automated-comments/javascript/general/no_parameter.md
```diff
@@ -1,2 +1 @@
-The function `%{function.name}` requires a parameter.
-The tests won't pass without it.
+In order to make the tests pass, you need to change the function `%{function.name}` to require a parameter.
```
ErikSchierboom:
```suggestion
In order to make the tests pass, the function `%{function.name}` needs to be changed to require a parameter.
```

automated-comments/javascript/gigasecond/dont_use_date_parse.md
```diff
@@ -1,7 +1,3 @@
-Use [`Date#getTime`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime),
-as `Date.parse(%<parameter.name>)` is not a good candidate.
+Your solution used `Date.parse(%<parameter.name>)`. This works, but isn't _intended_ to be used with a `Date` object. In fact, in order to work it first coerces the input value to a `string` and then parses it back into a date, which is unnecessary in this case.
```
ErikSchierboom:
```suggestion
Passing a `Date` as input to `Date.parse(%<parameter.name>)` works, but this function isn't _intended_ to work with `Date` objects. In fact, when `Date.parse` receives a `Date` parameter, it first coerces it to a `string` and then parses it back into a `Date`, which is unnecessary in this case.
```

automated-comments/javascript/gigasecond/dont_use_date_parse.md
```diff
@@ -1,7 +1,3 @@
...
 
-`Date.parse` is supposed to be called with a `string`, and therefore not
-_intended_ to be used like this. The reason it does work is because `Date.parse`
-first coerces the input value to a `string` and then parses it back into a date.
-These extra conversions are unnecessary.
+It would be better to use `%<parameter.name>.getTime()` as [`Date#getTime`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime) doesn't coerce the input but works on the Date directly.
```
ErikSchierboom: I'm not a fan of "It would be better". Maybe we rephrase this a bit?

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
F3PiX: I feel this is a different 'you', and totally okay to use here.

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
F3PiX: 'common' ? 'more common' ?

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
bitfield: I often phrase this as "Would you like to try...?" That makes it clear that this isn't a _correction_, as such; it's an alternative suggestion which may be worth exploring.

(Of course, the answer may be "No, I wouldn't". That's never happened yet, though.)

automated-comments/javascript/gigasecond/dont_use_date_parse.md
bitfield: "This works, but it's more efficient to use... because..."

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
iHiD: It depends on whether it's normal or common! :) If nearly everyone does it, then I'd say normal (e.g. it's an idiom or close to it). If it's common (eg. 60% of people do it) then I'd say "common" but then is it in an idiom and if not, should we not say it's common or normal, but just state it as an alternative?

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
iHiD: > I feel this is a different 'you', and totally okay to use here.

Yes! Agreed.

> Would you like to try..

Agree too.

```suggestion
Would you like to try updating your solution to use that?
```

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
ErikSchierboom: Thanks for the explanation. I'll then leave it to the go-people to determine if it is normal or common :)

tehsphinx: It's not like everybody does it, no. So `common` seems correct according to the definition @iHiD gave.

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
SleeplessByte: > (Of course, the answer may be "No, I wouldn't". That's never happened yet, though.)

On the JS track we get this _quite frequently_, which is why I stopped phrasing it like that.

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
bitfield: Perhaps "more common" is the best compromise here. I often like to say that certain things in Go are "conventional", too—and I mean that in a good way! Gophers love being conventional.

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
bitfield: > How does using that improve the code?

This is a perfectly sensible and nice suggestion. However, students tend to take such questions a bit too seriously, and I don't like the idea of someone thinking too hard about this (remember, this is the very first program they've ever written in Go).

We've talked a lot about the string concatenation issue and how much information and nuance it's wise to load students with on this first exercise. I favour a light touch here; if they use `fmt.Sprintf`, great. If not, I'll let them know that it exists and it's a powerful way to format data into strings. If they're not particularly interested now, they'll learn about it soon enough.

I'd ask if they'd like to try updating their solution to use `fmt.Sprintf`, and leave it at that.

bitfield: For comparison, here's what Exalysis says if the student uses `+=` to concatenate strings. As usual, @tehsphinx strikes a very warm and genial note:

> I see you are using `+` to concatenate the string. That is perfectly legit! If you didn't know about the `fmt` package, check it out. It can do lots of things related to string formatting, including concatenating strings.
>
> (Using `+` is probably the right choice here. I just wanted you to know about `fmt`, because it's awesome, and you'll be using it a lot.)

tehsphinx: I don't think that was entirely me...

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
tehsphinx:
```suggestion
In other situations, it is common to use `fmt.Sprintf`, which is a very powerful way of building up strings from multiple pieces.
```

tehsphinx: I'd go with `common`.

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
tehsphinx:
```suggestion
Can you try updating your solution to use `fmt.Sprintf`?
```

tehsphinx: `Would you like to try...` feels too soft. We are talking about a `disapprove` comment here. I'd even suggest to use `Can` if we want to make it a question.

tehsphinx: > `How does using that improve the code?`

We have such a question in another comment and have lots of students actually answer that. If we want answers, that is a good thing but here I would not try to get actual answers.

automated-comments/javascript/gigasecond/dont_use_date_parse.md
tehsphinx:
```suggestion
It is better to use `%<parameter.name>.getTime()` as [`Date#getTime`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime) doesn't coerce the input but works on the Date directly.
```

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
ErikSchierboom: > This is a perfectly sensible and nice suggestion. However, students tend to take such questions a bit too seriously, and I don't like the idea of someone thinking too hard about this

I feel the same way.

automated-comments/javascript/gigasecond/dont_use_date_parse.md
tehsphinx: Gathered from @kotp suggestion.

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
bitfield: > We are talking about a disapprove comment here.

I think this discussion is gradually bringing us to the conclusion that this _shouldn't_ be a disapprove. It's a 'would you like to try'.

tehsphinx: If that is the case, we have to make it `stern` again because it will not be copied in by a mentor but provided automatically by system. Asking a question would then not be a good idea at all.

bitfield: "Only an idiot would use `strings.Join()` in this situation. Therefore you must be an idiot. Please get a clue and stop wasting my time."

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
iHiD:
```suggestion
In other situations, it is more common to use `fmt.Sprintf`, which is a very powerful way of building up strings from multiple pieces.
```

automated-comments/csharp/general/missing_method.md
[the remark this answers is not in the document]
iHiD: I'm less keen on this change as we're removing the "call to action". I would suggestion maybe "a method needs to be added named ..."?

automated-comments/javascript/general/no_method.md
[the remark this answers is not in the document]
iHiD: Same comment as above. I'd like the call to action to be clearer.

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
iHiD: I feel like being explicit that we expect an action is important as I too have had lots of people say "no" when I phrase things as "would you like to.." and have had essays rather than code back when I ask questions that don't explicitly tell people to submit code.

How about we make it more explicit that we want/expect someone to do something, but still polite (I'm english - it's what we do :))

```suggestion
Could you try updating your solution to use `fmt.Sprintf` please?
```

iHiD: I'm not sure that's neutral "bot" tone enough for an approve comment, @bitfield :)

I think it would be helpful to get away from the word "stern" as that's not the intention. The intention for approval comments is that they're not **personal** - ie they don't feel like they're coming from a person. They should still be empathetic but they should be factual, rather than being a basis for discussion or conversation.

automated-comments/javascript/gigasecond/dont_use_date_parse.md
iHiD: I prefer "would be" I think. "is" feels a lot stronger, "would be" feels a bit more chatty and friendly to me.

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
kotp: It is "more common" than what?  Perhaps "it is common" is better here.

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
kotp: It is "more common" than what? Perhaps "it is common" is better here.

automated-comments/go/two-fer/strings_join_used_for_concatenation.md
bitfield: This has never really made sense to me. In my experience people learn better, and respond better, when the comments you give them are friendly, encouraging, and, yes, personal.

Sure, they're _selected_ by machine, but they're written by humans. Why shouldn't they sound human?

automated-comments/javascript/gigasecond/dont_use_date_parse.md
kotp: I use "would be" in conjunction with "if something is done".  It "is better" because of a specific reason, as given here.

SleeplessByte:
I think these have been applied throughout in other PRs and otherwise should be redone ;)

PR #1265, opened 2019-08-29
make compatible with more tar versions

VaclavSynacek:
`tar -xf something.tgz` works only in some tar versions such as recent GNU version. However `tar -xzf something.tgz` works with more versions of tar even those that do not have gzip autodetection, such as the bsd version.

### DIFF
--- a/walkthrough/index.html
+++ b/walkthrough/index.html
@@ -203,7 +203,7 @@ click on the downloaded file and select _Extract Here_.
 * If you’re using the command line (use the
 right archive file name for your architecture):
 ```
-tar -xf exercism-linux-64bit.tgz
+tar -xzf exercism-linux-64bit.tgz
 ```
 
 ---


SleeplessByte:
cc @exercism/ops

NobbZ:
If I recall correctly, the content of the `walkthrough` folder is generated from the content of another repository, in this particular case its probably this file/section:

https://github.com/exercism/interactive-cli-walkthrough/blob/master/contents/linux/tarball.tw2#L19-L23

kntsoriano:
@VaclavSynacek Thanks for this PR :) However, @NobbZ is correct that this file is generated from another repository. You could send your PR here: https://github.com/exercism/interactive-cli-walkthrough.

SleeplessByte:
cc @exercism/maintainers-admin

PR #1266, opened 2019-08-29
add the BSDs in list of OSes

VaclavSynacek:


SleeplessByte:
Can we do this without the word "just"?

SleeplessByte:
Closed because it should be changed in https://github.com/exercism/interactive-cli-walkthrough first then submitted here.

PR #1267, opened 2019-08-29
Add mentoring notes for the Leap exercise on the Elm track

ceddlyburge:


ceddlyburge:
Hi There, could somebody please take a look at this pull request, its about 6 weeks old now.

I would be happy to review another request in return.

Thanks, Cedd

tracks/elm/exercises/leap/mentoring.md
```diff
@@ -0,0 +1,90 @@
...
+isLeapYear year =
+    let
+        divisibleBy number = 
+            Basics.remainderBy number year == 0         
+    in
+        (
+            divisibleBy 4 && 
+            not (divisibleBy 100)
+        )
+        || divisibleBy 400
```
sshine: Would you not rather have

```elm
divisibleBy 4 && (not (divisibleBy 100) || divisibleBy 400)
```

to avoid checking for divisibility by 400 when you know that it isn't divisible by 4?

tracks/elm/exercises/leap/mentoring.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+The complicated if statement is a bit of a problem, as it is hard to describe in language, and hence talking about the code with others is difficult.
```
sshine: What complicated if-statement?

Do you mean the `&&`/`||`/`not`-based solution above?

tracks/elm/exercises/leap/mentoring.md
```diff
@@ -0,0 +1,90 @@
...
+
+Using a case statement turns the if statement in to a truth table, and is probably a bit clearer, but at the expense of some extra code:
```
sshine: The following snippet is one that I give as feedback on the Haskell track:

> I would categorize your solution strategy as "truth-table-based". I would also argue that such solutions are generally not ideal, because the chunks of `True`s and `False`s can be hard to mentally map back, so you have to move your eyes forth and back between the table and the "table header".
>
> The syntactic alternative closest to truth tables that is easier to read is use of [guards](http://learnyouahaskell.com/syntax-in-functions#guards-guards):
>
> ```haskell
> isLeapYear :: Integer -> Bool
> isLeapYear year
>   | isDivisibleBy 400 = True
>   | isDivisibleBy 100 = False
>   | isDivisibleBy   4 = True
>   | otherwise         = False
>    where
>      isDivisibleBy n = year `rem` n == 0
> ```

So I think this point can at least be debated.

tracks/elm/exercises/leap/mentoring.md
```diff
@@ -0,0 +1,90 @@
...
+			True
+
+		else if divisibleBy 100 then
+			False
+
+		else if divisibleBy 4 then
+			True
+
+		else
+			False
```
sshine: The following snippet is one that I give as feedback on the Haskell track:

> Don't write `if p then True else False`. This is equivalent to simply `p`. Similarly, don't write `if p then False else True`. This is equivalent to simply `not p`. You can combine `p` and `not p` with `&&` and `||` for your values of `p`.

On the other hand, there is one particular if-then-else solution that I do find readable (and not verbose):

```haskell
isLeapYear :: Integer -> Bool
isLeapYear year =
  if divisibleBy 100
  then divisibleBy 400
  else divisibleBy 4
  where
    isDivisibleBy n = year `rem` n == 0
```

tracks/elm/exercises/leap/mentoring.md
```diff
@@ -0,0 +1,90 @@
...
+		...
+	in
+		if is400YearSpecialCase then
+			True
+
+		else if is100YearSpecialCase then
+			False
+
+		else 
+			defaultCase
```
sshine: Correct me if I'm wrong: I presume the point of this example is to show that you can name things very particularly to add meaning to some nested expression. If this is the case, I agree very much with the sentiment, which I think is also embedded in recommending a helper function such as `divisibleBy`.

But I think this particular example is not a good one, since `is400YearSpecialCase` doesn't say as much as `divisibleBy 400`. It hides abstraction in a place where there is so little need for it: While I might not care how divisibility is decided, I do care that when determining leap years, the recipe operates at the abstraction level of determining divisibility.

tracks/elm/exercises/leap/mentoring.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+### Common suggestions
+
+Discussing the various options around the describability of the solution is worthwhile. The calculation of leap years is not going to change any time soon, so there is definitely an argument that the complicated conditional statement is ok. However most code deals with business logic, which changes much more frequently, and so the trade off's are different.
```
sshine: I'm not sure if the first example, `divisibleBy 4 && (not (divisibleBy 100) || divisibleBy 400)`, is the complicated conditional statement. I don't think it is more complicated than the alternatives.

tracks/elm/exercises/leap/mentoring.md
ceddlyburge: Hi Simon, thanks for the review. This is a defendable point, I've made the change

tracks/elm/exercises/leap/mentoring.md
ceddlyburge: I do mean the `divisibleBy 4 && (not (divisibleBy 100) || divisibleBy 400)` statement.

One of my main points is that this is very hard to describe in words (to a product owner or a business analyst or another programmer or whatever).

tracks/elm/exercises/leap/mentoring.md
ceddlyburge: Hi Simon. Elm doesn't support the guard syntax, so that isn't a possibility.

I would say that the "truth-table" is clearer, easier to describe in words and takes less mental mapping effort than the conditional statement. It isn't my favourite solution though.

However, even though I think the mental mapping is easy, there is quite a lot of it to do, and you make a good point about moving your eyes back and forth.

I have incorporated your comments.

tracks/elm/exercises/leap/mentoring.md
ceddlyburge: Hi Simon. I see your point here, but I'm afraid I don't agree with it. I am happy to have a little extra code to improve readability / clarity. Maybe there is a difference between the average elm (a beginner / easy to learn functional language) programmer and the average Haskell (a much more complete functional language) programmer.

I like the clarity of `if divisibleBy 400 then True`, and it is easy to describe in words. It takes me longer to work out what actually happens when reading your example.

So all in all I am afraid that I am not going to action this point.

tracks/elm/exercises/leap/mentoring.md
ceddlyburge: Hi Simon `divisibleBy 4 && (not (divisibleBy 100) || divisibleBy 400)` is the complicated conditional statement I am talking about. I concede that it isn't very complicated, but I think it still illustrates the point. And if you try describing it to somebody using words I think you will find it difficult.

tracks/elm/exercises/leap/mentoring.md
ceddlyburge: Hi Simon, I agree with this, the point was more of to illustrate concepts. I've deleted this section as it obviously causes confusion.

ceddlyburge:
Hi Simon

I've addressed (or, ahem, ignored) all of your comments now, so I would like to get this merged.

You said it was Ok not to action all of your comments, but in the cases where we think differently, another option would be to add both points of view in to the notes. Presumably if we have different opinions about it then other people will as well. Please let me know.

Thanks, Cedd

sshine:
Hi Cedd

Yes, it would have been OK to ignore all of my comments. I will incorporate that statement the next time I review mentor notes for a track I'm not responsible for. The comments were merely meant as sharing between tracks, not as requirements.

Having opinionated mentor notes is better than not having them, since they motivate more mentors to update them and have discussions that sync our collective knowledge, which in turns improve the overall experience of being mentored.

The mentor notes are, after all, just suggestions to what a mentor might focus on. In practice we're really granting mentors the right to say anything they like as long as it agrees with our [code of conduct](https://exercism.io/code-of-conduct). I don't even recall if there's a "report mentor" button.

ceddlyburge:
Hi Simon, thanks for this, I appreciate it.

Cheers, Cedd

PR #1272, opened 2019-08-31
Explain  in cli-walkthrough where to find token.

markjuh:
I was going through the interactive walkthrough, but I could not figure out what token to use until I stumbled upon https://exercism.io/cli.

### DIFF
--- a/walkthrough/index.html
+++ b/walkthrough/index.html
@@ -33,6 +33,8 @@ In order to configure the CLI, paste in the following text into your terminal:
 [CONFIGURE_COMMAND]
 ```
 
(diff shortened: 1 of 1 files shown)


rpottsoh:
@markjuh thanks for the suggestion.  This looks like a potentially helpful change.  The main issue at the moment is that this file is built and not maintained by hand.  [configuring_the_cli.tw2](https://github.com/exercism/interactive-cli-walkthrough/blob/master/contents/common/configuring_the_cli.tw2) is the file ASCII file you should be targeting to change.  I am going to close this PR, please open another in exercism/interactive-cli-walkthrough.  The maintainers there will take it from there.

PR #1280, opened 2019-09-02
Ruby: Short note on DRY - High scores

programatologist:
DRY - Do not Repeat Yourself concept in Ruby.


kotp:
Any reason this is closed @programatologist ?

programatologist:
@kotp 
After while of creating this PR, I have got unsure that this is a right recommendation, thought recommend after a bit more check!

kotp:
So perhaps unsure if DRY isn't a more general principle that probably gets mentioned, rather than something specific to this exercise?

kotp:
I do mention that aspect when mentoring this exercise.  But not because of DRY, but because that set is smaller.

PR #1282, opened 2019-09-03
Add Python mentor bio - Simona Ziauke

SimaSk:


SimaSk:
Invalid JSON

NobbZ:
No need to close the PR because of that, just repair the JSON in another commit. We would have squashed the commit then.

PR #1285, opened 2019-09-03
disable labeler temporarily

NobbZ:
Disable the workflow until a solution to actions/labeler#12 has been found.

After the linked issue has been solved, we probably need to also bump the version of the action.


NobbZ:
Please squash commits into a single on.

PR #1288, opened 2019-09-05
Initial version of mentor notes for Ruby/Tournament

tenebrousedge:
I would like to propose this code and accompanying text as an initial version of the mentor notes for the Ruby/Tournament problem. Please let me know of any errors in either the code or the copy, or if any element is excessive or lacking, or if the whole thing is bad and I should feel bad for writing it. In particular, let me know if there should be a fully coded example of the version of the solution which uses class properties to store the match results, and whether it's fair to describe that as a "naive" approach.

Thank you for your attention to this matter.


SleeplessByte (approved):
Merging because having something is better than nothing.

PR #1289, opened 2019-09-07
Rust: Bob - give a reason for using constants

efx:
This relates to the slack [thread](https://exercism-team.slack.com/archives/CASLGD18D/p1567640988021800) where I asked about why we don't give *whys* for some recommendations to improving solution. As I mentor solutions I'll open these smaller PRs to fill out the notes.

### DIFF
--- a/tracks/rust/exercises/bob/mentoring.md
+++ b/tracks/rust/exercises/bob/mentoring.md
@@ -62,6 +62,7 @@ If they don't use a constant:
 ```
 We can improve this by setting our output strings to be [constants](https://doc.rust-lang.org/std/keyword.const.html)
 ```
+*Why?* Constants help the compiler optimize and improve readability of your program.
 
 If they don't create a trimmed variable or use something other than `.trim()`:
 ```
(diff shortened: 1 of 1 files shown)


SleeplessByte:
cc @TheDarkula

efx:
@TheDarkula any objections here?

efx:
@Aloso and I actually had this conversation and, in this case, the compiler will optimize either way. So based on my limited understanding I don't think using a `const` is for performance here. I don't  agree then with giving performance as the primary justification.
Any thoughts Aloso? I can also pose a discussion on the rust users discourse.

efx:
Given we have demoted Bob I will probably end up closing this PR too.

efx:
Forgive my naivete here as well. I am striving to reconcile the conflict for personal sanity and to mentor better. I plugged in an example to [godbolt.org](https://godbolt.org/z/8WGqR-) and the machine code output does not look different when using a `const` or plain `i64`. Note this is based on the gigasecond example so may differ from strings.

PR #1301, opened 2019-09-16
Add mentor bio for bergjohan

bergjohan:


rpottsoh:
@iHiD `WIP` is getting in the way here.  I know you said you got rid of it but the PRs here are acting like it is....  It is preventing PRs from being merged.

iHiD:
@rpottsoh Interesting. It's been uninstalled, but that doesn't remove it from the branch checks. Useful to know - we'll probably see this on other tracks too. Should be fixed now :)

rpottsoh:
Ah, much better.  Thanks @iHiD

PR #1304, opened 2019-09-18
Ruby: Two Fer - No function named

kotp:
In Ruby, the closest thing we refer to as functions are Proc and lambda.
We want to indicate that there is no method named found.

ref: exercism/ruby-analyzer#16


iHiD:
Thanks!

PR #1309, opened 2019-09-21
Add Raoul as mentor

rrrlw:
First commit had an error, should be fixed in second commit.


mentors/r/r.json
```diff
@@ -1 +1,10 @@
-[]
+[
+  {
+  "github_username": "rrrlw",
```
NobbZ: Can you please indent fields another two spaces?

rrrlw: Fixed in commit 3617c2b

PR #1313, opened 2019-09-22
Rename Independent Mode to Practice Mode and explain oversubscribed tracks

iHiD:
The text is (hopefully!) self-explanatory.

This should be merged when https://github.com/exercism/website/pull/727 is merged.

### DIFF
--- a/pages/mentored_mode_vs_independent_mode.md
+++ b/pages/mentored_mode_vs_independent_mode.md
@@ -20,4 +20,4 @@ While we recommend sticking to one mode per track, it is possible to change if y
 
 ### Oversubscribed tracks
 
-Some tracks are oversubscribed, meaning here are two many students per mentor. In this situation, anyone starting the track is forced to start in Practice Mode, and no-one in Practice Mode will be able to request mentor help until the track is back in a "healthy" state. At the time of writing, we define an oversubscribed track as one that has a median wait time of over one week.
+Some tracks are oversubscribed, meaning there are too many students per mentor. In this situation, anyone starting the track is forced to start in Practice Mode, and no-one in Practice Mode will be able to request mentor help until the track is back in a "healthy" state. At the time of writing, we define an oversubscribed track as one that has a median wait time of over one week.
--- a/pages/mentored_mode_vs_independent_mode.md
+++ b/pages/mentored_mode_vs_independent_mode.md
@@ -1,6 +1,6 @@
 ### Introduction
 
-Tracks can be completed in two modes on Exercism - Mentored Mode and Independent Mode. You can choose a different mode for each track.
+Tracks can be completed in two modes on Exercism - Mentored Mode and Practice Mode. You can choose a different mode for each track.
 
 ### Mentored Mode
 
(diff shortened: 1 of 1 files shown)


pages/mentored_mode_vs_independent_mode.md
```diff
@@ -10,10 +10,14 @@ Every time you complete a solution, you'll also unlock extra bonus exercises. Th
...
 
 ### Changing mode on a track
 
-While we recommend sticking to one mode per track, it is possible to change if you decide you've made the wrong decision. You will find the button in the "Danger Zone" section of a track page, which is on the right-hand-side in Mentored Mode and the bottom of the page in Independent Mode.
+While we recommend sticking to one mode per track, it is possible to change if you decide you've made the wrong decision. You will find the button in the "Danger Zone" section of a track page, which is on the right-hand-side in Mentored Mode and the bottom of the page in Practice Mode.
+
+### Oversubscribed tracks
+
+Some tracks are oversubscribed, meaning here are two many students per mentor. In this situation, anyone starting the track is forced to start in Practice Mode, and no-one in Practice Mode will be able to request mentor help until the track is back in a "healthy" state. At the time of writing, we define an oversubscribed track as one that has a median wait time of over one week.
```
jordanspangenberg:
```suggestion
Some tracks are oversubscribed, meaning there are too many students per mentor. In this situation, anyone starting the track is forced to start in Practice Mode, and no-one in Practice Mode will be able to request mentor help until the track is back in a "healthy" state. At the time of writing, we define an oversubscribed track as one that has a median wait time of over one week.
```

iHiD: @jordanspangenberg Thanks!

pages/mentored_mode_vs_independent_mode.md
```diff
@@ -10,10 +10,14 @@ Every time you complete a solution, you'll also unlock extra bonus exercises. Th
...
+We designed Practice Mode for people who want to take fast, deep-dives into a language without waiting for mentor feedback. In Practice Mode you can take the exercises in whatever order you want and there is no progression or unlocking. You may request mentoring on one exercise at a time, but your solutions will have a lower priority than people in Mentored Mode.
 
 ### Changing mode on a track
 
-While we recommend sticking to one mode per track, it is possible to change if you decide you've made the wrong decision. You will find the button in the "Danger Zone" section of a track page, which is on the right-hand-side in Mentored Mode and the bottom of the page in Independent Mode.
+While we recommend sticking to one mode per track, it is possible to change if you decide you've made the wrong decision. You will find the button in the "Danger Zone" section of a track page, which is on the right-hand-side in Mentored Mode and the bottom of the page in Practice Mode.
```
kotp: As long as the "Danger Zone" remains well defined, should we state where on the page it is?  If the style changes, or even if the layout changes, we would have to remember to change this location.  And since some folks have been doing custom styles (stylish, anyone?) I don't think we can guarantee this is accurate.

iHiD: This PR doesn't introduce or change the content of the Danger Mode bit so I'd rather open a separate discussion for that, rather than it blocking this one.

kotp: I guess that is the problem with not having "one sentence per line" commits.  I saw green and didn't look much closer.

PR #1320, opened 2019-09-25
python: Remove unused import & fix typo

cspollar:
Variable was declared as a string when it should have been a list.


SleeplessByte:
Thank you @cmccandless for the review and thank you @cspollar for the content!

PR #1326, opened 2019-09-28
Terminology change from independent mode

kotp:
Now we call indepdent mode practice mode.

PR #1330, opened 2019-09-29
Fix incorrect edit to Changelog.

iHiD:
#1326 changed this file, but incorrectly renamed "Mentored Mode" to "Practice Mode". This fixes this and clarifies the wording.


kotp:
> I think the capitalization is on purpose right?

I would think so.

kotp:
Thanks for catching that. @iHiD

iHiD:
No probs! Thanks both.

PR #1331, opened 2019-09-29
Update 'sshine' presence on tracks

sshine:
- OCaml: Remove myself as mentor:
  I have not given OCaml feedback in months, so having me in this list would be dishonest to students.
 - SML: Update mentor bio.
 - Haskell: Update mentor bio.


SleeplessByte:
Yeah, should have been 3 PRs, but alas 🤷‍♂

PR #1337, opened 2019-10-01
Add mentoring notes for elixir/bank-account

### LINKED ISSUE #1340: elixir/bank-account: mentoring notes

filalex77:
Closes #1340.

### DIFF
--- a/tracks/elixir/exercises/bank-account/mentoring.md
+++ b/tracks/elixir/exercises/bank-account/mentoring.md
@@ -1,0 +1,142 @@
+### Reasonable solutions
+
+Using [`Agent`](https://hexdocs.pm/elixir/Agent.html):
+
+```elixir
+defmodule BankAccount do
+  use Agent
+
+  @opaque account :: pid
+
+  @spec open_bank() :: account
+  def open_bank() do
+    {:ok, account} = Agent.start_link(fn -> 0 end)
+    account
+  end
+
+  @spec close_bank(account) :: none
+  def close_bank(account), do: Agent.stop(account)
+
+  @spec balance(account) :: integer
+  def balance(account) do
+    if Process.alive?(account) do
+      Agent.get(account, & &1)
+    else
+      {:error, :account_closed}
+    end
+  end
+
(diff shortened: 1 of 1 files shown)


filalex77:
@neenjaw I came up with this; can you mentor me? 😺 

```elixir
defmodule BankAccount do
  @opaque account :: pid

  defp listen(from, balance) do
    new_balance =
      receive do
        :balance ->
          send(from, balance)
          balance

        {:update, amount} ->
          balance + amount
      end

    listen(from, new_balance)
  end

  @spec open_bank() :: account
  def open_bank(), do: spawn(fn -> listen(self(), 0) end)

  @spec close_bank(account) :: true
  def close_bank(account), do: Process.exit(account, :kill)

  @spec balance(account) :: integer
  def balance(account) do
    if Process.alive?(account) do
      send(account, :balance)

      receive do
        balance -> balance
      end
    else
      {:error, :account_closed}
    end
  end

  @spec update(account, integer) :: any
  def update(account, amount) do
    if Process.alive?(account) do
      send(account, {:update, amount})
    else
      {:error, :account_closed}
    end
  end
end
```

neenjaw:
I think your example using spawn, send, receive is quite nice and people can see how these work with a quick read through.

This exercise is tricky sometimes because people often make it complicated when the goal of this exercise is to show how messaging and concurrency works in elixir.

I think you should incorporate this. 

I'd like @chriseyre2000 to take a peek at this since when I did this exercise, he suggested I look for the variants incase there are features he thinks would be good to include.

chriseyre2000:
My previous entry was rejected as being too long.

There is a fourth solution that I have seen:

https://exercism.io/tracks/elixir/exercises/bank-account/solutions/1aee1d83a73944109bde989399a6861d

neenjaw:
@SleeplessByte - out of curiosity, how does this PR get merged? It would be nice to have for mentors

SleeplessByte:
@neenjaw I would expect someone with decent elixer knowledge to approve this, but it seems like that's you so here we go!

PR #1350, opened 2019-10-12
More efficient and straightforward reversing of Dict

rirze:
Instead of iterating over every character in the dictionary value, simply split on the comma and parse that instead.


SleeplessByte:
Thank you @cmccandless and @rirze

PR #1356, opened 2019-10-22
python/scrabble-score: Update mentoring.md

feature-engineer:
1. Using a dictionary for mapping score to letters only for it to be inverted, isn't necessary, and makes it less similar to the original table. Using tuples makes it both more readable and more efficient.

2. No need to indicate global and compute the dictionary once the function gets called, just use a constant outside the function. 

3. Name constants in all upper case, as per PEP 8.

4. Use a generator in sum, not a list.

### DIFF
--- a/tracks/python/exercises/scrabble-score/mentoring.md
+++ b/tracks/python/exercises/scrabble-score/mentoring.md
@@ -10,7 +10,7 @@ dictionary literal called LATTER_VALUE, and using it. Using it is quite simple:
 
 ```python
 def score(word: str) -> int:
-    return sum( [ LATTER_VALUE[ch] for ch in word.upper() ] )
+    return sum( [ LETTER_VALUE[ch] for ch in word.upper() ] )
 ```
 
 You will see solutions that need to traverse
--- a/tracks/python/exercises/scrabble-score/mentoring.md
+++ b/tracks/python/exercises/scrabble-score/mentoring.md
@@ -39,63 +39,41 @@ This is less prone to error, and the editing is straightforward,
 and illustrates an important idea.
 
 ```python
-numbers_to_letters = {
-    1 : 'A, E, I, O, U, L, N, R, S, T',
-    2 : 'D, G',
-    3 : 'B, C, M, P',
-    4 : 'F, H, V, W, Y',
-    5 : 'K',
-    8 : 'J, X',
-    10: 'Q, Z'
-    }
-
-def rev(d: dict) -> dict:
-    return { ch:val for val in d for ch in d[val].split(', ') }
+LETTERS_AND_SCORES = (
+#    Letter                           Value
+    ('A, E, I, O, U, L, N, R, S, T',    1),
+    ('D, G',                            2),
+    ('B, C, M, P',                      3),
+    ('F, H, V, W, Y',                   4),
+    ('K',                               5),
+    ('J, X',                            8),
+    ('Q, Z',                            10)
+)
+    
+LETTERS_TO_SCORES = { 
+    ch: score for letters, score in LETTERS_AND_SCORES 
+                    for ch in letters.split(', ') 
+}
 
 def score(word: str) -> int:
-    letter_value = rev(numbers_to_letters)
-    return sum( [ letter_value[ch] for ch in word.upper() ] )
+    return sum( LETTERS_TO_SCORES[ch] for ch in word.upper()  )
 ```
 
 ### Common Suggestions
 
 This is a good place to introduce a number of ideas:
-Comprehensions, the use a dictionary to rapidly 
-map between letters and values, and inverting a dictionary. 
+Comprehensions (both dictionary, and generator given to the sum), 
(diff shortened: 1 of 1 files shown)


SleeplessByte:
Are the changes you propose _more idiomatic_ or are they just preferences? If it's the latter, don't remove the old solutions/suggestions.

feature-engineer:
@cmccandless If you don't use the const convention for the dictionary by using all lower case, you'd get linters complaining about not using the const convention, for example pylint would issue the following:

>  C0103: Constant name "letters_to_scores" doesn't conform to UPPER_CASE naming style (invalid-name)

Constants aren't necessarily immutable, they're just variables which you define only once, and can't overwrite. It would have been better to have an immutable const, but I think that introducing MappingProxyType is too much for this exercise. 

Was there any other change you wanted to have?

SleeplessByte:
> Constants aren't necessarily immutable, they're just variables which you define only once, and don't want to overwrite. It would have been better to have an immutable const, but I think that introducing MappingProxyType is too much for this exercise.

In this case, you can show _intent_ by using `UPPER_CASE`.

cmccandless:
OK, I'm fine with keeping the changes to UPPER_CASE.

To summarize my suggestions:
### KEEP
- changes to make global variables UPPER_CASE
- conversion to iterator inside `sum()`
### REVERT
- changes to dictionary structure. The example here is written in a particular way  to demonstrate certain concepts (see the text above the code example), and for that reason should not be altered significantly.

yawpitch:
The intent of both the globals is to create a constant, so UPPER_CASE is correct in this case, though @feature-engineer it’s important to note that Pylint can often false positive on globals that are actually variable due to mutations elsewhere.

Personally, though, I’d tend to prefer a simple dict literal to this “tuple of tuples and reverse” work. There needs be _no_ runtime calculation cost to importing this module, and an appropriately sorted dict is no harder to maintain.

Also, said dict need not even include entries for any of the 1 point characters, so it’s not a significant number of lines to create.

feature-engineer:
@yawpitch I'm not sure I understand how Enum would be used here.

yawpitch:
For ease of `sum` it would be the IntEnum.

```
from enum import IntEnum

class Letters(IntEnum):
    A = E = I = O = U = L = N = R = S = T = 1
    D = G = 2
    B = C = M = P = 3
    F = H = V = W = Y = 4
    K = 5
    J = X = 8
    Q = Z = 10
    
def score(word: str) -> int:
     return sum(Letters[c] for c in word.upper())
```

feature-engineer:
I like @yawpitch 's suggested solution, it's better than the dictionary inversion, or dictionary literal.

Is there any objection to replace the dictionary inversion completely with the Enum approach? I don't see any reason to keep the dictionary inversion.

tracks/python/exercises/scrabble-score/mentoring.md
```diff
@@ -5,12 +5,12 @@ In Scrabble, rare letters are worth more than common letters.
...
 	
-The basic solution has two parts: building a dictionary called
-letter_value here, and using it. Using it is quite simple:
+The basic solution has two parts: defining a global dictionary using a 
+dictionary literal called LATTER_VALUE, and using it. Using it is quite simple:
 
 ```python
 def score(word: str) -> int:
-    return sum( [ letter_value[ch] for ch in word.upper() ] )
+    return sum( [ LATTER_VALUE[ch] for ch in word.upper() ] )
```
cmccandless:
```suggestion
    return sum( [ LETTER_VALUE[ch] for ch in word.upper() ] )
```

tracks/python/exercises/scrabble-score/mentoring.md
```diff
@@ -5,12 +5,12 @@ In Scrabble, rare letters are worth more than common letters.
 
 ### Reasonable solutions
 	
-The basic solution has two parts: building a dictionary called
-letter_value here, and using it. Using it is quite simple:
+The basic solution has two parts: defining a global dictionary using a 
+dictionary literal called LATTER_VALUE, and using it. Using it is quite simple:
```
cmccandless:
```suggestion
dictionary literal called LETTER_VALUE, and using it. Using it is quite simple:
```

tracks/python/exercises/scrabble-score/mentoring.md
```diff
@@ -38,64 +38,89 @@ to get a mapping from letters to their value in the game.
...
     3 : 'B, C, M, P',
     4 : 'F, H, V, W, Y',
     5 : 'K',
     8 : 'J, X',
     10: 'Q, Z'
-    }
+}
+    
+LETTERS_TO_SCORES = { 
+    ch: score for score, letters  in LETTERS_AND_SCORES
```
cmccandless: Iterating through a `dict` object returns an iterator of the keys only.

```suggestion
    ch: score for score, letters in LETTERS_AND_SCORES.items()
```

PR #1358, opened 2019-10-23
javascript/robot-name: Clarify "unpredictable" in problem requirements.

mikesamuel:
This clarifies something that sets off alarms for JavaScript security folk based on [well known attacks](https://blog.securityevaluators.com/hacking-the-javascript-lottery-80cc437e3b7f#.mwvrmr6aq)

> So for Firefox the script needs to simulate seven calls to Math.random() before generating the winning lottery numbers and in Chrome the script needs to simulate nine.

It also adds `crypto.getRandomValues()` as an acceptable variant.

### DIFF
--- a/tracks/javascript/exercises/robot-name/mentoring.md
+++ b/tracks/javascript/exercises/robot-name/mentoring.md
@@ -55,9 +55,14 @@ export class Robot {
 ```
 
 Variations include:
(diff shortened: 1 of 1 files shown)


SleeplessByte:
Whilst I like this change, we should also make it in the README.md. Tagging myself here to make sure I recall to do that.

PR #1369, opened 2019-11-04
Add bisect to talking points

feature-engineer:
People learning python should be familiar with python's standard library, this functionality is already included in python in the form of bisect.bisect_left with minimal wrapping for non existing values.

### DIFF
--- a/tracks/python/exercises/binary-search/mentoring.md
+++ b/tracks/python/exercises/binary-search/mentoring.md
@@ -67,4 +67,18 @@ def binary_search(list_of_numbers, number):
 ### Talking points
 - Using a single variable to track where the midpoint is on each iteration is not enough, because you also need to track how long the current sub-list is.
 - You can represent the high end of the range as the last index where the value could be (inclusive), or one past that (exclusive).  The exclusive approach seems more natural in Python, because that's the way slicing works.
+- There's a builtin binary seach functionality in python called bisect:
+
+```python
+import bisect
+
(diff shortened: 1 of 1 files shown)


feature-engineer:
@cmccandless That is why it is in talking points, and not in acceptable solutions. It's not a solution to this problem, but should be mentioned, as it is the correct and better solution to this problem outside the exercise.

tracks/python/exercises/binary-search/mentoring.md
```diff
@@ -67,4 +67,18 @@ def binary_search(list_of_numbers, number):
...
+import bisect
+
+
+def binary_search(list_of_numbers, number):
+    idx = bisect.bisect_left(list_of_numbers, number)
+    if idx == len(list_of_numbers):
+        raise ValueError("number not found")
+    if list_of_numbers[idx] == number:
+        return idx
+    raise ValueError("number not found")
```
cmccandless: Why not follow the example in the [docs](https://docs.python.org/3/library/bisect.html)?
```
if idx != len(list_of_numbers) and list_of_numbers[idx] == number:
	return idx
raise ValueError("number not found")
```

feature-engineer: That's a good suggestion, done.

SleeplessByte:
Thank you @cmccandless for the review and vigilance. @feature-engineer great addition!

PR #1372, opened 2019-11-06
kotlin/secret-handshake: Improve reasonable solution

eparovyshnaya:
The same logic is implemented 
 - concisely 
 - idiomatically 
 - without redundant internal data structure 
 - without dependency of `Signal` enumeration structure 

Besides both bitwise operators, mandatory for the task, 
the following highly demanded idiomatic Kotlin featured are used: 
 - [single-expression function](https://kotlinlang.org/docs/reference/idioms.html#single-expression-functions)
 - [`let`](https://kotlinlang.org/docs/reference/scope-functions.html#let) scope function
 - return type inference
 - [extension function](https://kotlinlang.org/docs/reference/extensions.html#extension-functions)

Additionaly [infix function](https://kotlinlang.org/docs/reference/functions.html#infix-notation) is exploited.

### DIFF
--- a/tracks/kotlin/exercises/secret-handshake/mentoring.md
+++ b/tracks/kotlin/exercises/secret-handshake/mentoring.md
@@ -1,25 +1,12 @@
 ### Reasonable Solutions
 ```kotlin
 object HandshakeCalculator {
-    fun calculateHandshake(input: Int): List<Signal> {
-        val signalMap = mapOf(
-                Signal.WINK to 1,
-                Signal.DOUBLE_BLINK to 2,
-                Signal.CLOSE_YOUR_EYES to 4,
-                Signal.JUMP to 8
-        )
+    fun calculateHandshake(n: Int) =
+        signals
+            .filterIndexed { i, _ -> n hasBit i }
+            .let { if (n hasBit signals.size) it.reversed() else it }
 
-        val signalsPresent = signalMap.filter(isPresentIn(input)).keys.toList()
-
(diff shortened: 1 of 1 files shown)


eparovyshnaya:
> I like it! Just for my information, what did you choose to `infix` the last function?

It appears to be more readable and even a bit witty as both bitwise operations used there are infix. 
I do not insist on the part, and can explain this textually after the solution itself. 

But `infix notion looks elegant and very simple, so if a student appears to be ready for some extra tools - here it is.

SleeplessByte:
Great work @eparovyshnaya . If you're up for it, perhaps it's time to add some:

```markdown
### Common suggestions

### Talking points
```

eparovyshnaya:
Gladly! There are several tasks where I have some wroughts. Can you please explain `Talking points`? What things should appear there?

SleeplessByte:
So in my opinion it's "things worth discussing". Usually these are idiomatic _options_ instead of specific suggestions. Example:

- single-expression function: talking point, not suggestion. It's a style change. Why would you use it/not use it?
- let scope function: suggestion! In the _official guide_ , use `let` with `if` clause if you want to conditionally change return value. Use `also` if you want a side-effect.
- return type inference: suggestion! This is mentioned in the _official guide_.
- extension function: talking point, not suggestion. It's not _necessary_ or _more_ idiomatic, but super nice to use. Pros/cons.

eparovyshnaya:
Sackful of thanks for the explanation!

PR #1373, opened 2019-11-08
Update mentoring notes for elixir/roman-numerals

angelikatyborska:
Mentioning that maps' keys are not ordered. I encountered this mistake in three solutions in a row just now.

### DIFF
--- a/tracks/elixir/exercises/roman-numerals/mentoring.md
+++ b/tracks/elixir/exercises/roman-numerals/mentoring.md
@@ -124,7 +124,7 @@ While map keys most of the time seems to have a stable order, in reality they do
 >
 > Key-value pairs in a map do not follow any order (that's why the printed map in the example above has a different order than the map that was created).
 
-To get rid of the problem, the student should use an ordered data structure, like a list of tuples (see first reasonable solution), or sort the keys explicitly:
+To get rid of the problem, the student should use an ordered data structure, like a list of tuples (see first reasonable solution), or sort the keys explicitly (but make sure it only happens once):
 ```elixir
 Map.keys(@integer_to_roman)
 |> Enum.sort(&>=/2)
--- a/tracks/elixir/exercises/roman-numerals/mentoring.md
+++ b/tracks/elixir/exercises/roman-numerals/mentoring.md
@@ -127,5 +127,5 @@ While map keys most of the time seems to have a stable order, in reality they do
 To get rid of the problem, the student should use an ordered data structure, like a list of tuples (see first reasonable solution), or sort the keys explicitly:
 ```elixir
 Map.keys(@integer_to_roman)
-|> Enum.sort_by(& -1 * &1)
+|> Enum.sort(&>=/2)
 ```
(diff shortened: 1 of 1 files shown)


tracks/elixir/exercises/roman-numerals/mentoring.md
```diff
@@ -82,3 +82,51 @@ def numerals(number, acc \\ "") do
...
+While map keys most of the time seems to have a stable order, in reality they don't. The student could be pointed to the [documentation of the Map module](https://hexdocs.pm/elixir/Map.html) that state:
+
+> ```elixir
+> %{"one" => :two, 3 => "four"}
+> %{3 => "four", "one" => :two}
+> ```
+>
+> Key-value pairs in a map do not follow any order (that's why the printed map in the example above has a different order than the map that was created).
+
+To get rid of the problem, the student should sort the keys explicitly:
```
NobbZ: To get rid of the problem, students should use an ordered container from the beginning, a list of tuples suites very well.

angelikatyborska: Yeah, good point! I edited the notes to mention both using a list and sorting keys as a possible fix. I think it depends on context with fix should be proposed. If the code is otherwise fine, suggesting to completely change the data structure will work, but if the student is visibly struggling with other things as well, and already relies on accessing the map via `@map[key]`, explicitly sorting the keys is a simpler fix for them to understand. 

Or areee there other harms of using a map for this exercise that I am missing?

NobbZ: A map is an unordered container of keys and values. And using `Access` to get values out of it already shows that one missed better opportunities to solve the exercise by a `Enum.reduce` over the list of tuples or recursion over the number and the list or some metaprogramming over the list.

Though you are right, if a student is struggling with other things as well and there is a lot of incremental rewriting necessary, then its an acceptable intermediary step, and could even be seen in an accepted solution.

tracks/elixir/exercises/roman-numerals/mentoring.md
```diff
@@ -82,3 +82,50 @@ def numerals(number, acc \\ "") do
...
+> %{"one" => :two, 3 => "four"}
+> %{3 => "four", "one" => :two}
+> ```
+>
+> Key-value pairs in a map do not follow any order (that's why the printed map in the example above has a different order than the map that was created).
+
+To get rid of the problem, the student should use an ordered data structure, like a list of tuples (see first reasonable solution), or sort the keys explicitly:
+```elixir
+Map.keys(@integer_to_roman)
+|> Enum.sort_by(& -1 * &1)
```
NobbZ: Only seeing it now, but I think `Enum.sort/2` is better suited here…

```suggestion
|> Enum.sort(&>=/2)
```

tracks/elixir/exercises/roman-numerals/mentoring.md
NobbZ: Oh, and very import, sorting the keys is only acceptable, if its done only once!

angelikatyborska: Indeed! I added that to the notes as well.

PR #1381, opened 2019-11-14
Update mentoring notes for secret handshake

MaksimDanilau:
I find this way a little bit archaic, but from my perspective, it's good to know all the possibilities of java.


MaksimDanilau:
Have an another idea.

PR #1406, opened 2019-12-10
Update l.json

logankilpatrick:
Added my mentor info.

### DIFF
--- a/mentors/julia/l.json
+++ b/mentors/julia/l.json
@@ -1,11 +1,10 @@
 [
   {
-  "github_username": "logankilpatrick",
-  "name": "Logan Kilpatrick",
-  "link_text": null,
-  "link_url": "https://github.com/logankilpatrick",
-  "avatar_url": null,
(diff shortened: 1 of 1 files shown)


mentors/julia/l.json
```diff
@@ -1 +1,11 @@
-[]
+[
+  {
+  "github_username": "logankilpatrick",
+  "name": "Logan Kilpatrick",
+  "link_text": null,
+  "link_url": "https://github.com/logankilpatrick",
+  "avatar_url": null,
+  "bio": "I started using Julia when I was getting started as a Software Engineering Intern at NASA. Since then, I have gone on to Intern with Julia Computing on the PumasAI project, win a new Julia Language contributor award for 2019, and authored a few small packages. The Julia programming language and community are incredible. I am looking forward to helping more in the future."
+}
```
rpottsoh: Each of the lines I have indicated should be indented 2 spaces each.  Otherwise, everything else with the entry is fine.

logankilpatrick: There are already 2 spaces there at the start of lines 2-8. do you mean that line 9 needs 2 spaces or that they all need 2 more spaces?

rpottsoh: Sorry, I wasn't clear.  Two more spaces each.  [] Should not be indented.  {} Should be indented twice and the lines in between a total of 4 spaces each.

logankilpatrick: No worries. Should be good now.

PR #1408, opened 2019-12-13
javascript/gigasecond: Correctly escape link

SleeplessByte:
Percent (`%`) characters need to be espaced in these files so that ruby's formatter doesn't break.


iHiD:
Thanks 💙

Issue #1416, opened 2019-12-18
Travis CI Issue: "eval bundle install --jobs=3 --retry=3 --deployment "


rpottsoh:
It appears Travis CI has been having issues for a few days now.

rpottsoh:
cc: @iHiD @kytrinyx

rpottsoh:
This may be related to #1172.  If so I am not sure where to find the `BUNDLED WITH` version to place into [Gemfile.lock](https://github.com/exercism/website-copy/blob/master/Gemfile.lock).  I'd be happy to update the file otherwise.

PR #1418, opened 2019-12-20
Fix Travis

iHiD:
Fixes https://github.com/exercism/website-copy/issues/1416


iHiD:
(cc @kntsoriano - we're going to need to change this all over the place)

rpottsoh:
Thanks for the fix @iHiD

iHiD:
Thanks for raising the issue too :)

PR #1428, opened 2019-12-27
remove suggestion to use parseInt with Array.map

pqnet:
parseInt cannot be used directly with `Array.prototype.map`. Try in your browser:
```
['1','2','3'].map(parseInt);
```
Why it does that? <details> <summary>Click for spoiler</summary> The function `parseInt` actually has 2 arguments, not one: the string to parse, and the callback for `map` will be passed three arguments (the element, its index, the array itself). Better explanation at https://stackoverflow.com/questions/262427/why-does-parseint-yield-nan-with-arraymap </details>

### DIFF
--- a/tracks/typescript/exercises/matrix/mentoring.md
+++ b/tracks/typescript/exercises/matrix/mentoring.md
@@ -55,6 +55,7 @@ A student may use `parseInt` instead of `number`, and a student may use getters
 ### Common suggestions
 - If a student uses `foreach`, intermediary bookkeeping and `push`, suggest [`Array#map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map).
 - If a student uses a dangerous version of transpose, suggest a non-destructive one, or make sure it's only ever called once.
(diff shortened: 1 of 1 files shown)


tracks/typescript/exercises/matrix/mentoring.md
```diff
@@ -55,7 +55,6 @@ A student may use `parseInt` instead of `number`, and a student may use getters
 ### Common suggestions
 - If a student uses `foreach`, intermediary bookkeeping and `push`, suggest [`Array#map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map).
 - If a student uses a dangerous version of transpose, suggest a non-destructive one, or make sure it's only ever called once.
-- If a student uses `map(arg => parseInt(arg))` or similar, explain they can drop the anonymous arrow function and pass in `parseInt` directly.
```
SleeplessByte: Instead replace the word `parseInt` by `Number` and remove the "or similar".

pqnet: yup, number is actually good

SleeplessByte:
Thank you! Great catch

SleeplessByte:
Thank you! Great catch

PR #1429, opened 2019-12-28
haskell/nucleotide-count: Fix typos, remove unused imports

sshine:


tracks/haskell/exercises/nucleotide-count/mentoring.md
```diff
@@ -103,15 +103,13 @@ separate definition of what a valid nucleotide is.
...
-import Control.Monad (mapM)
-import Data.Semigroup ((<>))
 import Text.Read (readEither)
 
 data Nucleotide = A | C | G | T
                 deriving (Eq, Ord, Show, Read, Enum, Bounded)
 
 nucleotideCounts :: String -> Either String (Map Nucleotide Int)
-nucleotideCounts = fmap (fillZeroes . frequencies) . mapM toNucleotide
+nucleotideCounts = fmap (fillZeroes . frequencies) . traverse toNucleotide
```
MasseR: What is the reasoning for changing `mapM` to `traverse`? Based on my experience newbies usually see `mapM` before `traverse`, so the former might be better in a pedagogical sense.

Also, the comment on line 129 is now false

> This solution also uses monadic error handling, but via `mapM toNucleotide`

sshine: This is an interesting discussion that I've only so far had in my head, so I'm happy to share it:

I initially used `mapM` because it was the one I learned first and liked the naming symmetry between `map` and `mapM`. I think your argument wrt. pedagogy is correct. I changed it to `traverse` because it has "smaller" type-class assumptions with `Applicative` being the super class:

```haskell
traverse :: (Traversable t, Applicative f) => (a -> f b) -> t a -> f (t b)
mapM :: (Traversable t, Monad m) => (a -> m b) -> t a -> m (t b)
```

so you basically don't *need* for the input to be a `Monad`. With the library definition [`mapM = traverse`](https://hackage.haskell.org/package/base/docs/src/Data.Traversable.html#Traversable), this seemed "simpler".

So I thought: If a student doesn't use either in their solution, exposing them to either one will yield some kind of insight by looking at its type: With `traverse` it will be an understanding of `Traversable` and with `mapM` it will be some link between `map` and `Monad`. I initially understood `mapM` as

```
mapM f [] = return []
mapM f (x:xs) = do
  y <- f x
  ys <- mapM f xs
  return (x:xs)
```

So I guess the difference in aim (whether it makes a big difference in practice is yet speculative), is whether you want to focus on generalized traversal ("it works for more than lists") or on lifting higher-order function into a context. ("it's like `map` but with effects").

I don't have a strong opinion, though, so I've reverted to `mapM`.

MasseR: > so I've reverted to `mapM`

I don't see the revert. Remembered to push?

sshine: Apparently because this change was made with GitHub's web editor, and perhaps as a consequence of this repo's permissions, the feature branch that is awaiting to be merged is residing in this repo and not in my copy of it. So I pushed to my copy and nothing happened. I've pushed to the right feature branch now. I'm not sure I like this behavior.

sshine:
@MasseR: Besides some small refactoring of existing solutions and minor rewording, I've added to example solutions. One is based on your `Monoid`-based solution which someone on Slack requested to be added, and the other one is based on `Validation`. Unfortunately it isn't compatible with the test suite (without changin a `Right` to a `Success` in it), but I think it's worthwhile having.

The `Validation`-based solution demonstrates why you want `traverse` rather than `mapM` since `Validation` is an `Applicative` that can't be a (lawful) `Monad`. But that wasn't why I added it, I just thought it would be neat to see how one could also accumulate error messages. E.g.

```
λ> nucleotideCounts "A!C$G_T?"
Failure "!$_?"
```

SleeplessByte (approved):
@exercism/website-copy-haskell or someone from @exercism/reviewers who knows Haskell, please sanity check.

sshine:
On behalf of the @exercism/website-copy-haskell singleton collective, I wish you a happy new year.

Let us re-address the need for reviews in December 2021.

(I think this should just be merged...)

ErikSchierboom:
Thanks @sshine!

PR #1453, opened 2020-01-19
Rename elisp to emacs-lisp

iHiD:
Re https://github.com/exercism/emacs-lisp/issues/135


iHiD:
cc @benreyn @ErikSchierboom

PR #1464, opened 2020-02-01
remove my bio from mentoring

efx:
I hope to contribute in other ways per my current skill level.

### DIFF
--- a/mentors/rust/e.json
+++ b/mentors/rust/e.json
@@ -1,13 +1,5 @@
 [
   {
-    "github_username": "efx",
-    "name": "Eli Flanagan",
-    "link_text": "a post every 6 months kind of blog",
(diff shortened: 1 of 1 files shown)


efx:
Reallocating to https://github.com/exercism/rust/issues/913#issuecomment-581067492 :smile:

rpottsoh:
@efx if you are looking to help out as a maintainer it doesn't mean you still can't be a mentor too.  I thought I would mention that in case you thought the two roles were mutually exclusive; they are not.  You are not under any obligation to wear both hats.

efx:
@rpottsoh thank you for the affirmation! It is mostly a personal distinction :smile: For now I would like to focus exclusively on maintaining because I feel that will best direct my limited resources. I will keep it in mind to ease back once we get through the v3 push.

PR #1483, opened 2020-02-29
Updated GitHub profile to new account.

CraigChamberlain:


CraigChamberlain:
ok will do. just how the auto formatted set it.

CraigChamberlain:
I'll do that tomorrow.  Can you maybe just get a formatter in the automatic
tests? This is a bit tedious for us both.  I would happily run through an
auto formatter of a different type if you formatting to a particular
convention.

PR #1491, opened 2020-03-03
Add Martin Delille

MartinDelille:
Since there was no place left, I used my initial for the json file name.


rpottsoh:
Take a look at a.json in the cop mentors folder to see how how it looks.  You need to emulate that in m.json.

MartinDelille:
Hopefully the file is not full! 🤣 

I didn't noticed the square brackets and the files I checked where empty or containing only one mentor...

Fixed, commit amended and force pushed.

Subsidiary question: what about moving to YAML for this kind of informations?

PR #1511, opened 2020-03-24
Add mentoring notes for Rust Balanced Brackets exercise

seanchen1991:
This PR adds an additional example solution and a Common Suggestions section to the mentoring notes for the Balanced Brackets exercise in Rust to help mentors during their reviews. I attempted to adhere to the Rust Reverse String problem's notes as much as possible.

### DIFF
--- a/tracks/rust/exercises/bracket-push/mentoring.md
+++ b/tracks/rust/exercises/bracket-push/mentoring.md
@@ -6,15 +6,52 @@
 
 A reasonable solution should do the following:
 
+- Keep track of the associations between specific brackets using a `HashMap`, a closure, or something of the like. 
+- Perform as few explicit checks for specific brackets as possible; instead, one possible strategy is to store openers and closers in separate `HashSet`s, and then check if the current char is contained in either set. 
+- Return an expression that evaluates to a boolean instead of using an if-else. 
+
 ### Examples
 
+This solution utilizes a `HashMap` to keep track of the associated pairs of brackets, and a `HashSet` to keep track of the closing brackets. 
+```rust
+use std::collections::{HashMap, HashSet};
+
+pub fn brackets_are_balanced(string: &str) -> bool {
+  let pairs: HashMap<_, _> = [('[', ']'), ('(', ')'), ('{', '}')]
+    .iter()
+    .cloned()
+    .collect();
+  let closers: HashSet<_> = pairs.values().cloned().collect();
+  let mut stack: Vec<char> = vec![];
+
+  for char in string.chars() {
+    match pairs.get(&char) {
+      // got back the closer associated with the char, so it was a valid opener
+      Some(&closer) => {
+        // add the closer to the stack
+        stack.push(closer);
+      },
+      None => {
+        if closers.contains(&char) {
+          let expected = stack.pop();
+          if expected != Some(char) { return false; }
+        }
+      }
+    }
+  }
+
+  stack.is_empty()
+}
+```
+
+This solution utilizes a `match` statement within a closure to keep track of the associated pairs of brackets.
 ```rust
 pub fn brackets_are_balanced(string: &str) -> bool {
     let complement = |c| match c {
         '[' => ']',
         '{' => '}',
         '(' => ')',
-        _ => unreachable!(),
+        _ => unreachable!()
     };
 
     let mut expected: Vec<char> = vec![];
--- a/tracks/rust/exercises/bracket-push/mentoring.md
+++ b/tracks/rust/exercises/bracket-push/mentoring.md
@@ -1,6 +1,7 @@
 ### Concepts
 
 - stack or recursion
+- `HashMap`s and `HashSet`s
 
 ### Reasonable solutions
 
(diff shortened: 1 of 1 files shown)


tracks/rust/exercises/bracket-push/mentoring.md
```diff
@@ -1,20 +1,58 @@
 ### Concepts
 
 - stack or recursion
+- `HashMap`s and `HashSet`s
```
coriolinus: No maps or sets are required for this exercise.

tracks/rust/exercises/bracket-push/mentoring.md
```diff
@@ -1,20 +1,58 @@
...
 
 ### Reasonable solutions
 
 A reasonable solution should do the following:
 
+- Keep track of the associations between specific brackets using a `HashMap`, a closure, or something of the like.
```
coriolinus: This is bad advice, leading to way too heavy an implementation. All that is necessary is a stack.

tracks/rust/exercises/bracket-push/mentoring.md
```diff
@@ -1,20 +1,58 @@
...
 
 - stack or recursion
+- `HashMap`s and `HashSet`s
 
 ### Reasonable solutions
 
 A reasonable solution should do the following:
 
+- Keep track of the associations between specific brackets using a `HashMap`, a closure, or something of the like. 
+- Perform as few explicit checks for specific brackets as possible; instead, one possible strategy is to store openers and closers in separate `HashSet`s, and then check if the current char is contained in either set.
```
coriolinus: This is not good advice. A minimal objection: it is more efficient to match each input character exactly once. If we're storing distinct hash sets, we need to perform at least two comparisons for each input.

A longer objection: it's not clear why it would be bad practice to perform explicit checks. I submit that the following [implementation](https://github.com/coriolinus/exercism-rs/blob/master/bracket-push/src/lib.rs) is easier to read and more efficient as the new example; match statements are faster than hash lookups:

```rust
    pub fn brackets_are_balanced(input: &str) -> bool {
        let mut bracket_stack = Vec::new();
        for ch in input.chars() {
            use BracketTypes::*;
            match ch {
                '(' => bracket_stack.push(Parenthesis),
                '[' => bracket_stack.push(Square),
                '{' => bracket_stack.push(Curly),
                ')' => {
                    if bracket_stack.pop() != Some(Parenthesis) {
                        return false;
                    }
                }
                ']' => {
                    if bracket_stack.pop() != Some(Square) {
                        return false;
                    }
                }
                '}' => {
                    if bracket_stack.pop() != Some(Curly) {
                        return false;
                    }
                }
                _ => {}
            }
        }
        bracket_stack.len() == 0
    }
```

tracks/rust/exercises/bracket-push/mentoring.md
```diff
@@ -1,20 +1,58 @@
...
+        }
+      }
+    }
+  }
+
+  stack.is_empty()
+}
+```
+
+This solution utilizes a `match` statement within a closure to keep track of the associated pairs of brackets.
```
coriolinus: It's an incidental implementation detail that this implementation uses a closure; the implementation would have been semantically identical and more idiomatic if it used a separate private `fn complement` instead. The stack always consists of chars.

tracks/rust/exercises/bracket-push/mentoring.md
```diff
@@ -32,3 +70,9 @@ pub fn brackets_are_balanced(string: &str) -> bool {
     expected.is_empty()
 }
 ```
+
+### Common Suggestions
+
+- Suggest utilizing a `HashSet` or `HashMap` to hold opening and closing brackets for easy fetching in the situation that a submission manually checks for them.
```
coriolinus: Disagree per the above.

tracks/rust/exercises/bracket-push/mentoring.md
```diff
@@ -32,3 +70,9 @@ pub fn brackets_are_balanced(string: &str) -> bool {
     expected.is_empty()
 }
 ```
+
+### Common Suggestions
+
+- Suggest utilizing a `HashSet` or `HashMap` to hold opening and closing brackets for easy fetching in the situation that a submission manually checks for them. 
+- A `HashSet` of opening brackets can be easily generated from a `HashMap` of bracket pairs with `pairs().keys().cloned().collect()`; likewise a `HashSet` of closing brackets can be easily generated with `pairs().values().cloned().collect()`.
```
coriolinus: This line is correct, but I give the opposite advice: avoid sets and maps for this exercise.

seanchen1991:
Thanks for all of the detailed feedback @coriolinus! I can resubmit another one that adds the solution you put forth above, along with tips and suggestions that follow from that and the other solution in the mentor notes that don't make use of `HashMap`s and `HashSet`s. Or do you think this set of mentor notes should just be left as-is for the time being?

seanchen1991:
@coriolinus Another question I just thought of: would you go so far as to say "A reasonable solution should _not_ use `HashMap`s and/or `HashSet`s?

PR #1516, opened 2020-03-27
[WIP] Update Clock mentoring notes for modern Rust

workingjubilee:
This PR pertains to the mentoring notes for Clock.
Some general cleanup also (typos, line length).
Recommended as an extension of the spirit of #[derive] usage.
Adds the suggestion to the "A reasonable solution should..." list and a recommended comment for quick usage.


workingjubilee:
This will be extended to a slightly more general review of the Clock exercise following some comments in Slack.

PR #1551, opened 2020-04-17
CPP: Add mentoring notes for space-age

elyashiv:


tracks/cpp/exercises/space-age/mentoring.md
```diff
@@ -0,0 +1,112 @@
...
+#define _SPACE_AGE_H_
+
+#include <cstdint>
+
+namespace space_age
+{
+    class space_age final
+    {
+    public:
+        explicit space_age(const uint64_t seconds);
```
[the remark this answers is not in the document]
MartinDelille: I'd like to know more about the `explicit` constructor. Any resource or info?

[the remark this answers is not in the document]
siebenschlaefer: [cppreference](https://en.cppreference.com/w/cpp/language/explicit) / [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-explicit) / [Jason Turner's best practices](https://github.com/lefticus/cppbestpractices/blob/master/03-Style.md#avoid-implicit-conversions)

tracks/cpp/exercises/space-age/mentoring.md
```diff
@@ -0,0 +1,112 @@
+### Reasonable solutions
+```cpp
+#if !defined(_SPACE_AGE_H_)
+#define _SPACE_AGE_H_
```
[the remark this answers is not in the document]
elyashiv: Didn't know that. Changed.

tracks/cpp/exercises/space-age/mentoring.md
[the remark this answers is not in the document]
elyashiv: @siebenschlaefer Can please you give me a referce for the namespace?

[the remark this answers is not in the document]
siebenschlaefer: @elyashiv Sure. §17.6.1.2.4 [headers] in the C++14 specification: "In the C ++ standard library, however, the declarations (except for names which are defined as macros in C) are within namespace scope (3.3.6) of the namespace std." ([§16.5.1.2.5](http://eel.is/c%2B%2Bdraft/headers#5) in the current working draft)

tracks/cpp/exercises/space-age/mentoring.md
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    double space_age::on_neptune() const
+    {
+        return on_planet(planet::NEPTUNE);
+    }
+}
+```
+### Common suggestions
+ * Save seconds as a const in the class
```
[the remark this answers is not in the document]
elyashiv: You are right. I'm just used to manage resources in a class, so I just make a non-copyable class without thinking unless I have a reason...

marvelou-s:
@elyashiv you should take into account extracted concepts, mistakes and approaches already done to [space-age](https://github.com/exercism/v3/commits/master/languages/cpp/reference/exercise-concepts/space-age.md) exercise in v3 repo.

elyashiv:
@marvelou-s I wasn't aware of the v3 changes... The concept file seems to make the mentoring notes redundant. How are they related?

marvelou-s:
They have nothing in common between each other. I think your idea is good, especially until v3 will not be released. In order to not reinvent the wheel I suggest to supplement the file with previously extracted concepts by another mentors. Maybe you will notice something what we was not ;)

elyashiv:
I want to merge this. Any thing else that should be done?

tracks/cpp/exercises/space-age/mentoring.md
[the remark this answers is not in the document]
siebenschlaefer: @MartinDelille Include guards vs. `#pragma once` still debated in the C++ community (see [StackOverflow](https://stackoverflow.com/a/34884735/1447760)).  
But [SF.8 of the C++ Core Guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf8-use-include-guards-for-all-h-files) takes a stand against `#pragma once` and recommends that you should always use `#include` guards in your header files.

elyashiv:
Is this ready to be merged?

elyashiv:
ping?

tracks/cpp/exercises/space-age/mentoring.md
```diff
@@ -0,0 +1,124 @@
...
+```
+### Common suggestions
+ * Make sure the data type is big enough to store a 33 bit value (needed for neptune)
+ * Make sure the data type is consistent (constructor, class member and return type from `seconds()`)
+ * Constants should be static.
+ * Arguments with names that indicate the units used.
+ * Use `constexpr` for the functions.
+### Talking points
+ * `explicit` constructor so you don't accidently do a type casting. [examples](https://en.cppreference.com/w/cpp/language/explicit).
+ * Prefer intialization over assinment. [Good explanation](https://stackoverflow.com/a/7350783)
```
Llcoolsouder, line 114: assignment

tracks/cpp/exercises/space-age/mentoring.md
```diff
@@ -0,0 +1,124 @@
...
+    { }
+
+    std::int64_t space_age::seconds() const
+    {
+        return _seconds;
+    }
+
+    double space_age::on_planet(const planet p) const
+    {
+        return _seconds * 1.0 / static_cast<std::int64_t>(p);
```
Llcoolsouder, line 62:
```suggestion
        return static_cast<double>(_seconds) / static_cast<std::int64_t>(p);
```
Seems more explicit than promoting implicitly through a multiply by 1.0.

tracks/cpp/exercises/space-age/mentoring.md
```diff
@@ -0,0 +1,124 @@
...
+### Talking points
+ * `explicit` constructor so you don't accidently do a type casting. [examples](https://en.cppreference.com/w/cpp/language/explicit).
+ * Prefer intialization over assinment. [Good explanation](https://stackoverflow.com/a/7350783)
+ * Being explicit about units.
+ * `signed` vs. `unsigned`.
+ ** `unsigned` can give some suprising results. [example](isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es102-use-signed-types-for-arithmetic)
+ * `double` vs. `float`
+ * Implementation in header:
+ ** This will turn the functions to be implicitly inline, and allow the compiler to do some optimizations on these very small functions.
+ ** This also means that every change in the implementation will require a recompile, and will create longer compile time.
```
Llcoolsouder, line 121:
```suggestion
 * Implementation in header:
   * This will turn the functions to be implicitly inline, and allow the compiler to do some optimizations on these very small functions.
   * This also means that every change in the implementation will require a recompile, and will create longer compile time.
```

tracks/cpp/exercises/space-age/mentoring.md
```diff
@@ -0,0 +1,124 @@
...
+ * `const` on seconds
+ ** We don't want to change it, so we can declare it const so we don't accidenlty do something stupid.
+ ** This won't allow us to use the assignment operator. Not needed for the exercise, but might be unintented.
```
Llcoolsouder, line 124:
```suggestion
 * `const` on seconds
   * We don't want to change it, so we can declare it const so we don't accidentally do something stupid.
   * This won't allow us to use the assignment operator. Not needed for the exercise, but might be unintended.
```

tracks/cpp/exercises/space-age/mentoring.md
```diff
@@ -0,0 +1,124 @@
...
+    enum class space_age::planet : std::int64_t
+    {
+        Earth = 31557600,
+        Mercury = 7600544,
+        Venus = 19414149,
+        Mars = 59355036,
+        Jupiter = 374355660,
+        Saturn = 929292360,
+        Uranus = 2651370000,
+        Neptune = 5200418600,
```
Llcoolsouder, line 48: Should we expect users to convert everything to seconds or stick with units of "Earth years"? Or does it even matter?

tracks/cpp/exercises/space-age/mentoring.md
```diff
@@ -0,0 +1,124 @@
...
+ * Make sure the data type is consistent (constructor, class member and return type from `seconds()`)
+ * Constants should be static.
+ * Arguments with names that indicate the units used.
+ * Use `constexpr` for the functions.
+### Talking points
+ * `explicit` constructor so you don't accidently do a type casting. [examples](https://en.cppreference.com/w/cpp/language/explicit).
+ * Prefer intialization over assinment. [Good explanation](https://stackoverflow.com/a/7350783)
+ * Being explicit about units.
+ * `signed` vs. `unsigned`.
+ ** `unsigned` can give some suprising results. [example](isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es102-use-signed-types-for-arithmetic)
```
Llcoolsouder, line 117:
```suggestion
 * `signed` vs. `unsigned`.
   * `unsigned` can give some surprising results. [example](isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es102-use-signed-types-for-arithmetic)
```

tracks/cpp/exercises/space-age/mentoring.md
elyashiv: I don't think it matters.
One think that can be improved here is adding a unit to this values - seconds in this case.
I do wander if there is a way to use std::chrono for this, as it makes the units nice and clear, and allows converting easily.

tracks/cpp/exercises/space-age/mentoring.md
```diff
@@ -0,0 +1,124 @@
...
+        return on_planet(planet::Uranus);
+    }
+
+    double space_age::on_neptune() const
+    {
+        return on_planet(planet::Neptune);
+    }
+}
+```
+### Common suggestions
```
SleeplessByte, line 106:
```suggestion
### Common suggestions

```

tracks/cpp/exercises/space-age/mentoring.md
```diff
@@ -0,0 +1,124 @@
...
+ * Make sure the data type is big enough to store a 33 bit value (needed for neptune)
+ * Make sure the data type is consistent (constructor, class member and return type from `seconds()`)
+ * Constants should be static.
+ * Arguments with names that indicate the units used.
+ * Use `constexpr` for the functions.
+### Talking points
```
SleeplessByte, line 112:
```suggestion
### Talking points

```

tracks/cpp/exercises/space-age/mentoring.md
```diff
@@ -0,0 +1,124 @@
+### Reasonable solutions
```
SleeplessByte, line 1:
```suggestion
### Reasonable solutions

```

SleeplessByte (dismissed):
LGTM now. @elyashiv small nitpick. 

Let's merge this within the week regardless if Elyashiv makes the change.

angelikatyborska:
> @angelikatyborska dismissed SleeplessByte’s stale review via 3834116 1 minute ago 
Interesting that applying suggestions from an approving review causes the review to be automatically dismissed.

SleeplessByte (approved):

PR #1563, opened 2020-04-25
[JavaScript]Update mentoring.md for Leap

hayashi-ay:
I fix a sample code because it does't pass the test case, `year divisible by 4, not divisible by 100: leap year'`.

The previous one returns `16` when a input is `2016`, so I fix `true` is returned.

### DIFF
--- a/tracks/javascript/exercises/leap/mentoring.md
+++ b/tracks/javascript/exercises/leap/mentoring.md
@@ -14,7 +14,7 @@ export const isLeap = (year) => year % 4 === 0 && (year % 100 !== 0 || year % 40
 ```
 
 ```javascript
-export const isLeap = (year) => !(year % 4) && year % 100 || !(year % 400);
+export const isLeap = (year) => !(year % 4) && !!(year % 100) || !(year % 400);
 ```
 
 ```javascript


hayashi-ay:
but how about readability? Do you think double negate decreases readability or not? I doubt whether I should teach students that solution.

yawpitch:
Readability is why we do direct comparisons in eg Python.

```python
def is_leap(year):
    return year % 4 == 0 and (year % 100 != 0 or year % 400 == 0)
```

So --- while I don't use JS -- something like:

```javascript
export const isLeap = (year) => (year % 4 == 0) && (year % 100 != 0) || (year % 400 == 0);
```

hayashi-ay:
@SleeplessByte Hi, could you review this? Also, I submit another [PR][pr].

[pr]: https://github.com/exercism/javascript/pull/857

hayashi-ay:
@tejasbubane @SleeplessByte Could you review this PR? It just fixes one line so it won't take much time.

SleeplessByte:
I don't think double negative, in this case, increases readability and I agree with @yawpitch; direct comparisons are much better. However, in JavaScript, relying on falsy _is_ idiomatic, so that 's why it was a reasonable solution in the first place.

yawpitch:
> I don't think double negative, in this case, increases readability and I agree with @yawpitch; direct comparisons are much better. However, in JavaScript, relying on falsy _is_ idiomatic, so that 's why it was a reasonable solution in the first place.

Just to clarify, False-y is also idiomatic in Python; we choose to de-normalize the idiom when it makes the reading of the code more natural. So we see `if x % y == 0` (“if x modulo divides by y with no remainder”) as easier to read than `if not x % y` (“if the try truthiness of x modulo divided by y, inverted, is True”.

In a compiled language of course compiler optimization could de-sugar both to the same runtime code, but in Python we tend to accept the runtime hit for the easier read. That said it’s really just a taste thing.

SleeplessByte:
Yes, because it's a taste thing we accept both ;)

PR #1570, opened 2020-04-27
elixir/word-count: Update mentoring.md

Eiermann:
Added a short solution using Enum.frequencies.

### DIFF
--- a/tracks/elixir/exercises/word-count/mentoring.md
+++ b/tracks/elixir/exercises/word-count/mentoring.md
@@ -68,6 +68,7 @@ defmodule WordCount do
   
 end
 ```
+
 ### Common suggestions
 
 #### German `öüä`
--- a/tracks/elixir/exercises/word-count/mentoring.md
+++ b/tracks/elixir/exercises/word-count/mentoring.md
@@ -66,7 +66,6 @@ defmodule WordCount do
     |> List.flatten
     |> Enum.frequencies
   end
-  
 end
 ```
 
--- a/tracks/elixir/exercises/word-count/mentoring.md
+++ b/tracks/elixir/exercises/word-count/mentoring.md
@@ -56,7 +56,18 @@ defmodule WordCount do
   end
 end
 ```
+```elixir
+# Utilizing Enum.frequencies
+defmodule WordCount do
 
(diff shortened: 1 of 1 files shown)


tracks/elixir/exercises/word-count/mentoring.md
```diff
@@ -56,7 +56,18 @@ defmodule WordCount do
...
+defmodule WordCount do
 
+  def count(sentence) do
+    Regex.scan(~R"[[:alnum:]-]+"u, String.downcase(sentence))
+    |> List.flatten
+    |> Enum.frequencies
+  end
+  
+end
+```
```
SleeplessByte, line 70:
````suggestion
```

````

File: tracks/elixir/exercises/word-count/mentoring.md [hunk 2]
```diff
@@ -56,7 +56,18 @@ defmodule WordCount do
...
+```elixir
+# Utilizing Enum.frequencies
+defmodule WordCount do
 
+  def count(sentence) do
+    Regex.scan(~R"[[:alnum:]-]+"u, String.downcase(sentence))
+    |> List.flatten
+    |> Enum.frequencies
+  end
+
```
NobbZ, line 68:
```suggestion

```

tracks/elixir/exercises/word-count/mentoring.md
```diff
@@ -56,7 +56,18 @@ defmodule WordCount do
   end
 end
 ```
+```elixir
```
NobbZ, line 59:
```suggestion

```elixir
```

NobbZ (changes requested):
Even though `Enum.frequencies/1,2` exists, I'm not really sure if it is within "first principles" or a valid use of standard library…

Also we have to remember, that not everyone already has access to elixir 1.10, and therefore they might not have access to this function, as it has been added with 1.10.

Therefore I think it should not only the example added, but the use of `Enum.frequencies/1,2` could be something to talk about, but nothing that would actually block a students progress.

NobbZ (approved):

SleeplessByte:
Thanks @NobbZ for re-checking.

PR #1576, opened 2020-05-01
[Julia] more mentoring notes

cmcaine:


cmcaine:
@SaschaMann

tracks/julia/exercises/difference-of-squares/mentoring.md
```diff
@@ -4,6 +4,24 @@
...
+- Analytical solutions for sum of the first N square numbers: [square pyramidal number](https://en.wikipedia.org/wiki/Square_pyramidal_number)
+
 ### Talking points
 
+- You might be interested to know that `sum(1:n)` already uses Gauss' formula. You can see this with `@edit sum(1:n)`
 - For integer division, students should use `÷` and not `/` (see [#138](https://github.com/exercism/julia/issues/138))
+
+### Interesting solutions
+
+n111b111's solution:
```
SaschaMann, line 16: Perhaps adding a link to their solution on exercism.io makes it easier to link this to students? I'm not sure if they can access it though

tracks/julia/exercises/difference-of-squares/mentoring.md
```diff
@@ -4,6 +4,24 @@
 
 ![](https://wikimedia.org/api/rest_v1/media/math/render/svg/5852a9aaf138078bfb5a60e1e6cad558b57a9c3e)
 
+- Analytical solutions for sum of the first N square numbers: [square pyramidal number](https://en.wikipedia.org/wiki/Square_pyramidal_number)
+
 ### Talking points
 
+- You might be interested to know that `sum(1:n)` already uses Gauss' formula. You can see this with `@edit sum(1:n)`
```
SaschaMann, line 11:
```suggestion
- You might be interested to know that `sum(1:n)` already uses Gauss' formula. You can see this with `@less sum(1:n)`
```

This is personal preference but for only looking up definitions, `@less` might be nicer.

tracks/julia/exercises/leap/mentoring.md
```diff
@@ -0,0 +1,17 @@
...
+We don't need to bikeshed on this too long. Just prompt them to improve their code if they do something needlessly convoluted. `if` statements are okay, but consider prompting them with the first or third suggestions.
+
+
+### Common suggestions
+
+(Formatted so you can copy and paste easily)
+
+- Whenever you have an expression like ``if condition; return true; else; return false``, you can rewrite it as ``return condition``.
+- Obviously you can do as you please, but it is conventional to have one space either side of each operator: ``year % 400 == 0``.
+- Consider writing a truth table for this exercise and see if you can work out what the simplest predicates are and how they could be combined with ``&&`` and ``||``
```
SaschaMann, line 12: Could you add a link to a simple example truth table? Programmers without formal CS/maths background might not be familiar with it.

tracks/julia/exercises/nucleotide-count/mentoring.md
```diff
@@ -0,0 +1,40 @@
...
+Students that use four counting variables rather than mutating a dict should be shown the dict solution or asked to come up with it on their own if there is some doubt that they knew they could do that.
+
+We shouldn't require the mutating-a-dict strategy because it's actually pretty slow.
+
+
+### Example solutions
+
+```julia
+function count_nucleotides(strand)
+    counter = Dict(base => 0 for base in "ACGT")
```
SaschaMann, line 12: That's a really nice way of constructing this dict, haven't seen that before :D

tracks/julia/exercises/difference-of-squares/mentoring.md
cmcaine: I prefer to give it inline to avoid having to open a other page to see.

Also, users can change their solutions to something else whenever they like.

tracks/julia/exercises/difference-of-squares/mentoring.md
cmcaine: I've found @edit less buggy 🤷‍♂️

tracks/julia/exercises/leap/mentoring.md
cmcaine: Done

cmcaine:
I think this is better than the status quo and therefore good to go :)

tracks/julia/exercises/nucleotide-count/mentoring.md
ScottPJones: I would rather *not* show the Dict solution - most of my time mentoring these past two weeks has been explaining that `Dict` or `Set` is often *not* their friend, if they care at all about performance.

tracks/julia/exercises/nucleotide-count/mentoring.md
```diff
@@ -0,0 +1,40 @@
...
+    for base in strand
+        if haskey(counter, base)
+            counter[base] += 1
```
ScottPJones, line 15: This to me is also something I bring up with anybody who has expressed interest in getting a performant solution.
That makes 3 Dict lookups (I feel even 1 will make things too slow)

tracks/julia/exercises/nucleotide-count/mentoring.md
```diff
@@ -0,0 +1,40 @@
...
+        end
+    end
+    return counter
+end
+```
+
+```julia
+function count_nucleotides(strand)
+    counts = Dict(count(==(base), strand) for base in "ACGT")
+    if sum(values(counts)) != length(strand)
```
ScottPJones, line 27: This also - for Julia's `String` type, because it is UTF-8 encoded (loosely speaking, it's not validated, which makes it even slower), `length(strand)` is an O(N) operation - so if you are checking a lot of strands / long strands, that will be a lot slower than it needs to be.

tracks/julia/exercises/nucleotide-count/mentoring.md
SaschaMann: Do you have an example of a solution without Dicts that could be added here?

tracks/julia/exercises/rotational-cipher/mentoring.md
```diff
@@ -0,0 +1,80 @@
...
+"Explain what this method does here"
+function rotate(n, str)
+    ...
+end
+```
+
+
+### Talking points
+
+- You can make this a fair bit faster if you trust the string is ASCII and use codeunits() and some unsigned integer arithmetic. Unsigned int literals look like 0xF1 or 0b11110001
```
SaschaMann, line 79: I don't think it's good practice to assume that strings are ASCII unless they're appropriately typed.

tracks/julia/exercises/nucleotide-count/mentoring.md
cmcaine: The reason for this is that I would like the students to know how to use dictionaries and index collections.

I generally encourage students to give correct, concise answers and not to worry too much about performance, and I think that's a decent default for real-world programming, or at least it has been for me.

Scott, you have more experience than me, though :)

I think we should show the more imperative/explicit fast solutions, too, though. Something for these after-exercise markdown files that we're getting in v3 (for all exercises or just concept?)

tracks/julia/exercises/rotational-cipher/mentoring.md
cmcaine: The exercise spec promises ASCII, but maybe it should be changed, it's not terribly realistic, after all.

tracks/julia/exercises/nucleotide-count/mentoring.md
cmcaine: I know, but I don't think that makes this a bad solution.

tracks/julia/exercises/nucleotide-count/mentoring.md
cmcaine: Without context, I don't think any solution can be said to be too slow.

I think it's fine to point out how to make things more performant, but these answers are also fine, and I think they may generalise to a wider set of problems better.

tracks/julia/exercises/rotational-cipher/mentoring.md
SaschaMann: I think that would  broaden the scope of the exercise too much. I just think we shouldn't encourage them to assume it's ASCII when implementing a generic method from the Julia code PoV, i.e. while it's, imo, fine to point out an implementation that works better for ASCII, they should also be aware to dispatch on `ASCIIString` (or w/e it's called nowadays) or add a check.

cmcaine: There's no base ASCIIStr and the one in Strs.jl performs validation. I think this is an okay performance optimisation, if documented :shrug: 

I can remove this suggestion, though.

cmcaine:
Review comments have been addressed. Please approve and merge :)

tracks/julia/exercises/nucleotide-count/mentoring.md
ScottPJones: I think that exercises where a `Dict` is the optimal solution, on all fronts, should be added then.
Many people try out Julia precisely because of they've heard about the performance, and if they use non-performant techniques, and then see that it's not any faster than Python, then they will no longer be interested in pursuing it.
(That almost happened to me, when I saw how much slower Julia was for string handling!)

In my background, performance has always been critical - it meant the difference between people choosing our solution, or the competitions.
I've seen projects (and companies) die because not enough care was taken about performance.

I'd encourage students to first come up with a clean, easy to read solution, but never to stop just there, show them where that might end up with O(N) or O(N^2) algorithms, and how to improve on that.

ScottPJones: > Do you have an example of a solution without Dicts that could be added here?

Yes, I'll do so tomorrow or Monday!

tracks/julia/exercises/nucleotide-count/mentoring.md
ScottPJones: Well - anything that deals with DNA samples, and doesn't scale, I don't think would be a useful solution for any real-world problem.

SaschaMann:
Very nice! Thanks :)

tracks/julia/exercises/nucleotide-count/mentoring.md
cmcaine: I wrote one last night and added it :)

cmcaine: Well, we still construct a dict right at the end, but that could be a named tuple if the spec was changed.

cmcaine:
Turns out double \` aren't enough to show the backticks in exercism to make them copyable. I guess I'll try escaping them with a backslash next time or something.

SaschaMann:
If that doesn't work, open an issue in [exercism/exercism](https://github.com/exercism/exercism)

tracks/julia/exercises/nucleotide-count/mentoring.md
ScottPJones: Yes, I think using a named tuple would be very nice (showing off an important feature of Julia also)

cmcaine:
Maybe of interest to folk talking about the performance of the count_nucleotides function, I did some benchmarking this eve:

```jl
# A student's solution
function count_nucleotides1(strand)
    nucleotides = Dict(
        'A' => 0,
        'G' => 0,
        'T' => 0,
        'C' => 0
    )
    for l in strand
        !haskey(nucleotides, l) && throw(DomainError("Invalid strand"))
        nucleotides[l] += 1
    end
    return nucleotides
end

function count_nucleotides2(strand)
    ca = cc = cg = ct = 0
    for nuc in strand
        if nuc == 'A'
            ca += 1
        elseif nuc == 'C'
            cc += 1
        elseif nuc == 'G'
            cg += 1
        elseif nuc == 'T'
            ct += 1
        else
            throw(DomainError(nuc, "only A, C, G and T are valid nucleotides"))
        end
    end
    return Dict('A' => ca, 'C' => cc, 'G' => cg, 'T' => ct)
end

function count_nucleotides3(strand)
    counts = Dict(base => count(==(base), strand) for base in "ACGT")
    if sum(values(counts)) != length(strand)
        throw(DomainError(strand, "only A, C, G and T are valid nucleotides"))
    end
    counts
end

using Random: randstring
using BenchmarkTools: @btime

input = randstring("ACGT", Int(1e6));

@btime count_nucleotides1(input) # 20ms
@btime count_nucleotides2(input) # 7ms
@btime count_nucleotides3(input) # 5ms


# Now guarantee that the input is ASCII (usually speeds up iterating a string
# a lot if you don't need to guess how long the next char is)
using Strs: ASCIIChr, ASCIIStr

ascii_input = ASCIIStr(input);

# This is slightly slower than the String version, but I don't know why.
function count_nucleotides2a(strand::ASCIIStr)
    ca = cc = cg = ct = 0
    for nuc in strand
        if nuc == ASCIIChr('A')
            ca += 1
        elseif nuc == ASCIIChr('C')
            cc += 1
        elseif nuc == ASCIIChr('G')
            cg += 1
        elseif nuc == ASCIIChr('T')
            ct += 1
        else
            throw(DomainError(nuc, "only A, C, G and T are valid nucleotides"))
        end
    end
    return Dict('A' => ca, 'C' => cc, 'G' => cg, 'T' => ct)
end

function count_nucleotides3a(strand::T) where T
    counts = Dict(base => count(==(base), strand) for base in T("ACGT"))
    if sum(values(counts)) != length(strand)
        throw(DomainError(strand, "only A, C, G and T are valid nucleotides"))
    end
    counts
end

# Not as big an improvement as I expected, tbh.
@btime count_nucleotides2a(ascii_input) # 8ms (+1ms)
@btime count_nucleotides3a(ascii_input) # 4ms (-1ms)

# Threaded versions of a bunch of Base Julia functions.
import ThreadsX

function count_nucleotides3b(strand::T) where T
    counts = Dict(base => ThreadsX.count(==(base), strand) for base in T("ACGT"))
    if sum(values(counts)) != length(strand)
        throw(DomainError(strand, "only A, C, G and T are valid nucleotides"))
    end
    counts
end

@btime count_nucleotides3b(ascii_input) # 2.6ms (-2ms)
```

That's 380 MB/s. SIMDJson parses at 2.5 GB/s (on a single core, I think), so we're definitely not getting the most out of the hardware, but this is respectable for pretty low effort, I think

PR #1579, opened 2020-05-03
[Python] "Reverse String" mentoring notes

lancelote:
Mentoring notes for "Reverse String" Python track side exercise

### DIFF
--- a/tracks/python/exercises/reverse-string/mentoring.md
+++ b/tracks/python/exercises/reverse-string/mentoring.md
@@ -4,22 +4,23 @@ The problem requires the student to reverse a given string.
 
 ### Reasonable solutions
 
-[Extended Slice Notation][1] is the fastest way to reverse a string in Python
+[Extended Slice Notation][1] is the fastest way to reverse a string in Python.
 
 ```python
 def reverse(text):
     return text[::-1]
 ```
 
-Though it can be tricky for beginners to comprehend, also some developers
-found it not quite readable, so
+Though it can be tricky for beginners to comprehend and some developers
+consider it less readable, so
 
 ```python
 def reverse(text):
     return "".join(reversed(text))
 ```
 
-is also a good straightforward solution with a decent performance.
+... is also a reasonable straightforward solution with an acceptable
+performance.
 
 ### Unreasonable solutions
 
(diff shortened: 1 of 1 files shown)


tracks/python/exercises/reverse-string/mentoring.md
```diff
@@ -0,0 +1,74 @@
...
+### Reasonable solutions
+
+[Extended Slice Notation][1] is the fastest way to reverse a string in Python
+
+```python
+def reverse(text):
+    return text[::-1]
+```
+
+Though it can be tricky for beginners to comprehend, also some developers
```
yawpitch, line 14: Awkward grammar here, perhaps "Though it can be tricky for beginners to comprehend and some developers
consider it less readable, so"?

tracks/python/exercises/reverse-string/mentoring.md
```diff
@@ -0,0 +1,74 @@
...
+found it not quite readable, so
+
+```python
+def reverse(text):
+    return "".join(reversed(text))
+```
+
+is also a good straightforward solution with a decent performance.
```
yawpitch, line 22: To visibly show that the sentence has been interrupted by the code example put an ellipsis at the front of this string, ie "... is also a".

Also, I would suggest replacing "good" with "reasonable" and "decent" with "acceptable" ... this is **definitely** the less idiomatic and less performant solution, we should accept it, but not promote it.

tracks/python/exercises/reverse-string/mentoring.md
```diff
@@ -0,0 +1,74 @@
...
+def reverse(text):
+    i = len(text) - 1
+    new_word = ''
+    while i > -1:
+        new_word = new_word + text[i]
+        i = i - 1
+    return new_word
+```
+
+- Manual string concatenation is slow and should be generally avoided
```
yawpitch, line 36: IMHO "should generally be" is easier to read than "should be generally".

tracks/python/exercises/reverse-string/mentoring.md
```diff
@@ -0,0 +1,74 @@
+### Problem and challenges
+
+The problem requires the student to reverse a given string.
+
+### Reasonable solutions
+
+[Extended Slice Notation][1] is the fastest way to reverse a string in Python
```
yawpitch, line 7: Be consistent with punctuation; some sentences end with a period, others do not.

lancelote:
Thank you for the review, I greatly appreciate your time 🙇🏻‍♂️

PR #1581, opened 2020-05-04
Added my bio under n in Rust-track

niilz:
This pull request contains my bio, compliant with the Exercism-Tutor-Bio-Template.

**Note:** I have added a *rust-tag* which looks like this ```:rust:```. But I couldn't find whether it's supported. If not I can change it to a  🦀 emoji. Unless no emojis are supported or if you prefer not to use them in the bios.

### DIFF
--- a/mentors/rust/n.json
+++ b/mentors/rust/n.json
@@ -1,1 +1,10 @@
-[]
+[
+  {
+    "github_username": "niilz",
+    "name": "Nils Haberstroh",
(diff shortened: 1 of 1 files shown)


rpottsoh:
I don't know if the emoji will be supported on not.  Once your bio has been merged you should go find it on the web site.  If the emoji isn't what you would like then open another PR and modify your bio accordingly.

PR #1588, opened 2020-05-09
python/matrix: Mention separation of concerns

WeeDom:
I think it's worth at least introducing the concept of SoC at this stage, to get students thinking along those lines. I've had positive feedback from the students I've mentored on this subject

### DIFF
--- a/tracks/python/exercises/matrix/mentoring.md
+++ b/tracks/python/exercises/matrix/mentoring.md
@@ -55,3 +55,26 @@ update an entry.
 
 In any case, this isn't a topic to burden a beginner with, 
 but you may want to comment about it if you see it in the solution. 
+
+### Separation of Concerns
+
+You will see solutions where the student has split `matrix_string` in the constructor, then done the actual construction of the matrix in the getters.
+
+```
+class Matrix:
+    def __init__(self, matrix_string):
+        self.matrix = matrix_string.split('\n')
+
+    def row(self, index):
+        return list(map(int, self.matrix[index - 1].split(' ')))
+
+    def column(self, index):
+        column = []
+        for row in self.matrix:
+            column.append(int(row.split(' ')[index - 1]))
+        return column
+```
+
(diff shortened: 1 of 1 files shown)


tracks/python/exercises/matrix/mentoring.md
```diff
@@ -55,3 +55,26 @@ update an entry.
...
+    def column(self, index):
+        column = []
+        for row in self.matrix:
+            column.append(int(row.split(' ')[index - 1]))
+        return column
+```
+
+You should explain SoC to your student, and ask them to submit another iteration where the getters (`row`, `column`) do their one job - interrogate the matrix in `self` for the data requested.
+
+There's a bit of ambiguity as to whether the constructor should also construct the matrix in such a way that `column` can be retrived with something like `return self.columns[index - 1]`. It's probably not important enough to decline approval if they've taken that route, although I'd ask where that ends. If we want to retrieve the top-left value, or the bottom right, do we write another constructor line to store that in `self`? I'd argue not, but it's fairly subjective.
```
yawpitch, line 80: Up until this paragraph I'm with you, but the exercise doesn't really imply how the matrix itself should be constructed, and the fact that you end up waffling about the import of your own point kind of argues against the value of including this line.

tracks/python/exercises/matrix/mentoring.md
```diff
@@ -55,3 +55,26 @@ update an entry.
...
+```
+class Matrix:
+    def __init__(self, matrix_string):
+        self.matrix = matrix_string.split('\n')
+
+    def row(self, index):
+        return list(map(int, self.matrix[index - 1].split(' ')))
+
+    def column(self, index):
+        column = []
```
yawpitch, line 72: Just for visual simplicity I'd implement both **row** and **column** using a single list comprehension; the point we're trying to make here is not parsing the matrix string itself in either method when it could be done in the `__init__`, and the two different implementations are distractions from that point.

SleeplessByte:
@WeeDom if you're still interested, see the review notes above.

BethanyG:
Closing as @WeeDom agrees this is no longer relevant.

PR #1591, opened 2020-05-12
Add bob rundle as C++ mentor

bobrundle:


mentors/cpp/b.json
```diff
@@ -1 +1,10 @@
-[]
+[
+    {
+        "github_username": "bobrundle",
+        "name": "Bob Rundle",
+        "link_text": null,
+        "link_url": null,
+        "avatar_url": null,
+        "bio": "I have close to 40 years of experience designing and deploying commercial software products for the oil and gas industry."
+    }
```
rpottsoh, line 9:
```suggestion
  {
    "github_username": "bobrundle",
    "name": "Bob Rundle",
    "link_text": null,
    "link_url": null,
    "avatar_url": null,
    "bio": "I have close to 40 years of experience designing and deploying commercial software products for the oil and gas industry."
  }
```
Indents should only be 2 in for each level.

PR #1625, opened 2020-05-28
Add mentor bio for filaretov

filaretov:
Hey, I just became a mentor on the Python track and this is my bio.


mentors/python/f.json
```diff
@@ -15,5 +15,14 @@
     "link_url": null,
     "avatar_url": null,
     "bio": "I love Python. I'm from Portugal. Started learning by myself a few years ago. Love to teach so that way I assure myself that I fully understand. Currently studying Software Engineering. I'm awesome :)"
+  },
+
```
rpottsoh, line 19:
```suggestion
```
I know that line 10 is also blank, but it shouldn't be either.

filaretov: Right, my bad. I removed both blank lines.

PR #1639, opened 2020-06-02
julia: add commentary to secret-handshake mentor notes

cmcaine:
Sorry about the extra PR, I wanted to see if github would allow merging
twice.


SaschaMann:
No worries, thanks for adding the comments :)

PR #1640, opened 2020-06-05
Update d.json to add drentropy

DrEntropy:
Adding my bio as per the Mentor instructions.

### DIFF
--- a/mentors/cpp/d.json
+++ b/mentors/cpp/d.json
@@ -1,1 +1,8 @@
-[]
+[{
+    "github_username": "drentropy",
+    "name": "Ronald Legere",
+    "link_text": "",
+    "link_url": "",
+    "avatar_url": null,
+    "bio": "C++ is not elegant but it does provide awesome power and expressiveness when mastered. These days I primarily use C++ for audio development but also tinker with gaming."
+  }]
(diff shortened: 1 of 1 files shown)


mentors/cpp/d.json
```diff
@@ -1 +1,8 @@
-[]
+[{
+    "github_username": "drentropy",
+    "name": "Ronald Legere",
+    "link_text": "",
+    "link_url": "",
+    "avatar_url": null,
+    "bio": "C++ is not elegant but it does provide awesome power and expressiveness when mastered. These days I primarily use C++ for audio development but also tinker with gaming."
+  }]
```
rpottsoh, line 8:
```suggestion
[
  {
    "github_username": "drentropy",
    "name": "Ronald Legere",
    "link_text": "",
    "link_url": "",
    "avatar_url": null,
    "bio": "C++ is not elegant but it does provide awesome power and expressiveness when mastered. These days I primarily use C++ for audio development but also tinker with gaming."
  }
]
```
Content is fine, just minor formatting changes.

DrEntropy: Thanks! I didnt notice that for some reason.

PR #1641, opened 2020-06-06
Julia: improve robot-name example

cmcaine:
Thanks for the code review, Seth!

### DIFF
--- a/tracks/julia/exercises/robot-name/mentoring.md
+++ b/tracks/julia/exercises/robot-name/mentoring.md
@@ -37,11 +37,15 @@ There's some basic stats you can do:
 
 using Random: shuffle!
 
-# Use Int32 to save some bits because this is quite a big vector.
+# We shuffle all possible robot names in advance so that we can issue them in
+# constant time.
+#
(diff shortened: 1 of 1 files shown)


tracks/julia/exercises/robot-name/mentoring.md
```diff
@@ -37,57 +37,60 @@ There's some basic stats you can do:
...
-    # Robots are comfortable with integer names ;)
     id::Int32
 
     """
         Robot()
 
     Construct a robot with a unique name within its class.
     """
-    Robot() = new(mint_id())
+    Robot() = new(mint_id!())
```
SaschaMann, line 56: I didn't know that the trailing `!` is also used for functions that mutate global or module state, but I guess there are some functions that do it (like `Random.seed!`). Good to know :)

SaschaMann:
Thanks!

tracks/julia/exercises/robot-name/mentoring.md
cmcaine: I'm not actually 100% sure about this, but I think it makes sense.

PR #1645, opened 2020-06-12
Julia: add mentoring notes for clock

cmcaine:
Code examples are lightly edited by me. Errors are my own, etc.

@ScottPJones, I edited your code slightly to remove an `==` method and
to use `value()` rather than `.value`. Let me know if you think your
original version should stand.

Also let me know if you'd prefer I not use your code as an example :)

### DIFF
--- a/tracks/julia/exercises/clock/mentoring.md
+++ b/tracks/julia/exercises/clock/mentoring.md
@@ -1,0 +1,118 @@
+## Things to watch out for
+
+- It is preferred to use \``value(m)`\` rather than \``m.value`\`. Fields are generally considered private in Julia. You can import \``value`\` from Dates.
+
+If the student has a field for both hours and minutes, then
+the trickiest case is when minutes are less than and not a multiple of 60 (e.g. -61).
+Students often end up doing something quite convoluted to solve this.
+
+Students often use `%` or `rem` and a subtraction when `mod` is probably what they want.
+
+Your humble correspondent thinks that this is best solved by converting the hours to minutes, as is done in OTDE's solution below.
+
+## Fun tricks
+
+You can define the subtraction in terms of unary negation and addition, like this:
+
+````
+```julia
+-(c::Clock, m::Minute) = c + -m
+```
+````
+
+## Example solutions
+
+OTDE's solution
+
+````
+```julia
+using Dates
+
+struct Clock
+    min::Int
+    Clock(min) = new(mod(min, 60 * 24))
+end
+
+Clock(hour, min) = Clock(hour * 60 + min)
+
+Base.:(+)(c::Clock, min::Minute) = Clock(c.min + value(min))
+Base.:(-)(c::Clock, min::Minute) = Clock(c.min - value(min))
+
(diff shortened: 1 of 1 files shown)


tracks/julia/exercises/clock/mentoring.md
```diff
@@ -0,0 +1,118 @@
+## Things to watch out for
+
+- It is preferred to use \``value(m)`\` rather than \``m.value`\`. Fields are generally considered private in Julia. You can import \``value`\` from Dates.
```
SaschaMann, line 3: I think an explanation of why and when this is the case would be helpful. On first look, this sounds like one has to add Java levels of boilerplate without (m)any benefits.

tracks/julia/exercises/clock/mentoring.md
```diff
@@ -0,0 +1,118 @@
...
+    minutes::Int8
+    Clock(h,m) = new(rem(h + div(m,60,RoundDown),24,RoundDown),rem(m,60,RoundDown))
+end
+
+Base.:+(c::Clock,m::Dates.Minute) = Clock(c.hours, c.minutes + value(m))
+Base.:-(c::Clock,m::Dates.Minute) = Clock(c.hours, c.minutes - value(m))
+Base.:+(c::Clock,h::Dates.Hour) = Clock(c.hours + value(h), c.minutes)
+Base.:-(c::Clock,h::Dates.Hour) = Clock(c.hours - value(h), c.minutes)
+
+Base.show(io::IO, c::Clock) = print(io, '"', lpad(c.hours,2,'0'), ":", lpad(c.minutes,2,'0'), '"')
```
SaschaMann, line 116: Perhaps add a suggestion above that Printf is a nicer solution to this than lpad.

tracks/julia/exercises/clock/mentoring.md
cmcaine: That's fair, but I'm not sure I have a good explanation beyond the same reasons that you'd do it for Java or any other language. Can you think of one?

tracks/julia/exercises/clock/mentoring.md
cmcaine: I'm not sure I want to venture that opinion. I personally prefer it, but I think Scott prefers this with lpad. printf is its own weird little language.

tracks/julia/exercises/clock/mentoring.md
SaschaMann: In some cases it makes the code more generic. `value(x)` will work for any x as long as there's a method for it, while `x.value` will require the field. You won't reap the benefits of multiple dispatch if you always access fields directly.

tracks/julia/exercises/clock/mentoring.md
SaschaMann: Maybe "if you're already familiar with Printf in other languages, consider using..." or "consider taking a look at Printf as a way to format values in a string, it's more complex but also more flexible than lpad" as an optional suggestion/talking point?

tracks/julia/exercises/clock/mentoring.md
cmcaine: I feel like that's the same benefit as in Java :shrug: But we could restate it, I suppose for the scientists and novices who might not be familiar.

cmcaine:
I think both review comments are now addressed.

PR #1649, opened 2020-06-14
Julia: add mentoring notes for pangram

cmcaine:


ScottPJones:
When I've got a minute, I'll have to add some "what if you want more performant solution" notes.
The best I came up with used just a single UInt (in a CPU register) to keep track of whether any of A-Z and a-z had been seen, zeroing the bit when found, and checking for 0 to be able to exit early.

ScottPJones:
I could probably speed it up even more though, by looking at each byte, subtracting 0x40, checking if it was then > 0x3f, if not, masking by 0x1f (to deal with lower case), clearing the bit and checking for 0.  Might even be able to do that with only 1 branch (only taken once, if you get a pangram).

cmcaine:
That would be very welcome. I have a solution that uses a packed UInt, but I think it is slower and uglier than the solution with a byte array. Your sketch above sounds a little more sophisticated.

cmcaine:
IMO, this can be merged now, btw.

cmcaine:
No worries, that's what I thought :)

PR #1658, opened 2020-06-27
rust/poker: Add mentoring notes

happysalada:
Here is a first pass at adding mentoring notes for the rust poker exercise. The exercise is quite long (~200 loc) and there are many ways to proceed. Also I'm sure there are many more comments to be added.

Curious to have some feedback on this.

### DIFF
--- a/tracks/rust/exercises/poker/mentoring.md
+++ b/tracks/rust/exercises/poker/mentoring.md
@@ -3,7 +3,7 @@
 - Iterators
 - Iterator Methods
 - Pattern matching
-- Modelling a problem with the type system
+- Modeling a problem with the type system
 
 ### Reasonable Solutions
 
--- a/tracks/rust/exercises/poker/mentoring.md
+++ b/tracks/rust/exercises/poker/mentoring.md
@@ -1,0 +1,232 @@
+### Concepts
+
+- Iterators
+- Iterator Methods
+- Pattern matching
+- modelling a problem with the type system
+
+### Reasonable Solutions
+
+A reasonable solution should:
+
+- Use iterator whenever possible instead of for loops with conditionals inside
+- Not store the string reference given as an argument inside a data structure.
+- Represent the problem space extensively with the type system. Using u8 to represent cards is possible but not very readable.
+- **_optional_** use pattern matching on slices for the check if a hand is a straight of to check the card counts. The pattern matching on slice was stabilized on 1.42 [ref](https://blog.rust-lang.org/inside-rust/2020/03/04/recent-future-pattern-matching-improvements.html)
+
+### Examples
+
+```rust
+use std::cmp::Ordering;
+use std::collections::{HashMap, HashSet};
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+enum Suit {
+    Clover,
+    Diamond,
+    Spade,
+    Heart,
+}
+
+impl Suit {
+    pub fn new(raw: char) -> Suit {
+        match raw {
+            'C' => Suit::Clover,
+            'D' => Suit::Diamond,
+            'S' => Suit::Spade,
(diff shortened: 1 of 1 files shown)


tracks/rust/exercises/poker/mentoring.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Common suggestions
+
+#### No need to store a reference to the given string
+
+storing hand_str inside the Hand struct adds overhead and isn't necessary. It's better to filter on the given hands argument to return the winning hands.
+
+#### Using complex logic to determine the hand category
+
+a lot of students will go into 10-20 lines logic trying to determine the number of pairs, the number of triples... That logic can usually be replaced with collecting ranks into a hashmap and pattern macthing on it.
```
SleeplessByte, line 232:
```suggestion
Many students will go into 10-20 lines logic trying to determine the number of pairs, the number of triples. That logic can usually be replaced with collecting ranks into a `HashMap` and us pattern matching on it.
```

tracks/rust/exercises/poker/mentoring.md
```diff
@@ -0,0 +1,232 @@
...
+        .collect::<Vec<_>>()
+        .into()
+}
+```
+
+### Common suggestions
+
+#### No need to store a reference to the given string
+
+storing hand_str inside the Hand struct adds overhead and isn't necessary. It's better to filter on the given hands argument to return the winning hands.
```
SleeplessByte, line 228:
```suggestion
Storing `hand_str` inside the `Hand` struct adds overhead and isn't necessary. It's better to filter on the given hands argument to return the winning hands.
```

SleeplessByte: Perhaps you can add _why_ this is better.

tracks/rust/exercises/poker/mentoring.md
```diff
@@ -0,0 +1,232 @@
+### Concepts
+
+- Iterators
+- Iterator Methods
+- Pattern matching
+- modelling a problem with the type system
```
SleeplessByte, line 6:
```suggestion
- Modelling a problem with the type system
```

SleeplessByte (commented):
Few minor nits. Would like an approval from @exercism/website-copy-rust tho.

happysalada:
@SleeplessByte Thanks for taking the time to have a look!

SleeplessByte:
@exercism/reviewers can someone whomstve knows rust check this?

tracks/rust/exercises/poker/mentoring.md
```diff
@@ -0,0 +1,232 @@
+### Concepts
+
+- Iterators
+- Iterator Methods
+- Pattern matching
+- Modelling a problem with the type system
```
coriolinus, line 6:
```suggestion
- Modeling a problem with the type system
```

tracks/rust/exercises/poker/mentoring.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Reasonable Solutions
+
+A reasonable solution should:
+
+- Use iterator whenever possible instead of for loops with conditionals inside
```
coriolinus, line 12: This advice always makes me skeptical. Rust's iterators are powerful and nice, but it's not justified to claim that they should always be preferred over for loops. In particular, if contained code requires a bunch of `?` operators, for loops can be much simpler to use.

happysalada: In general I completely agree with you. However in this case, the emphasis is on the `conditionals inside` rather than having a for loop with some ifs and complicated logic, having iterators with combinators (.filter and the like) usually makes intent much clearer (IMO)

happysalada:
Thank you for taking the time to review!

PR #1676, opened 2020-07-08
Fix JS Pascal's Triangle Mentoring Notes

ghjk-code:
Fix mentoring notes for the Pascal's Triangle exercise to align to the new expected code (a single function instead of a class).
Linked to exercism/javascript#864.

### DIFF
--- a/tracks/javascript/exercises/pascals-triangle/mentoring.md
+++ b/tracks/javascript/exercises/pascals-triangle/mentoring.md
@@ -26,7 +26,7 @@ export const rows = size => {
   const result = [[1]];
 
   for (let y = 1; y < size; y++) {
-    let lastRow = [];
+    const lastRow = [];
 
     for (let x = 0; x <= y; x++) {
       const up = result[y - 1];
--- a/tracks/javascript/exercises/pascals-triangle/mentoring.md
+++ b/tracks/javascript/exercises/pascals-triangle/mentoring.md
@@ -19,8 +19,9 @@ export const rows = levels => {
 
 ```javascript
 export const rows = size => {
-  if (size === 0)
+  if (size === 0) {
     return [];
+  }
 
   const result = [[1]];
 
--- a/tracks/javascript/exercises/pascals-triangle/mentoring.md
+++ b/tracks/javascript/exercises/pascals-triangle/mentoring.md
@@ -1,47 +1,44 @@
 ### Reasonable solutions
 
 ```javascript
-function buildRow(result, element, index, array) {
-  return result.concat([element + (array[index + 1] || 0)])
+const buildRow = (result, element, index, array) => {
+  return [...result, element + (array[index + 1] || 0)];
 }
 
-function makeTriangle(levels) {
+export const rows = levels => {
+  if (levels === 0)
+    return [];
+
   const result = [[1]]
-  for (let x = 1; x < levels; x += 1) {
-    result.push(result[x - 1].reduce(buildRow, [1]))
+  for (let x = 1; x < levels; x++) {
+    result.push(result[x - 1].reduce(buildRow, [1]));
   }
-  return result
-}
-
-export class Triangle {
-  constructor(levels) {
-    this.rows = makeTriangle(levels)
-  }
-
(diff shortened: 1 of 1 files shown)


tracks/javascript/exercises/pascals-triangle/mentoring.md
```diff
@@ -1,47 +1,44 @@
 ### Reasonable solutions
 
 ```javascript
-function buildRow(result, element, index, array) {
-  return result.concat([element + (array[index + 1] || 0)])
+const buildRow = (result, element, index, array) => {
```
SleeplessByte, line 4:
```suggestion
const buildRow = (result, element, index, array) => {
```

This is _as good_ as using a `function declaration`. It's a good idea to mention below the solution that any way of writing a function is fine.

tracks/javascript/exercises/pascals-triangle/mentoring.md
```diff
@@ -1,47 +1,44 @@
...
-                const right = up[x] || 0
-
-                this.lastRow.push(left + right)
-            }
-            this.rows.push(this.lastRow)
-        }
+export const rows = size => {
+  if (size === 0)
+    return [];
+
```
SleeplessByte, line 24:
```suggestion
  if (size === 0) {
    return [];
  }

```

tracks/javascript/exercises/pascals-triangle/mentoring.md
```diff
@@ -1,47 +1,44 @@
...
+  const result = [[1]];
+
+  for (let y = 1; y < size; y++) {
+    let lastRow = [];
```
SleeplessByte, line 28:
```suggestion
    const lastRow = [];
```

tracks/javascript/exercises/pascals-triangle/mentoring.md
ghjk-code: Indeed, but here, the aim was to have a consistent style on this site where we only used arrow function in exercises templates...

And, is this the exercise to mention this point? This exercise comes really late and I guess that we (JS mentors) have already spoken about standard / arrow functions, don't you think so?

tracks/javascript/exercises/pascals-triangle/mentoring.md
ghjk-code: Yes, OK: I am using Guidelines which enforces braces simplification as much as we can, but it looks like it is not so frequent... I will adjust that!

tracks/javascript/exercises/pascals-triangle/mentoring.md
ghjk-code: Not sure I agree with that... Have you some rules / guidelines about that?

Personally, I use `const` when I don't want to change this variable (by reference or value)... This works more as an indicator that we should not touch this variable, this way I know that its value is the one at initialization...

What are your thoughts about that?

tracks/javascript/exercises/pascals-triangle/mentoring.md
SleeplessByte: I just mention it always, because some mentors start mentoring late exercises.

tracks/javascript/exercises/pascals-triangle/mentoring.md
SleeplessByte: The standard `eslint` configuration (`eslint/recommended`) will recommend to use a `const` when the _binding_ never changes. That in line with what `const` means in JavaScript context. 

The whole "but it confuses me because const doesn't mean constant" or "I want to indicate that a variable should never change" to me is just not a good idea and confuses people even more. The counter argument is that when you have a variable declaration on line 1, you won't know on line 50 anymore if you used `const` or `let`. If It's _vital_ that it doesn't change, `const` is going to help you (but `TypeScript` might, and `Object.freeze` will help).

SleeplessByte: If you change this, we can get this merged!

SleeplessByte (approved):

PR #1681, opened 2020-07-15
pages/how_do_mentor_queues_work: Is the estimate warranted?

elireiner:
In the mentor discussion of the page linked to below it says: the median waiting time for mentoring on this exercise is about 8 hours.
https://exercism.io/my/solutions/e0a9b0112945454a95ca527639092d46

### DIFF
--- a/pages/how_do_mentor_queues_work.md
+++ b/pages/how_do_mentor_queues_work.md
@@ -3,5 +3,5 @@ Normally, though, they work through either a track's queue (e.g. all the Ruby su
 
 Queues are ordered algorithmically, but fundamentally acts like first-in-first-out queues with core exercises prioritized over side exercises, and those submitted in "Mentored Mode" prioritized over those in "Practice Mode".
 
-Because there is no strict linear queue, and because mentors' rates of mentoring and mentoring behavior changes daily, we currently do not provide time estimates.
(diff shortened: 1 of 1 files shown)


SleeplessByte:
I think we actually _always_ provide estimates now. That page should be revised.

cc @exercism/product

pages/how_do_mentor_queues_work.md
```diff
@@ -3,5 +3,5 @@ Normally, though, they work through either a track's queue (e.g. all the Ruby su
 
 Queues are ordered algorithmically, but fundamentally acts like first-in-first-out queues with core exercises prioritized over side exercises, and those submitted in "Mentored Mode" prioritized over those in "Practice Mode".
 
-Because there is no strict linear queue, and because mentors' rates of mentoring and mentoring behavior changes daily, we currently do not provide time estimates.
+Because there is no strict linear queue, and because mentors' rates of mentoring and mentoring behavior changes daily, we currently do not always provide time estimates.
```
iHiD, line 6:
```suggestion
We provide time estimates for Core Exercises based on the median response time over recent weeks, and these can sometimes be inconsistent with the time you may wait. Because there is no strict linear queue, and because mentors' rates of mentoring and mentoring behavior changes daily, these estimates should be treated as our "best guess", rather than any guarantee.

iHiD:
Thanks. I've updated it to reflect where we currently stand.

ErikSchierboom (approved):
LGTM!

PR #1697, opened 2020-08-02
feat: draft isogram mentoring notes

micuffaro:


tracks/go/exercises/isogram/mentoring.md
```diff
@@ -0,0 +1,76 @@
...
+If the student goes this route we may suggest the `map[rune]bool` solution, or at the very least to add a comment clarifying their intent.
+
+**Checking characters**
+
+- Since the requirements only mention whitespace and hyphens, checking only for those is acceptable.
+- `unicode.IsLetter()` is otherwise a great alternative that will take care of all invalid/unwanted characters.
+
+### Talking points
+
+- The storage is being used as a [Set](<https://en.wikipedia.org/wiki/Set_(abstract_data_type)>), so some time can be spent with the student discussing pros and cons of this data structure.
```
bitfield, line 72:
```suggestion
- The storage is being used as a [Set](<https://en.wikipedia.org/wiki/Set_(abstract_data_type)>), so some time can be spent with the student discussing pros and cons of this data structure. We can also reference [Finding whether a Go map key exists](https://bitfieldconsulting.com/golang/map-key-exists) for some examples of using `bool` maps as sets.
```

tracks/go/exercises/isogram/mentoring.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+Working directly with code points this way is both a clever and efficient way to solve the problem, at the expense of readability.  
+If the student goes this route we may suggest the `map[rune]bool` solution, or at the very least to add a comment clarifying their intent.
+
+**Checking characters**
+
+- Since the requirements only mention whitespace and hyphens, checking only for those is acceptable.
+- `unicode.IsLetter()` is otherwise a great alternative that will take care of all invalid/unwanted characters.
+
```
bitfield, line 69: It's very common to see code like this:

```go
if unicode.IsLetter(v) {
	... // check map
}
```

This is a nice opportunity to point out the 'early return' pattern:

```go
if !unicode.IsLetter(v) {
	continue
}
... // check map
```

and explain its advantages:

1. The remaining code doesn't need to be indented, improving readability
2. No need for a subsequent `else` clause, because we already returned
3. It's easier to read non-negative `if` conditions, and the action associated with the condition (`continue`) is right next to the condition itself in the source code, instead of many lines later.

micuffaro: Thanks @bitfield , I addressed all your points.

tracks/go/exercises/isogram/mentoring.md
```diff
@@ -0,0 +1,109 @@
...
+```
+index := charVal - 'A'
+if alpha[index] {
+	return false
+}
+alpha[index] = true
+```
+
+Working directly with code points this way is both a clever and efficient way to solve the problem, at the expense of readability.  
+If the student goes this route we may suggest the `map[rune]bool` solution instead, and direct them to [Writing Slower Go Programs](https://bitfieldconsulting.com/golang/slower) for tips on when to trade off readability against performance (hint: hardly ever).
```
tehsphinx, line 68: Maybe also add that an index check should be done in this case. If there is a strange character in the string (which shouldn't happen but could happen), it would panic without an index range check.

tracks/go/exercises/isogram/mentoring.md
```diff
@@ -0,0 +1,109 @@
...
+if !unicode.IsLetter(v) {
+	continue
+}
+... // check map
+```
+
+and explain its advantages:
+
+- The remaining code doesn't need to be indented, improving readability
+- No need for a subsequent `else` clause, because we already returned
```
tehsphinx, line 97:
```suggestion
- No need for a subsequent `else` clause, because we already returned/continued
```

tracks/go/exercises/isogram/mentoring.md
tehsphinx: Note: IsLetter does not help here:

```go
func main() {
	fmt.Println('ä', unicode.IsLetter('ä'))
}
// Output: 228 true
```

micuffaro:
Sorry for the delay, I started a new job and had a busy couple days :)
I updated the doc with @tehsphinx 's suggestions. Would be cool to merge this so we can start using it.

PR #1701, opened 2020-08-08
java/difference-of-squares: Update mentoring.md for difference of squares

safarmer:
Added a simple solution that performs the calculation in a single iteration


SleeplessByte (approved):

SleeplessByte:
Thank you! Sorry this took a while, but looks great.

PR #1708, opened 2020-08-13
rust/sublist: Update for the solutions that call 2 time windows.

musitdev:
Some student have solution where they call 2 times windows instead of checking the size(with the list inverted). I thought there was a performance penalty but it's not the case. It's equivalent to the proposed solution because the windows function check the size.

### DIFF
--- a/tracks/rust/exercises/sublist/mentoring.md
+++ b/tracks/rust/exercises/sublist/mentoring.md
@@ -12,6 +12,7 @@ A reasonable solution should do the following:
   a larger list
 - Utilize Eq for slices
 - not reimplement the functionality of the `windows` method
+- some student call 2 time the windows function instead of checking the size. For the performance point of view it's the same because the windows function do the check.
 
 ### Examples
 


tracks/rust/exercises/sublist/mentoring.md
```diff
@@ -12,6 +12,7 @@ A reasonable solution should do the following:
   a larger list
 - Utilize Eq for slices
 - not reimplement the functionality of the `windows` method
+- some student call 2 time the windows function instead of checking the size. For the performance point of view it's the same because the windows function do the check.
```
coriolinus, line 15: Please rewrite for grammar / clarity.

musitdev: Is it better?

coriolinus: Not really. Reading that text, I still don't understand the following points:

- Is this point recommending one technique over the other?
- If it is recommending one, which is it recommending?
- How is an additional `windows` call equivalent to a `len` call? Naively, I'd expect `O(n)` in the former case, and `O(1)` in the latter.

SleeplessByte:
Closed because stale for 3 years

PR #1718, opened 2020-08-31
julia: pythagorean-triplet mentor notes

cmcaine:


cmcaine:
A student claims that it is always true that `b < c` for the later example, but I'm not convinced. I'm also not a great mathematician, so it could just be obvious :/

Opinion welcome.

tracks/julia/exercises/pythagorean-triplet/mentoring.md
```diff
@@ -0,0 +1,70 @@
...
+        end
+    end
+    return triplets
+end
+```
+````
+
+Telling Julia the type of `triplets` allows it to store the values without indirection. Without an annotation some (all?) version of Julia will type `triplets` as `Vector{Any}`, which means a vector of pointers to heap-allocated objects, which is slower.
+
+We could tighten up the bounds above, if we wanted:
```
cmcaine, line 32:
```suggestion
A small improvement is to tighten up the loop bounds using some simple arithmetic and ceiling division (`cld`):
```

tracks/julia/exercises/pythagorean-triplet/mentoring.md
```diff
@@ -0,0 +1,70 @@
...
+
+```
+    # Lower bound because the smallest triple is 3, 4, 5.
+    # Upper bounds implied by a < b < c && a + b + c == n.
+    for a in 3:cld(n, 3)
+        for b in a+1:cld(2n, 3)
+```
+
+A linear time algorithm derived by solving some simultaneous equations:
```
cmcaine, line 41:
```suggestion
This linear time algorithm derived by solving some simultaneous equations is a much greater improvement:
```

SaschaMann:
I might have a logic error in here somewhere, but I think that follows from how `b = ...` is derived. `b = c` and `b > c` are ruled out if `a^2 + b^2 = c^2` [see below]. Since we derive `b` from that equation, we know it holds true.

- Let b = c. Then a^2 + b^2 = c^2 <=> a^2 = 0 but we know that a != 0.
- Let b > c. a^2 + b^2 = c^2 <=> a^2 = c^2 - b^2. If b > c, a^2 < 0.

Does that make any sense? (I'm not sure :D )

SaschaMann:
Euclid's formula would be another solution: https://en.wikipedia.org/wiki/Pythagorean_triple#Proof_of_Euclid's_formula

cmcaine:
I think that makes sense!

cmcaine:
I think I tried to come up with something from euclid's formula but decided it wasn't very good. I'm not sure, though, I did this several months ago!

SaschaMann (approved):
LGTM. Thanks for writing these :)

tracks/julia/exercises/pythagorean-triplet/mentoring.md
```diff
@@ -0,0 +1,70 @@
...
+    for a in 3:cld(n, 3) - 1
+        # Derived by eliminating c from these simultaneous
+        # equations and solving for b:
+        #     a^2 + b^2 = c^2
+        #     a + b + c = n
+        b = (n^2 - 2n * a) / (2 * (n - a))
+        if a < b && isinteger(b)
+            c = n - a - b
+            # TODO: prove that b < c to remove this check.
+            b < c && push!(triplets, (a, b, c))
```
cmcaine, line 64:
```suggestion
            # Proof that b < c:
            # Let b = c. Then a^2 + b^2 = c^2 ⇔ a^2 = 0 but we know that a ≠ 0.
            # Let b > c. a^2 + b^2 = c^2 ⇔ a^2 = c^2 - b^2.
            #    If b > c, a^2 < 0, which is impossible for real numbers.
            push!(triplets, (a, b, c))
```

cmcaine:
@SaschaMann Can you re-approve this, I just added a note about that proof :)

SaschaMann (approved):

PR #1719, opened 2020-09-01
julia: nucleotide-count: clarify example #2

SaschaMann:


SaschaMann:
This makes it clearer what's going on there, since it's quite a dense expression.

cmcaine (approved):
Sure :)

PR #1722, opened 2020-09-02
Add n8henrie python bio

n8henrie:


mentors/python/n.json
```diff
@@ -54,5 +54,13 @@
     "link_url": null,
     "avatar_url": null,
     "bio": "I think Python is a great language for doing everyday tasks, improving mental models fundamental to computer science, and implementing really cool projects.  I use it every day, and encourage others to start thinking about how it can improve their lives."
+  },
+  {
+    "github_username": "n8henrie",
+    "name": "Nathan Henrie",
+    "link_text": "n8henrie.com",
```
rpottsoh, line 61:
```suggestion
    "link_text": null,
```
use `null` for link text since it is the same as `link_url`.

PR #1725, opened 2020-09-09
Julia: update nucleotide-count mentoring notes

cmcaine:
Some minor changes to the mentoring notes.


ErikSchierboom (approved):
I don't know Julia, but it looks good

cmcaine:
Thanks for reviewing while Sascha is busy :)

PR #1734, opened 2020-09-14
fixing community mentoring code for Bob exercise

nllerandi:
The following community provided code will not work as `message.length` is referencing the original untrimmed string, causing the 'ending with whitespace' test to fail (ie, `message` has not been trimmed yet).

`const isAsking = message => message.trim()[message.length - 1] === '?'`

Screen Shot 2020-09-14 at 5 12 27 PM


tejasbubane (approved):

tejasbubane:
Thanks @nllerandi 👍

PR #1746, opened 2020-09-24
ruby/series - Update mentoring.md

abinoam:
Add `each_char` over `chars`.

See: https://ruby-doc.org/core-2.7.1/String.html#method-i-chars

```
chars → an_array click to toggle source

Returns an array of characters in str. This is a shorthand for str.each_char.to_a.

If a block is given, which is a deprecated form, works the same as each_char.
```

### DIFF
--- a/tracks/ruby/exercises/series/mentoring.md
+++ b/tracks/ruby/exercises/series/mentoring.md
@@ -5,7 +5,6 @@ New concepts: Iterating over an array (preferably with `each_cons`, avoid `each_
 ```ruby
 
 class Series
-
   def initialize(numerals)
     @numerals = numerals.chars
   end
--- a/tracks/ruby/exercises/series/mentoring.md
+++ b/tracks/ruby/exercises/series/mentoring.md
@@ -26,6 +26,7 @@ end
     @numerals = numerals.chars
   ```
 - For more advanced students favor `each_char` over `chars`. See details at the "Talking points".
+- Ampersand syntax is a bonus point. `map(&:join)`.
 
 ### General 
 - The Instructions point to a beginner friendly [explanation of iterating in Ruby:](http://jeromedalbert.com/ruby-how-to-iterate-the-right-way/)
--- a/tracks/ruby/exercises/series/mentoring.md
+++ b/tracks/ruby/exercises/series/mentoring.md
@@ -25,6 +25,7 @@ end
   ```
     @numerals = numerals.chars
   ```
+- For more advanced students favor `each_char` over `chars`. See details at the "Talking points".
 
 ### General 
 - The Instructions point to a beginner friendly [explanation of iterating in Ruby:](http://jeromedalbert.com/ruby-how-to-iterate-the-right-way/)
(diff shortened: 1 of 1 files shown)


tracks/ruby/exercises/series/mentoring.md
```diff
@@ -33,12 +34,13 @@ and to the rubydocs [Enumerable](https://ruby-doc.org/core/Enumerable.html) with
 
 ### Talking points
 - `each_cons` instead of an iterator `with_index`: In Ruby, you rarely have to write iterators that need to keep track of the index. Enumerable has powerful methods that do that for us.
-- `chars`: instead of `split('')`   
+- `chars`: instead of `split('')`. Or even better, `each_char`.
```
kotp, line 37: This could be another bullet point:

```suggestion
- `chars`: instead of `split('')`.
- `each_char`: if an `Array` is not specifically necessary or wanted.
```

tracks/ruby/exercises/series/mentoring.md
```diff
@@ -33,12 +34,13 @@ and to the rubydocs [Enumerable](https://ruby-doc.org/core/Enumerable.html) with
...
-- `unless` , inline: With `unless` instead of `if`, we can show what "good" looks like for the conditional statement.
+- `unless`, inline: With `unless` instead of `if`, we can show what "good" looks like for the conditional statement.
 - `error`: Custom error message? (Only if the first submission meets the Minimal Solution.)
-- `map(&:join)`: instead of map with block, but at this point in the track it's okay to just accept it if students use it, no need to require it or dive into the subject of `Symbol#to_proc`  
+- `map(&:join)`: instead of map with block, but at this point in the track it's okay to just accept it if students use it, no need to require it or dive into the subject of `Symbol#to_proc`
+- `each_char` may be preferrable over `chars` for advanced students as it returns an `Enumerator` that just yields each char without creating an intermediate `Array`. More: `String#chars` with a block has a deprecations warning in more recent Ruby versions. `str.chars` is just a shorthand for `str.each_char.to_a`.
```
kotp, line 43:
```suggestion
- `each_char` may be preferable over `chars`  as it returns an `Enumerator` that will yield each character without creating an intermediate `Array`. More: `String#chars` with a block has a deprecation warning in more recent Ruby versions. `str.chars` is a shorthand for `str.each_char.to_a`.
```

kotp (changes requested):
There is  a single commit here, and it goes beyond what is stated as "the change" on line 52.

Should there be one commit that does precisely what line 52 suggests?  And another for commit (even if ti is in this PR) for the other concerns?

This means that if for some reason the incidental things are decided that we need to revert, we can easily, without disrupting what I think is good with what was listed on line 52.

abinoam:
Perfect.  Thanks for your feedback. I'll fix it.

PS: It's in one commit because I (lazily) crafted the PR directly at github.com website.

abinoam:
Hi @kotp,

I've tried to improve the PR based on your suggestions. Feel free to give any further suggestion if the PR still doesn't feel right (or I didn't understand it right).
I'm ok to change it many times as needed 👍

tracks/ruby/exercises/series/mentoring.md
```diff
@@ -33,12 +34,14 @@ and to the rubydocs [Enumerable](https://ruby-doc.org/core/Enumerable.html) with
...
-- `unless` , inline: With `unless` instead of `if`, we can show what "good" looks like for the conditional statement.
+- `unless`, inline: With `unless` instead of `if`, we can show what "good" looks like for the conditional statement.
 - `error`: Custom error message? (Only if the first submission meets the Minimal Solution.)
-- `map(&:join)`: instead of map with block, but at this point in the track it's okay to just accept it if students use it, no need to require it or dive into the subject of `Symbol#to_proc`  
+- `map(&:join)`: instead of map with block, but at this point in the track it's okay to just accept it if students use it, no need to require it or dive into the subject of `Symbol#to_proc`.
+- `each_char` may be preferable over `chars` as it returns an `Enumerator` that will yield each character without creating an intermediate `Array`. More: `String#chars` with a block has a deprecation warning in more recent Ruby versions. `str.chars` is a shorthand for `str.each_char.to_a`.
```
kotp, line 44:
```suggestion
- `each_char`: may be preferable over `chars` as it returns an `Enumerator` that will yield each character without creating an intermediate `Array`. More: `String#chars` with a block has a deprecation warning in more recent Ruby versions. `str.chars` is a shorthand for `str.each_char.to_a`.
```

kotp (changes requested):
Looks good to me, only change in this feedback is adding the colon to give the same style as the others in the list.

abinoam:
> Looks good to me, only change in this feedback is adding the colon to give the same style as the others in the list.

Fixed! 👍

kotp (approved):

kotp:
Thank you @abinoam !

PR #1748, opened 2020-09-27
Mentor notes Python/Leap: Added most common first attempt code

AyaanGH:
This is for the mentor sheet of the Python track - Leap year.

A quick reference to the most common correct solution for mentors to evaluate if the student's program is working as expected.

### DIFF
--- a/tracks/python/exercises/leap/mentoring.md
+++ b/tracks/python/exercises/leap/mentoring.md
@@ -1,3 +1,24 @@
+### Most common first attempt
+
+This is the most common solution by beginners to the leap year problem.
+Nudge the student towards the more concise solution by letting them know
+they can replace the 3 IF and 4 return statements with zero IF statements and a single return.
+This is done by using the logical operators OR & AND.
+
+```python
+def leap_year(year):
+    if year % 4 == 0: 
+        if year % 100 == 0: 
+            if year % 400 == 0: 
+                return True
+            else: 
+                return False
+        else: 
+             return True
+    else: 
+        return False
+```
+
 ### Reasonable solutions
 
 This is a nice, concise, very efficient solution:


tracks/python/exercises/leap/mentoring.md
```diff
@@ -1,3 +1,24 @@
+### Most common first attempt
+
+This is the most common solution by beginners to the leap year problem.
+Nudge the student towards the more concise solution by letting them know
+they can replace the 3 IF and 4 return statements with zero IF statements and a single return.
+This is done by using the logical operators OR & AND.
```
kotp, line 6: Use backticks and real operator names, maybe?

Also, is it true that this is the most common solution by beginners?  If that is the case, I don't have a problem with stating so.  If it hasn't been confirmed by actually counting, then I would feel uncomfortable claiming that, or presenting it, as fact.

kotp (changes requested):

tracks/python/exercises/leap/mentoring.md
AyaanGH: > Use backticks and real operator names, maybe?

I'm sorry, I don't understand what you mean? 


I have seen two students write virtually the same code as presented in this pull request, In fact, on my first attempt with the leap year problem, I arrived at the same code. I don't have the statistics to back up the claim, but given that the leap year problem focuses on control flow, I would assume that beginners would gravitate to solving this problem using multiple IF statements. 

That being said, I understand where you are coming from and I'm happy to reword the mentor guide to claim _it is one of the most common solutions_ rather than the definitive claim it _is the most common_.

kotp: For the backticks and real operator names, I am talking about doing this:

```
they can replace the 3 `if` and 4 `return` statements with zero `if` and a single `return`.
```

The reason is in part the consistency.  The if and the return words are mixed case, the "IF" and the "return" not being in the same style, so it appears there may be two different kinds of things. (And maybe it is appropriate.)  Hopefully that makes sense.  I mean to say that `if` is a real thing, as is `return` and being code, you can use the backtick to indicate this.

I might say "a common solution may be" and that would be indicative.

We could more simply say "if this solution is presented" without making the judgement of "beginner" or "many" or "most".  This may come from those beginning in the language, but well versed in a different programming language.  We don't care so much.  But we can give suggestions for this kind of solution.

SleeplessByte:
@AyaanGH if you're still interested, see Kotp's response above.

tracks/python/exercises/leap/mentoring.md
DanielNoord, line 6:
```suggestion
This is a common solution to the leap year problem. You can nudge the student towards
a more concise solution by letting them know that they can replace the third
`if ... else` block with a single `return year % 400 == 0` statement.
```

tracks/python/exercises/leap/mentoring.md
```diff
@@ -1,3 +1,24 @@
...
+
+```python
+def leap_year(year):
+    if year % 4 == 0: 
+        if year % 100 == 0: 
+            if year % 400 == 0:
```
DanielNoord, line 12:
```suggestion
            if year % 400 == 0:  # This can be improved upon.
```

DanielNoord:
@kotp I made suggestion to the original text. Not sure if you want the comment in the code to be added, but it might help identify what needs to be changed. Could you do a review of my proposed changes and if agreed commit them to the branch?

tracks/python/exercises/leap/mentoring.md
kotp: Why would we reverse the suggestion of not using if statements, as a mentor note?

I do agree with removing "most common solution by beginners" as we can not be sure they are beginners, and the "most common" can change over time, implying a maintenance requirement.

DanielNoord: Oh I misinterpreted their suggestion, I though they referred to the 3rd `IF`.

```python
def leap_year(year):
    return year % 4 == 0 and (year % 100 == 0 or year % 400 == 0)
 ```

I mean, is this actually better though? I know that after passing through the `Exercism` course myself the first time and then contributing to other projects I really had to adjust to not trying to fit everything on a single 80 character line. In terms of readability I do think you're losing somewhat here.
For this exercise it is probably fine, but by suggesting to write code like this beginners will probably think it is always "better" to write such one-liners (which I personally don't agree with).

Comment (kotp) [reply tracks/python/exercises/leap/mentoring.md:6]: It does not unduly check for 400.  It is not for the sake of "a single line" or "80 characters".

It is a mentor note, and so the mentor may not always suggest this, taking into consideration where students are based on the code and conversation that is had.

But yeah, this is not because of "one line", only an "understand the language and its constructs" note, and I would say that it is is up to the student to gain the experience to know when applying this and something like this is "better".

File: tracks/python/exercises/leap/mentoring.md (hunk 2 above)
Comment (kotp) [reply tracks/python/exercises/leap/mentoring.md:12]: How can this be improved upon?  I mean, if this can be improved upon, why is this not also mentioned at line 10 and 11 as they are identical in functionality to this statement.

Comment (DanielNoord) [reply tracks/python/exercises/leap/mentoring.md:12]: This was based on my previous suggestion when I thought the original suggestion was only to replace the third `if`.

Comment (kotp) [reply tracks/python/exercises/leap/mentoring.md:12]: Can probably mark it as resolved then?

File: tracks/python/exercises/leap/mentoring.md (hunk 1 above)
Comment (DanielNoord) [reply tracks/python/exercises/leap/mentoring.md:6]:
```suggestion
This is a common solution to the leap year problem. You can nudge the student towards
a more concise solution by letting them know that they can replace all `if` and `return`
statements with one `return` statement that concatenates all checks using `and` and `or`.
```

tracks/python/exercises/leap/mentoring.md
DanielNoord: Yes 😄

tracks/python/exercises/leap/mentoring.md
DanielNoord: @kotp Something like this then?

BethanyG:
Given the length of time this has been open, the merge conflicts, and the rather ... muddled ... discussion, I am going to close this PR. 

@AyaanGH and @DanielNoord -- should either of you want to submit a PR for future changes to the `leap` mentor notes, we can gladly review again - just please keep in mind the formatting notes @kotp put in above.  I'd also add that language like 

> _"most common solution by beginners"_

Is probably not ideal, and would suggest something  closer to what @DanielNoord suggested, that says 

> _"This is a common solution to the leap year problem."_

I also took a quick look at the [community solutions](https://exercism.org/tracks/python/exercises/leap/solutions?passed_head_tests=true) for `leap` that still pass the current tests.  

Many of them use a one-liner, but fail to employ short-circuiting to good advantage.  Or they use `if...else` but also fail to employ short-circuiting.  For comparison purposes, you can see the solution by user [bobahop](https://exercism.org/tracks/python/exercises/leap/solutions/bobahop) vs [my own much less efficient one](https://exercism.org/tracks/python/exercises/leap/solutions/BethanyG).  

So if mulit-line examples are introduced, I'd suggest focusing on examples that echo those same underlying logic issues already outlined in the notes, and maybe add a talking point about how some students may focus on a one liner because they feel it's "efficient" or "looks better", without understanding the order of evaluation issue.

PR #1764, opened 2020-09-30
please acept my pull request

darshanphotography:
please acsept my pull request


kotp:
Thank you for reading happens in the summary, the "title" line of a details block.  At the point of first reading it, it is very unlikely that it is read, and so the communication of thanking them for reading it seems premature.

However, it is probably a change that could be made while changing the tone of this document.  Does it need to happen?  Does it need to happen now?

I am not sure this change should happen.

Closing with that idea in mind, but I may be wrong.  We can always reopen the issue if it is decided this should come in later.

PR #1769, opened 2020-10-01
julia: expand nucleotide-count notes

cmcaine:


tracks/julia/exercises/nucleotide-count/mentoring.md
```diff
@@ -92,4 +86,15 @@ function count_nucleotides4(strand)
     result
 end
 ```
+
+Its speed comes from three sources:
+
+1. Iterating a vector of bytes instead of characters (though this algorithm is still unicode-safe).
+   This is faster because iterating characters from a string involves scanning the same vector of bytes but also doing a little work every iteration to establish if the current byte is the start of a multibyte sequence (and, if so to reassemble that sequence into its proper 32 bit character).
+2. The loop body is extremely simple and fast on x86 processors (a load, an addition and a store) and branchless (there are no conditionals within the loop body).
+   Branchless code is often faster because branching hinders CPU instruction level parallelism and stops compilers using "Single Instruction Multiple Data" (SIMD) instructions (though this loop body can't benefit from SIMD anyway).
```
SaschaMann, line 95: Perhaps add a note on how one can see that using `@code_native`?

SaschaMann (approved):
Great explanation :)

tracks/julia/exercises/nucleotide-count/mentoring.md
cmcaine: A dangerous thing to ask when I'm procrastinating.

SaschaMann (approved):
Oh wow, that's much more detailed than I expected. Great work!

SaschaMann (approved):

tracks/julia/exercises/nucleotide-count/mentoring.md
```diff
@@ -92,4 +86,98 @@ function count_nucleotides4(strand)
...
+
+(I used `code_native(count_nucleotides4, (String,); syntax=:intel)` which you could read as "Show me the native code produced when compiling `count_nucleotides4` with the argument types `(String,)` (i.e. one argument that is a `String`), and show me it in Intel's assembly syntax". You could also use `@code_native count_nucleotides4("foo")` and you'll get the same stuff but in the AT&T assembly format.)
+
+```nasm
+L320:
+        movzx   edx, byte ptr [r14 + rcx]
+        ; Load the address of counts (I don't know why this happens
+        ; repeatedly, but whatever).
+        mov     rdi, qword ptr [rax]
+        inc     qword ptr [rdi + 8*rdx]
```
cmcaine, line 142: @bergjohan If you have some time, I'd appreciate it if you could consider this question briefly? Maybe there's an obvious answer that I am ignorant of!

Do you know if there's a reason for this bit `mov     rdi, qword ptr [rax]`. I think that's loading the address of `rax` into `rdi`, but I don't understand why it would be done inside a loop where it doesn't look like `rax` is getting modified?

bergjohan: Wow, cool feature, being able to output assembly like this!

Yeah, I don't see any reason why we wouldn't be able to move the `mov rdi, qword ptr [rax]` line outside of the loop, instead of executing it in every iteration.

It seems like `rax` is a pointer to the `counts` array, so we're loading 8 bytes from the address that `rax` points to.
In C this would look like: https://godbolt.org/z/3344zv

cmcaine: Yeah, it's a nice feature! Thanks for the second opinion <3

PR #1780, opened 2020-10-02
julia: secret-handshake: minor improvements

cmcaine:


tracks/julia/exercises/secret-handshake/mentoring.md
```diff
@@ -8,6 +8,23 @@
...
 
 ````
+pdb5627's solution (slightly edited)
+
+This short and simple solution is fast, extremely easy to read, and clearly demonstrates the bitwise operations that most solutions use.
+Four of the lines are obviously very similar to each other and it's totally fine to write code that is "unrolled" like this.
+
+```julia
+function secret_handshake(n)
+    s = []
```
SaschaMann, line 18:
```suggestion
    s = String[]
```

SaschaMann (approved):
LGTM. Sorry I must've missed the first notification for the team

tracks/julia/exercises/secret-handshake/mentoring.md
cmcaine: I deliberately left this off, but I suppose it's not so scary.

SaschaMann (approved):

PR #1782, opened 2020-10-06
julia: add triangle mentoring notes

cmcaine:


cmcaine:
I don't know if I should drop the second example by me and just use the one by Nosferican :/

Probably not worth thinking about given how rarely this exercise comes up!

tracks/julia/exercises/triangle/mentoring.md
```diff
@@ -0,0 +1,117 @@
...
+end
+
+
+function is_triangle(sides)
+    (l1,l2,l3) = sort(sides)
+    return l1 > 0 && l1+l2>l3
+end
+```
+
+xyshell eliminated the double sort to get okay performance, which is neat:
```
cmcaine, line 97:
```suggestion
xyshell eliminated the double sort to get okay performance, which is neat.
But now `is_equilateral` mutates its input argument and doesn't have a `!` at the end of its name, so that's a bit dangerous:
other Julians might use this code under the assumption that the input will not be mutated and then get strange bugs.
```

ErikSchierboom (commented):
I think there is some quoting error somewhere in the source: https://github.com/exercism/website-copy/blob/d399fad87038d8c5793ac0946f8f549052565564/tracks/julia/exercises/triangle/mentoring.md

ErikSchierboom (approved):
Okay!

cmcaine:
Thanks!

PR #1785, opened 2020-10-07
julia: improve pythagorean-triplets notes

cmcaine:


tracks/julia/exercises/pythagorean-triplet/mentoring.md
```diff
@@ -5,14 +5,15 @@
 
 ## Example solutions
 
-A O(n^2) time algorithm:
-
 ````
+A O(n²) time algorithm:
```
SaschaMann, line 9: Since you started here, do you feel like search/replace-ing all ^2 in comments with `²`? :D

SaschaMann (approved):
Looks good!

cmcaine:
there's a small bug here, ill fix it soon

cmcaine:
I forgot about this PR and fixing that bug :) Done now.

cmcaine:
I think there's a better bound for this, btw.

A student used the code:

```jl
for a in 1:floor(Int, (1-sqrt(2)/2)*N)
```

But I couldn't work out how to derive that because I am not very good at maths :)

tracks/julia/exercises/pythagorean-triplet/mentoring.md
```diff
@@ -27,18 +56,35 @@ function pythagorean_triplets(n)
...
+> Note: you can come up with even tighter bounds than these. Share your working with your mentor if you do!
+
+The intuition on upper bounds is that if we were working with real numbers, the biggest `a` can be is `n/3 - ε`,
+where `ε` is the smallest real number,
+because `b` and `c` must both be larger.
+Similarly, the biggest `b` can be is `n/2 - ε` because `c` must be larger than it.
+
+When we move that to the Integers, we observe that the maximum remainder of `n/3` is 2.
+We already know that the biggest `a` can be is  when `a`, `b`, and `c` are as close as possible, so that gives us something like this: `fld(n, 3) - 1 < fld(n, 3) + x < fld(n, 3) + y` where `x < y && -1 + x + y < 2`.
+Similarly, the biggest remainder of `n/2` is 1, so the largest `b` can be is when our variables are something like `1 < fld(n, 2) - 1 < fld(n, 2) + 1`, so an upper bound on `b` is `fld(n, 2) - 1`.
```
cmcaine, line 83:
```suggestion
Similarly, the biggest remainder of `n/2` is 1, so the largest `b` can be is when our variables are `1 < fld(n, 2) - 1 < fld(n, 2) + 1`, so an upper bound on `b` is `fld(n, 2) - 1`.
```

tracks/julia/exercises/pythagorean-triplet/mentoring.md
```diff
@@ -25,34 +55,51 @@ function pythagorean_triplets(n)
...
 end
 ```
-````
 
-Telling Julia the type of `triplets` allows it to store the values without indirection. Without an annotation some (all?) version of Julia will type `triplets` as `Vector{Any}`, which means a vector of pointers to heap-allocated objects, which is slower.
+Calculating c from a and b clearly eliminates a huge number of iterations and is the main reason for the speedup.
 
-A small improvement is to tighten up the loop bounds using some simple arithmetic and ceiling division (`cld`):
+A much more modest improvement is achieved by telling Julia the type of `triplets` so that it can store the values without indirection.
+Without an annotation some (all?) version of Julia will type `triplets` as `Vector{Any}`, which means a vector of pointers to heap-allocated objects, which is slower.
```
SaschaMann, line 62: All, I think. That said, the speed-up in this case is incredibly small for some reason:

`112.999 μs (58 allocations: 2.77 KiB)` vs `111.599 μs (6 allocations: 3.17 KiB)` for the cubic solution and n=100.

tracks/julia/exercises/pythagorean-triplet/mentoring.md
```diff
@@ -25,34 +55,51 @@ function pythagorean_triplets(n)
...
-        for b in a+1:cld(2n, 3)
+    for a in 3:fld(n, 3)-1
+        for b in a+1:fld(n, 2)-1
 ```
 
-This linear time algorithm derived by solving some simultaneous equations is a much greater improvement:
+> Note: you can come up with even tighter bounds than these. Share your working with your mentor if you do!
+
+The intuition on upper bounds is that if we were working with real numbers, the biggest `a` can be is `n/3 - ε`,
+where `ε` is the smallest real number,
```
SaschaMann, line 77: This is a bit misleading, I think. The smallest real number is 0 (unless we're considering negatives). Maybe

```suggestion
where `ε` is the smallest non-zero real number,
```

if you want to avoid the usual "Let ε > 0, ..."?

tracks/julia/exercises/pythagorean-triplet/mentoring.md
```diff
@@ -25,34 +55,51 @@ function pythagorean_triplets(n)
...
+because `b` and `c` must both be larger.
+Similarly, the biggest `b` can be is `n/2 - ε` because `c` must be larger than it.
+
+When we move that to the Integers, we observe that the maximum remainder of `n/3` is 2.
```
SaschaMann, line 81:
```suggestion
When we move that to the integers, we observe that the maximum remainder of `n/3` is 2.
```

tracks/julia/exercises/pythagorean-triplet/mentoring.md
```diff
@@ -25,34 +55,51 @@ function pythagorean_triplets(n)
...
 ```
-````
 
-Telling Julia the type of `triplets` allows it to store the values without indirection. Without an annotation some (all?) version of Julia will type `triplets` as `Vector{Any}`, which means a vector of pointers to heap-allocated objects, which is slower.
+Calculating c from a and b clearly eliminates a huge number of iterations and is the main reason for the speedup.
 
-A small improvement is to tighten up the loop bounds using some simple arithmetic and ceiling division (`cld`):
+A much more modest improvement is achieved by telling Julia the type of `triplets` so that it can store the values without indirection.
+Without an annotation some (all?) version of Julia will type `triplets` as `Vector{Any}`, which means a vector of pointers to heap-allocated objects, which is slower.
+Annotating with an immutable type like we have done (which by definition will be of known size) allows Julia to simply store the items one after the other without indirection.
```
SaschaMann, line 63: Another speed-up can be achieved by using `sizehint!`. However, I'm not sure if there's a formula to estimate the size in advance without overshooting too much. Do you know a way?

SaschaMann (approved):
Looks good, left some minor comments.

As a side note, I just saw https://en.wikipedia.org/wiki/File:Pythagorean_Triples_Scatter_Plot.png - Could be a fun unmentored bonus task to plot one's solution

tracks/julia/exercises/pythagorean-triplet/mentoring.md
```diff
@@ -5,14 +5,44 @@
 
 ## Example solutions
 
-A O(n^2) time algorithm:
-
 ````
+There are three reasonably common solutions to this problem, a cubic time solution, a quadratic time solution and a linear time solution.
```
SaschaMann, line 9: Could you perhaps add a TL;DR of what "cubic time" etc mean? The terms may not be familiar to people without a CS/num. maths background.

SaschaMann: (This could be a follow-up PR, or some general mentoring note unspecific to this exercise)

cmcaine: Do you think O(n^3), O(n^2), etc is clearer? Could just use those.

tracks/julia/exercises/pythagorean-triplet/mentoring.md
cmcaine: I think the amount of time spent mallocing tuples will be negligible in all cases except maybe the linear one. I should probably change this to explain that this is nice to do anyway so that later code infers better. Might be a bit complicated?

tracks/julia/exercises/pythagorean-triplet/mentoring.md
cmcaine: Good catch. I'll use:

```suggestion
where `ε` is the smallest real number greater than 0,
```

tracks/julia/exercises/pythagorean-triplet/mentoring.md
SaschaMann: Maybe. I think it might be a bit complicated.

Maybe some general note like `It's best to avoid Any-type arrays whenever possible for performance reasons. You can find more details about why that is in the [Performance Tips](https://docs.julialang.org/en/v1/manual/performance-tips/index.html#man-performance-abstract-container)`?

tracks/julia/exercises/pythagorean-triplet/mentoring.md
SaschaMann: I think the same applies to those notations. Not fully sure though

SaschaMann (dismissed):
Feel free to address or ignore the other comments and then merge, this is good as it is :) Thanks for writing these detailed notes, they're really helpful when mentoring!

tracks/julia/exercises/pythagorean-triplet/mentoring.md
cmcaine: Even in the linear case the number of appends is tiny compared to the number of other operations, so I think it just doesn't matter. I'll mention that giving a more descriptive type annotation will help inference in functions that are consuming the output of this function.

cmcaine:
I think all the comments are resolved now. Thanks for the feedback, this is improved!

SaschaMann (approved):
Nice, I think this works well

PR #1790, opened 2020-10-16
Add mentoring notes for Dart Leap exercise

Corentin-Leffy:
Hello 👋 

This PR initializes mentoring notes for the Leap exercise in Dart to help mentors during their reviews. I started from the F# one and made some adjustments to make it conform to Dart requirements.

Thanks 🥰

### DIFF
--- a/tracks/dart/exercises/leap/mentoring.md
+++ b/tracks/dart/exercises/leap/mentoring.md
@@ -1,0 +1,36 @@
+### Reasonable solutions
+
+#### Single expression
+
+```dart
+class Leap {
+  bool leapYear(int year) =>
+      year % 4 == 0 && year % 100 != 0 || year % 400 == 0;
+}
+```
+
+#### Using a helper function
+
+```dart
+class Leap {
+  bool leapYear(int year) {
+    final isDivisibleBy = (int n) => year % n == 0;
+
+    return isDivisibleBy(4) && !isDivisibleBy(100) || isDivisibleBy(400);
+  }
+}
+```    
+
+### Common suggestions
+
+- Instead of using several, potentially nested if-statement, suggest to remove the if-statements by returning a single expression using the boolean `&&` and `||` operators.
+- Instead of explicitly returning `true` or `false` (e.g. `if x then true else false`), suggest to simply return the boolean condition (`x`).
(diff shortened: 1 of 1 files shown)


tracks/dart/exercises/leap/mentoring.md
```diff
@@ -0,0 +1,36 @@
...
+- Instead of explicitly returning `true` or `false` (e.g. `if x then true else false`), suggest to simply return the boolean condition (`x`).
+- If students are doing more than three checks in their if-statement, ask if they could get it down to just three checks. You could give them a hint to following the instructions in the README : 
+```
+a year MUST be a multiple of four AND NOT be a multiple of 100
+a year MUST be a multiple of 400 (implies multiple of 4 and 100)
+```
+
+### Talking points
+
+- Students often use parentheses, but as the `&&` operator has a higher precedence than the `||` operator, this is not necessary. You can check that in the [Dart Lang Specification](https://dart.dev/guides/language/specifications/DartLangSpec-v2.2.pdf) at section **20.2 - Operator Precedence**.
```
devkabiir, line 36: I believe Explicit is better than Implicit. Although due to operator precedence parantheses are not required, I prefer having them than not, for the next person who looks at the code, they might have a different background and might not know dart's operator precedence off the top of their head. Explicit parantheses reduces confusion for the next person. I don't know if and how my point of view should be added/phrased in these notes. So I leave this as a comment only.

devkabiir (approved):

tracks/dart/exercises/leap/mentoring.md
Corentin-Leffy: I understand your point of view and totally agree with you about **explicit** over **implicit**. 

I don't know if there is different operator precedence for other language. For example, Dart operator precedence is the same as Java one.

Also, explicit parentheses can add so much noise to a condition that the condition can be hard to read. 

I'm OK with this text as a comment, but if you find a way to phrased it, I'll gladly add it to these notes 👍

devkabiir:
Am I to merge this myself? I seem to have the merging rights.

rpottsoh:
You may if you like.

Stargator:
@devkabiir Thanks for all your work to keep an eye on Dart stuff. I'm slowly coming back onboard after a LONG year.

PR #1793, opened 2020-10-18
julia: minimal mentor notes for run-length-encoding

cmcaine:
This doesn't get submitted often, so I cba writing loads of notes right
now :)

### DIFF
--- a/tracks/julia/exercises/run-length-encoding/mentoring.md
+++ b/tracks/julia/exercises/run-length-encoding/mentoring.md
@@ -1,7 +1,7 @@
 ## Common comments
 
 - You might not know that repeatedly concatenating strings like this causes a lot of strings to be allocated (they are not appended in place). If you want to avoid that you can use an `IOBuffer` or (less efficiently) a vector of `Char`.
-- Using `isnumeric` is safe here because the input is defined quite restrictively, but in general it is not a good idea if what you really mean is `'0' <= c <= '9'` because `isnumeric` matches a lot more characters and is a lot slower. `isdigit` is a safe alternative.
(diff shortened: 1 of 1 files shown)


tracks/julia/exercises/run-length-encoding/mentoring.md
```diff
@@ -0,0 +1,108 @@
+## Common comments
+
+- You might not know that repeatedly concatenating strings like this causes a lot of strings to be allocated (they are not appended in place). If you want to avoid that you can use an `IOBuffer` or (less efficiently) a vector of `Char`.
+- Using `isnumeric` is safe here because the input is defined quite restrictively, but in general it is not a good idea if what you really mean is `'0' <= c <= '9'` because `isnumeric` matches a lot more characters and is a lot slower. `isdigit` is a safe alternative.
```
SaschaMann, line 4:
```suggestion
- Using `isnumeric` is safe here because the input is defined quite restrictively, but in general it is not a good idea if what you really mean is `'0' <= c <= '9'` because `isnumeric` matches a lot more characters, e.g. `¾` or `௰`, and is a lot slower. `isdigit` is a safe alternative.
```

Might help to clarify because it's not immediately obvious what a numeric character is other than 0-9. These examples are from the docs.

tracks/julia/exercises/run-length-encoding/mentoring.md
```diff
@@ -0,0 +1,108 @@
...
+
+
+## Example solutions
```
SaschaMann, line 7: Do you think we should include these even without commentary? One can look up community solutions on the side already, so I'm not sure this adds much?

SaschaMann (approved):
Thanks!

Left two comments but it's up to you what to do with them, feel free to merge it as it is, too.

tracks/julia/exercises/run-length-encoding/mentoring.md
cmcaine: Lots of community solutions are poor, so I think it's worth providing good examples

SaschaMann (approved):

SaschaMann:
Thanks!

PR #1802, opened 2020-10-26
rust/clock: chrono crate

bobahop:
Suggest using the chrono crate that was introduced in the Gigasecond exercise.


SleeplessByte (approved):
Looks good to me. I close-reopened this to make the CI run.

PR #1811, opened 2020-10-30
Update mentoring.md

alkhatib-shopify:


kotp:
While I like the solution with `tally` I don't know if this is the "minimal approvable solution".  It is, however a good talking point; "Can the new method `tally` do much of this work for us?"

kotp (changes requested):
While we are here we should change the class name from WordCount to Phrase, I think.

alkhatib-shopify:
> While I like the solution with `tally` I don't know if this is the "minimal approvable solution". It is, however a good talking point; "Can the new method `tally` do much of this work for us?"

Wouldn't this be similar to the Matrix solution? 
https://github.com/exercism/website-copy/blame/master/tracks/ruby/exercises/matrix/mentoring.md#L8

Since ruby now offers a `.tally` shouldn't we encourage people to use it rather than implement their own?

SleeplessByte:
@kotp did you miss @alkhatib-shopify 's response?

kotp:
> @kotp did you miss @alkhatib-shopify 's response?

Yes, I did.

I think we should encourage it, but until Ruby 2.7 is the target version, it doesn't make sense to bring it up to the front of the class yet, so to speak.   If the target is 2.7+ then it is fine, of course.

alkhatib-shopify:
I’ll reopen the PR from my personal GitHub.

SleeplessByte:
Great @alkhatib-shopify : so long story short, we like it, and we'll merge it if you do.

PR #1815, opened 2020-11-01
Update mentoring.md

MrGeek1337:
removed unused imports and replaced unnecessary for loop with single assignment


cmccandless (approved):
Looks good to me

PR #1819, opened 2020-11-07

cmcaine:


SaschaMann (approved):
Do you happen to know any good resources on learning regex that could be added as a link? Might be useful as talking point in case someone ask for help.

cmcaine:
I don't know of a particularly good resource off the top of my head. Would be a good enhancement to find one, tho.

cmcaine:
MDN probably has a good guide? And Regex Golf is a good revision / exercise tool: https://alf.nu/RegexGolf

And interactive tools like this are probably good, but I haven't reviewed it https://regexone.com/

tracks/julia/exercises/word-count/mentoring.md
```diff
@@ -0,0 +1,124 @@
...
+        end
+    end
+    return counts
+end
+```
+
+This solution uses a regex, but you could equally do something like:
+
+```julia
+split(input, c -> !(isalnum(c) || c == '''); keepempty=false)
```
tomerarnon, line 53: Where are you getting the `isalnum` function?

Also this isn't quite right, because splitting on `'` is not desired for a contraction. This solution (which of course can be `countmap`-ed manually if desired) splits only on `isspace`, and then strips punctuation separately (only removing punct chars on the edges of words). Filtering for empty strings has to happen after that anyway in case an empty string was formed by stripping punctuation.

```julia
wordcount(sent) = countmap(filter!(!isempty, strip.(ispunct, split(lowercase(sent)))))
```

SaschaMann: Seems like `isalnum` has been deprecated, which I wasn't aware of: https://discourse.julialang.org/t/isalnum-deprecated/8504

tracks/julia/exercises/word-count/mentoring.md
```diff
@@ -0,0 +1,124 @@
...
+This is a straightforward algorithm and it is easy to prove that it follows the specification exactly.
+Implementations that split by whitespace and punctuation tend to count words that the specification does not ask for, e.g. "x1".
+You can view that as a good thing or a bad thing!
+For what it is worth, traditional word counting code like `wc` treat a mix of letters and digits as a word.
+
+```julia
+function wordcount(input)
+    counts = Dict{String, Int}()
+    # Unicode version matching any letters or numbers: r"\pL+'\pL|\pL+|\pN+"
+    for m in eachmatch(r"[A-Za-z]+'[A-Za-z]|[A-Za-z]+|[0-9]+", input)
```
tomerarnon, line 69: This one fails on contractions like `we're`

```julia
julia> wordcount2("I said 123 and we're done")
Dict{String,Int64} with 7 entries:
  "e"    => 1
  "done" => 1
  "said" => 1
  "123"  => 1
  "we'r" => 1
  "and"  => 1
  "i"    => 1
```

tracks/julia/exercises/word-count/mentoring.md
cmcaine: Yeah, I should have checked that.

I think `isletter(c) || isdigit(c) || c == '''` should be used.

The idea is to replicate the regex pretty closely, so I don't want to show your method here.

Should maybe show it later, tho? Would need to introduce `countmap`, too.

cmcaine: The regex also includes underscore, which isn't desirable. Maybe we should mention that

tracks/julia/exercises/word-count/mentoring.md
cmcaine: Huh. They pass the tests. I should have written my own to check.

cmcaine: Oh, there should just be an extra plus. I missed it off when transcribing, I think.

cmcaine: Thanks!

cmcaine:
Fixed issues in #1821

PR #1820, opened 2020-11-08

tomerarnon:


tracks/julia/exercises/word-count/mentoring.md
```diff
@@ -121,4 +121,94 @@ function wordcount(sentence)
...
+    d = Dict{eltype(words), Int}()
+    for w in words
+        d[w] = get(d, w, 0) + 1
+    end
+    d
+end
+```
+It's possible to create the dictionary more efficiently with the `countmap` function from `StatsBase` (which only hashes the key once per check). With countmap, you can code-golf the above into:
+```julia
+wordcount_withstatsbase(sentence) = countmap(filter!(!isempty, strip.(ispunct, split(lowercase(sentence)))))
```
cmcaine, line 140:
```suggestion
using StatsBase: countmap

wordcount_withstatsbase(sentence) = countmap(filter!(!isempty, strip.(ispunct, split(lowercase(sentence)))))
```

tracks/julia/exercises/word-count/mentoring.md
```diff
@@ -121,4 +121,94 @@ function wordcount(sentence)
...
+
+
+** Totally manual ** 
+
+This method iterates through the string one character at a time and switches behavior at word boundaries (when `iswordchar` changes from true to false or false to true). The dictionary has `SubString` keys in this version, which are like the string-version of Array `views`.
+```julia
+function wordcount(sent)
+
+    sent = lowercase(sent)
+    d = Dict{SubString, Int}()
```
cmcaine, line 151: Should this be SubString{String} or is that the default?

tracks/julia/exercises/word-count/mentoring.md
```diff
@@ -121,4 +121,94 @@ function wordcount(sentence)
...
+
+    in_word = false; j = 0
+    # For our purposes, almost everything is a valid word character.
+    iswordchar(c) = !ispunct(c) && !isspace(c)
+    for i in eachindex(sent)
```
cmcaine, line 156: Because of the lookahead this might be better with a while loop?

Lookahead is only on ', so I think either is fine.

tracks/julia/exercises/word-count/mentoring.md
```diff
@@ -121,4 +121,94 @@ function wordcount(sentence)
...
+        c = sent[i]
+        if iswordchar(c) && i < lastindex(sent)
+            # if we're not already in a word, then this character marks
+            # the start of a new word. If we are in a word, just keep going.
+            if !in_word
+                in_word = true
+                j = i
+            end
+        elseif in_word
```
cmcaine, line 165: continue and if might be a little clearer than elseif?

I had to look twice seeing in_word set and then tested.

tracks/julia/exercises/word-count/mentoring.md
```diff
@@ -121,4 +121,94 @@ function wordcount(sentence)
...
+            in_word = false
+            k = SubString(sent, j:(i-1))
+            d[k] = get(d, k, 0) + 1
+        end
+    end
+    d
+end
+```
+
+### Benchmark comparison in julia v1.5.1
```
cmcaine, line 182: Not sure about including benchmarking code. If we do, we should use bold for headers and make it reproducible with the earlier examples (i.e. they should be renamed wordcount1, wordcount2, etc).

tracks/julia/exercises/word-count/mentoring.md
cmcaine:
```suggestion
    d = Dict{SubString{typeof(sent)}, Int}()
```

```julia
julia> view("hello", 1:2) |> typeof
SubString{String}
```

cmcaine: Maybe `sent` should be `input` for consistency, too.

tracks/julia/exercises/word-count/mentoring.md
tomerarnon: Writing it as a while loop is a bit of a hassle because of `nextind`.  I don't want to assume strictly ascii inputs. It' could be done but I don't think it would end up clearer.

tracks/julia/exercises/word-count/mentoring.md
tomerarnon: The `view` version only works on v1.6 I think. At least it doesn't for me on 1.2 and 1.5

I agree `input` would be preferable. I'll change it.

tracks/julia/exercises/word-count/mentoring.md
tomerarnon: Yes, fair enough. What is your preference here, omit the benchmark, or rename the functions?

tracks/julia/exercises/word-count/mentoring.md
tomerarnon: I'm assuming this is what you meant. I'm not sure it's necessarily more readable, since the second `if` only happens if the first doesn't, and the `continue` may be hard to miss. I'll give it a minute and then look at it again.

```julia
function wordcount6_4(input)

    input = lowercase(input)
    d = Dict{SubString{String}, Int}()

    in_word = false; j = 0
    # For our purposes, almost everything is a valid word character.
    iswordchar(c) = !ispunct(c) && !isspace(c)
    for i in eachindex(input)
        c = input[i]
        
        if iswordchar(c)
            # if we're not already in a word, then this character marks
            # the start of a new word. If we are in a word, just keep going.
            if i < lastindex(input) && !in_word
                in_word = true
                j = i
            end
            continue
        end

    
        if in_word
            # If `c` is not a word character, but we are in a word, then the
            # previous character marks the end of a word. However, if the current
            # character is a ' we need to see if this word is a contraction by
            # looking at the next character
            if c == '''
                i < lastindex(input) &&
                iswordchar(input[nextind(input, i)]) &&
                continue
            end

            in_word = false
            k = SubString(input, j:(i-1))
            d[k] = get(d, k, 0) + 1
        end
    end
    d
end
```

tracks/julia/exercises/word-count/mentoring.md
tomerarnon: Also, it should be `SubString{String}`! Thanks for that correction, I forgot to respond to it!

tracks/julia/exercises/word-count/mentoring.md
cmcaine: Cool. no need then.

tracks/julia/exercises/word-count/mentoring.md
cmcaine: I think I prefer this version, but please choose whichever you prefer, this was just a suggestion :)

tracks/julia/exercises/word-count/mentoring.md
cmcaine: For the record, I said this on slack:

> I think we should just omit it. Users can run their own benchmarks.
> If solutions have substantially different run times we should report that in the prose above the relevant solution.

tracks/julia/exercises/word-count/mentoring.md
```diff
@@ -121,4 +121,75 @@ function wordcount(sentence)
...
+        # for a lookahead and we need to record the last word.
+        if i == lastindex(input)
+            j = in_word ? j : i
+            i = iswordchar(c) ? i : i-1
+            if j <= i
+                k = SubString(input, j:i)
+                d[k] = get(d, k, 0) + 1
+            end
+            break
+        end
```
tomerarnon, line 169: I ran into an annoying edge case with final-character words. This takes care of that, but isn't too pretty.

tracks/julia/exercises/word-count/mentoring.md
```diff
@@ -121,4 +121,78 @@ function wordcount(sentence)
...
+    input = lowercase(input)
+    d = Dict{SubString{typeof(input)}, Int}()
+    iswordchar(c) = !ispunct(c) && !isspace(c)
+
+    in_word = false
+    j = 0
+    for i in eachindex(input)
+        c = input[i]
+
+        if i == lastindex(input)
```
cmcaine, line 160: You mentioned not liking this. Here's a `do while` loop that moves the lastindex check all the way to the end:

```julia
function wordcount(input)
    input = lowercase(input)
    d = Dict{SubString{typeof(input)}, Int}()
    iswordchar(c) = !ispunct(c) && !isspace(c)

    isempty(input) && return d

    in_word = false
    start_of_word = 0
    prev = 1
    idx = 1
    while true
        chr = input[idx]
        if iswordchar(chr)
            if !in_word
                in_word = true
                start_of_word = idx
            end
        elseif in_word
            # We've reached the end of a word
            in_word = false
            word = SubString(input, start_of_word:prev)
            d[word] = get(d, word, 0) + 1
        end
        if idx == lastindex(input)
            break
        end
        prev = idx
        idx = nextind(input, idx)
    end
    # Did input end on a word?
    if in_word
        word = SubString(input, start_of_word:idx)
        d[word] = get(d, word, 0) + 1
    end
    return d
end
```

Dunno if you prefer this. I don't mind.

cmcaine: I don't know if LICM will hoist `lastindex` out, btw. If it doesn't then this does a load of unnecessary work.

tracks/julia/exercises/word-count/mentoring.md
```diff
@@ -145,72 +145,51 @@ wordcount_withstatsbase(sentence) = countmap(filter!(!isempty, strip.(ispunct, s
 
 ** Totally manual ** 
 
-This method iterates through the string one character at a time and switches behavior at word boundaries (when `iswordchar` changes from true to false or false to true). The dictionary has `SubString` keys in this version, which are like the string-version of Array `views`.
+This method iterates through the string one character at a time and switches behavior at word boundaries (when `iswordchar` changes from true to false or false to true). The dictionary has `SubString` keys in this version, which are like the string equivalent of Array `views`. It is about 20% faster than the StatsBase version on small inputs like those in the tests (performance is almost identicaly for inputs without thousands of characters). It's about 1.5x-2x faster than the next-fastest regex version, likely owing to the smaller number of allocations.
```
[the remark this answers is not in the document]
tomerarnon: Ha! Obviously typed very quickly. It was basically meant to say the performance is the same for long inputs. I'll remove it for being superfluous though.

tracks/julia/exercises/word-count/mentoring.md
tomerarnon: Wow, I just assumed the lastindex check would be hoisted and didn't even bother to check! Precomputing it is a full 10% faster. 
The above version though didn't work for me on contractions. I thought only the lookahead was missing, but it turned out more involved since adding it naively introduced a boundserror. The current working version, uncommented looks like this (it's a shame syntax highlighting is better in comments than in the markdown preview...)

```julia
function wordcount(input)
    input = lowercase(input)
    d = Dict{SubString{typeof(input)}, Int}()
    iswordchar(c) = !ispunct(c) && !isspace(c)

    in_word = false
    prev = 0
    lastind = lastindex(input)
    for i in eachindex(input)
        c = input[i]
        if i == lastind
            prev = in_word ? prev : i
            i = iswordchar(c) ? i : i-1
            if prev <= i
                k = SubString(input, prev:i)
                d[k] = get(d, k, 0) + 1
            end
        elseif iswordchar(c)
            if !in_word
                in_word = true
                prev = i
            end
        elseif in_word
            c == ''' && iswordchar(input[nextind(input, i)]) && continue

            in_word = false
            k = SubString(input, prev:i-1)
            d[k] = get(d, k, 0) + 1
        end
    end
    d
end
```

cmcaine: Lol, I should stop posting up quick, untested code.

Here's a version that works, but I think your one is better.

```jl
function wordcount(input)
    input = lowercase(input)
    d = Dict{SubString{typeof(input)}, Int}()
    iswordchar(c) = !ispunct(c) && !isspace(c)

    isempty(input) && return d

    in_word = false
    start_of_word = 0
    prev = 1
    idx = 1
    while true
        chr = input[idx]
        if iswordchar(chr)
            if !in_word
                in_word = true
                start_of_word = idx
            end
        elseif in_word
            # We've reached the end of a word
            in_word = false
            word = SubString(input, start_of_word:prev)
            d[word] = get(d, word, 0) + 1
        end
        if idx == lastindex(input)
            break
        end
        prev = idx
        idx = nextind(input, idx)
    end
    # Did input end on a word?
    if in_word
        word = SubString(input, start_of_word:idx)
        d[word] = get(d, word, 0) + 1
    end
    return d
end
```

cmcaine: And yeah, I think Julia is currently not very compatible with LLVM's LICM pass, or possibly with some other LLVM pass, and lots of stuff that one would like hoisted isn't. I think the Julia compiler people are aware and there's some work that might fix it in the future.

e.g. anything that can error usually can't be hoisted, and while lastindex won't error on valid Strings, it might error on strings that contain bytes that are illegal in UTF-8, for example.

cmcaine (approved):

PR #1821, opened 2020-11-08

cmcaine:
Fix regex in examples 2 and 3 and use a consistent case for "regex".

### DIFF
--- a/tracks/julia/exercises/word-count/mentoring.md
+++ b/tracks/julia/exercises/word-count/mentoring.md
@@ -27,10 +27,10 @@ Dict{String,Int64} with 6 entries:
 
 **Split, then strip quotation**
 
-This solution, like all the others below uses "Regex".
+This solution, like all the others below uses "regex".
 Regex is a concise language for describing patterns in strings.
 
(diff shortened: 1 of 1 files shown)


tracks/julia/exercises/word-count/mentoring.md
```diff
@@ -27,10 +27,10 @@ Dict{String,Int64} with 6 entries:
 
 **Split, then strip quotation**
 
-This solution, like all the others below uses "Regex".
+This solution, like all the others below uses "regex".
 Regex is a concise language for describing patterns in strings.
```
SaschaMann, line 31:
```suggestion
Regex (short for regular expression) is a concise language for describing patterns in strings.
```

tomerarnon (approved):

tracks/julia/exercises/word-count/mentoring.md
cmcaine: I think it is best not to expand this because:

1. "Regular expression" is a name, but doesn't help you understand the concept at all
2. PCRE regexes aren't regular expressions

PR #1829, opened 2020-11-13
Fix typo

OMEGA-Y:


rpottsoh:
I am inclined not to approve these changes as the perceived misspellings are not misspellings according to the Queen's English.  I am not aware of a mandate that only American English is to be used here.

OMEGA-Y:
Oh, I didn't know that fact.
I'll close this.

PR #1830, opened 2020-11-14
rust/sublist refactor and pattern matching

paolosimone:
### Refactor the original solution

- move sublist-check subroutine to separate function for better readability
- remove length checks as they are implied in subroutine 

### Add pattern matching alternative solution

If a list is both a sublist and a superlist of another one then they are equals.
Using this notion we can convert the `if/else` cascade to an (arguably) more idiomatic `match` statement.


SleeplessByte (commented):
Looks fine with me. I hope @exercism/website-copy-rust can pick this up. If this isn't picked up within ~2 weeks, feel free to tag `[at]exercism/reviewers`.

angelikatyborska:
Excuse me, I clicked the "close" button accidentally 😭

coriolinus (approved):

SleeplessByte (approved):

PR #1833, opened 2020-11-22
Add Ron Au to list of Rust Mentors

ronCYA:


mentors/rust/r.json
```diff
@@ -6,5 +6,13 @@
     "link_url": null,
     "avatar_url": null,
     "bio": "An avid believer of the Feynman technique. Currently a professional trying to learn Rust and also teach it at the same time."
+  },
+  {
+    "github_username": "ronCYA",
+    "name": "Ron Au",
+    "link_text": "ronvoluted.com",
```
rpottsoh, line 13: link_text should be something other than the link_url.  If the link_text wants to be the same as link_url then
```suggestion
    "link_text": null,
```

rpottsoh (changes requested):

mentors/rust/r.json
ronCYA: That makes sense, apologies for that. This has been ammended.

rpottsoh (approved):

PR #1842, opened 2020-11-29
[Elixir] Add automated comment for pacman-rules

angelikatyborska:
Related analyzer PR: https://github.com/exercism/elixir-analyzer/pull/30

I feel like I should justify why the student should use `and` over `&&`, but the only reason I can think of is that they will need it later in guards, not necessarily in this exercise. I think @neenjaw is the author of the advice to write an analyzer with the rule, so maybe he has a better idea.


neenjaw:
The reason to use `and` over `&&` is because `and` is for when you expect boolean values, and will raise an error when a non-boolean is used. Since we work with boolean values in this exercise, `and` should be used. (Source: https://elixir-lang.org/getting-started/basic-operators.html)

Furthermore `and` is guard-friendly.

angelikatyborska:
> The reason to use and over && is because and is for when you expect boolean values, and will raise an error when a non-boolean is used.

But that is only true about the first argument, which makes it even worse than `&&` IMHO, because it gives you a false sense of a safety.

```
iex()> true and "foo"
"foo"
iex()> "foo" and true
** (BadBooleanError) expected a boolean on left-side of "and", got: "foo"
```

I don't understand why anyone would design an operator that is supposed to be symmetric like this.

neenjaw:
:shrug:

angelikatyborska:
Well, anyway, I think this can be merged without any changes

SleeplessByte (approved):

SleeplessByte:
And so it was merged without any changes.

PR #1845, opened 2020-12-05
rust/reverse-string: Type on the left preferable to Explicit turbofish

gilescope:
In this example it's extreamly clear what the return type is. I don't think it's a great example of where to use an explicit turbofish.

It's pretty controversial whether turbofish is the 'better' way:
https://github.com/rust-analyzer/rust-analyzer/issues/3737

In general, I would argue that option 1 without the turbofish is the more idiomatic way to specify the type. There's less angled brackets than option 2 so it's easier for the eye to read and its slightly shorter:

Option 1:
```rust
let x: Vec<char> = x.iter().collect();
```

Option 2:
```rust
let x = x.iter().collect::<Vec<char>>();
```

(I do agree that for larger functions having a type on each line is very helpful, but for this small function I could not see that a rust developer would really use collect with turbofish for a one line function. I do agree there's a teaching point about having an explicit type per line, but I don't think this exercise is the right place to make that point.)

### DIFF
--- a/tracks/rust/exercises/reverse-string/mentoring.md
+++ b/tracks/rust/exercises/reverse-string/mentoring.md
@@ -12,7 +12,7 @@ A reasonable solution should:
 - Use [.chars()](https://doc.rust-lang.org/std/primitive.str.html#method.chars),
 [.rev()](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.rev), and
 [.collect()](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect)
-- Be explicit with `.collect()` by using the turbofish operator `::<>`
+- No need to be explicit with `.collect()` by using the turbofish operator `::<>` because the collected type is in the type signature.
 - Not use other iterators or iterator methods like `.into_iter()` or
 `.flat_map()`
 - Not use variables or loops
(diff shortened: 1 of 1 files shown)


SleeplessByte:
Feels like a style choice, not necessarily idiomatic usage. 🤷🏽 

If you can get a @exercism/rust  -acean to sign of with this, all good with me.

petertseng (commented):
I confirm that the Rust *track maintainers* prefer **not** to specify the type of `collect`, as evidenced in https://github.com/exercism/rust/pull/807#discussion_r268400494.

Since this document is for mentors and I am not a mentor, I am not authorised to express an opinion on this pull request; I merely provide this information in hopes that it can help.

petertseng:
EBWODP: The link should point to https://github.com/exercism/rust/pull/807#discussion_r268400790 instead. My apologies.

petertseng:
EBWODP: I also confirm in my personal capacity that if I were asked to explicitly specify the type of `collect` in the example solutions listed, I would question the need to do so.

Therefore, even though I am still not authorised to express an opinion on this pull request, I want to tell you that if I were authorised, I would argue in favour of merging it.

workingjubilee:
In favor of merging. There's no reason to explicate the type here, and I don't remember ever dinging someone for it when I mentored this problem.

gilescope:
@SleeplessByte I think there's concensus here.

SleeplessByte (approved):

SleeplessByte:
Thank you all!

PR #1861, opened 2020-12-17
[Julia] Update mentoring notes for raindrops

Luapulu:
Updated notes to reflect insights from better benchmarks.

See also #1857

### DIFF
--- a/tracks/julia/exercises/raindrops/mentoring.md
+++ b/tracks/julia/exercises/raindrops/mentoring.md
@@ -111,15 +111,30 @@ The following three solutions are slower but make use of concatentaion so as to 
 
 **Compact**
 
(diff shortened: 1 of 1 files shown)


Luapulu:
I have a faster version of the IOBuffer solution.

```
function raindrops2(number)
    buf = IOBuffer(sizehint=15)

    (f3 = number % 3 == 0) && write(buf, "Pling")
    (f5 = number % 5 == 0) && write(buf, "Plang")
    (f7 = number % 7 == 0) && write(buf, "Plong")

    if f3 | f5 | f7
        return String(take!(buf))
    else
        return string(number)
    end
end
```

cmcaine:
You can also do `position(buf) != 0` or (cheekily) `buf.size != 0`.

cmcaine:
I think this is obsoleted? Feel free to reopen.

PR #1862, opened 2020-12-17
julia: revise raindrops mentoring notes

cmcaine:
See also #1861, #1857

@Luapulu, what do you think of this?


SaschaMann (approved):
Looks good to me!

Would be nice to have some benchmarking guide as part of the track docs at some point

cmcaine:
Great :)

Yeah, I was thinking that a PR might be in order to BenchmarkTools to make it easier to benchmark variants of a function, then we could talk about that. I don't really like that the standard is just to eyeball the summary statistics, we should be able to compare the distributions, and give a confidence that `f1` is faster than `f2` or something.

In the likely event that I don't get around to that anytime soon, a guide on how to do benchmarking with the existing tools is probably a good idea.

Luapulu:
Looks good. Thanks!

I guess I can delete my other PR where I was working on this.

PR #1864, opened 2020-12-18
[Julia] One more iteration to the raindrops notes

Luapulu:
I added a better compact solution, which is almost as fast as the tree of ifs and the lookup. 

```julia
function raindrops(n)
    (f3 = n % 3 == 0) && (s = "Pling")
    (f5 = n % 5 == 0) && (s = f3 ? "PlingPlang" : "Plang")
    (f7 = n % 7 == 0) && (s = f3 | f5 ? string(s, "Plong") : "Plong")
    return f3 | f5 | f7 ? s : string(n)
end
```

In fact, this solution is basically as fast as the switch, so I removed the switch. (I only added it to show a compact but fast solution. The new solution is just as fast *and* more compact)

I removed some of the details on benchmarking because with this solution you can no longer easily replace a `*=` with a `=` to see the difference between concatenating and using literal strings. I guess we could add @cmcaine's solution back in to show a basic version of the same format of solution and then also show this optimised version?

### DIFF
--- a/tracks/julia/exercises/raindrops/mentoring.md
+++ b/tracks/julia/exercises/raindrops/mentoring.md
@@ -138,81 +138,26 @@ function raindrops(number)
 end
 ```
 
-
-**Switch**
-
-A slightly slower version formatted as a single switch rather than a tree of ifs.
-
-```julia
-function raindrops(number)
-    f3 = number % 3 == 0
-    f5 = number % 5 == 0
-    f7 = number % 7 == 0
-
-    if f3 & f5 & f7
-        "PlingPlangPlong"
-    elseif f3 & f5 & !f7
-        "PlingPlang"
-    elseif f3 & !f5 & f7
-        "PlingPlong"
-    elseif f3 & !f5 & !f7
-        "Pling"
-    elseif !f3 & f5 & f7
-        "PlangPlong"
-    elseif !f3 & f5 & !f7
-        "Plang"
-    elseif !f3 & !f5 & f7
-        "Plong"
-    else
-        string(number)
-    end
-end
-```
-
 The following three solutions are slower but make use of concatentaion so as to be more concise, readable or extendable.
 
 **Compact**
 
-This solution is short, easy to read and almost as fast in the average case as the solutions above!
-It is fast because it avoids one string concatenation in the common case that `number` is divisible by 3
-and when `number` is divisible by 3, but not 5 and 7, the statically allocated string "Pling" is returned almost instantly.
+This solution is short, easy to read and almost fast as the solutions above in the average case!
+It is fast because it avoids string concatenation in most cases, except where the number is divisible by both 7 and 3 or 5.
 
 ```julia
 function raindrops(number)
-    s = ""
-    # The first equals here lets us avoid
-    # dynamically allocating a string in the
-    # case that number is only divisible by 3.
-    number % 3 == 0 && (s = "Pling")
-    number % 5 == 0 && (s *= "Plang")
-    number % 7 == 0 && (s *= "Plong")
-    isempty(s) && (s = string(number))
-    return s
+    (f3 = number % 3 == 0) && (s = "Pling")
+    (f5 = number % 5 == 0) && (s = f3 ? "PlingPlang" : "Plang")
+
+    # notice that here is the only place where we have to concatenate strings
+    (f7 = number % 7 == 0) && (s = f3 | f5 ? string(s, "Plong") : "Plong")
+
+    # We could also use the short circuiting || operator,
+    # but the bitwise | is simpler and has virtually no performance consequence here.
+    return f3 | f5 | f7 ? s : string(number)
 end
 ```
-
-Try changing `s = "Pling"` to `s *= "Pling"` and benchmarking with 3, 14, and 15 to see what's going on.
-
-<details>
-<summary>Expected benchmark results and explanation</summary>
-
-raindrops3a is the version with `s *= "Pling"`.
-
-```julia
-using BenchmarkTools
-
-@btime raindrops3($(3))         # 0 allocations
-@btime raindrops3a($(3))        # 1 allocation
-
-# 14 is not divisible by 3, so there's no difference here.
-@btime raindrops3($(14))        # 1 allocation
-@btime raindrops3a($(14))       # 1 allocation
-
-# 15 is divisible by 3 and 5 and we can still avoid the first allocation.
-@btime raindrops3($(15))        # 1 allocation
-@btime raindrops3a($(15))       # 2 allocations
-```
-</details>
 
 
 **Extendable**
--- a/tracks/julia/exercises/raindrops/mentoring.md
+++ b/tracks/julia/exercises/raindrops/mentoring.md
@@ -52,28 +52,25 @@ bar(x) = x < 10 ? 10 : log(x)
 If you want to benchmark lots of variants of the same function, you can define them as `raindrops1`, `raindrops2`, etc. and do some simple code generation like this:
 
 ```julia
-for i in 1:8
-   print("$i: ")
-   @btime foreach($(Symbol(:raindrops, i)), 1:1000)
-end
-```
-
-```
-1:   20.100 μs (914 allocations: 42.84 KiB)
-2:   85.000 μs (4457 allocations: 233.03 KiB)
-3:   27.800 μs (1256 allocations: 53.53 KiB)
-4:   24.300 μs (914 allocations: 42.84 KiB)
-5:   21.600 μs (914 allocations: 42.84 KiB)
-6:   35.700 μs (1457 allocations: 59.81 KiB)
-7:   47.600 μs (1589 allocations: 63.94 KiB)
-8:   86.500 μs (4457 allocations: 233.03 KiB)
+julia> for i in 1:6
+          print("$i: ")
+          @btime foreach($(Symbol(:raindrops, i)), 1:1000)
+       end
+1:   19.511 μs (914 allocations: 42.84 KiB)
+2:   21.166 μs (914 allocations: 42.84 KiB)
+3:   21.845 μs (914 allocations: 42.84 KiB)
+4:   37.608 μs (1589 allocations: 63.94 KiB)
+5:   31.062 μs (1256 allocations: 53.53 KiB)
+6:   21.833 μs (980 allocations: 44.91 KiB)
+7:   47.231 μs (1589 allocations: 63.94 KiB)
+8:   77.441 μs (4457 allocations: 233.03 KiB)
 ```
 </details>
 
 **if / else tree**
 
 ```julia
-function raindrops(number)
+function raindrops1(number)
     f3 = number % 3 == 0
     f5 = number % 5 == 0
     f7 = number % 7 == 0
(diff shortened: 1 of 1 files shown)


cmcaine:
I think this is more confusing and less general than the switch (once you understand the concept of the switch you can re-use it in other exercises, whereas this is essentially just an oddly formatted variant on the if/else tree).

This also removes content that might help a student understand the underlying principle better, so I think this should not be merged. If you're not convinced we can ask Sascha for a tie-breaking ruling :)

Luapulu:
I added explanation of the three functions now. I think there is some educational value in seeing the progression of solutions. Not only does it clarify the concept of reducing concatenations to improve performance, but it also illustrates the tradeoff between readability and performance.

tracks/julia/exercises/raindrops/mentoring.md
```diff
@@ -52,28 +52,25 @@ bar(x) = x < 10 ? 10 : log(x)
...
+          @btime foreach($(Symbol(:raindrops, i)), 1:1000)
+       end
+1:   19.511 μs (914 allocations: 42.84 KiB)
+2:   21.166 μs (914 allocations: 42.84 KiB)
+3:   21.845 μs (914 allocations: 42.84 KiB)
+4:   37.608 μs (1589 allocations: 63.94 KiB)
+5:   31.062 μs (1256 allocations: 53.53 KiB)
+6:   21.833 μs (980 allocations: 44.91 KiB)
+7:   47.231 μs (1589 allocations: 63.94 KiB)
+8:   77.441 μs (4457 allocations: 233.03 KiB)
```
SaschaMann, line 66: Should the times be normalised perhaps? The exact time is somewhat meaningless and having values relative to the fastest one would make it easier to read imo.

kotp: I agree, though it could be understood that these are example times.   Since they are in time order, they are already "relative to the fast one".

SaschaMann: >  Since they are in time order

They aren't.

kotp: Doh!

Luapulu: I've tried different options, like displaying trial judgements, but most ways I can think of add more confusion than they alleviate. Changing the order to be in perfect ascending order of times will also be more confusing I think.

kotp: If the code were changed and suggested to create different names, then the names could be used rather than a comment indicating the solution.  The results could be ordered, and I don't know why that would be confusing.  I would be confusing if you had to comment them manually, but if it can be ordered by time and the names are part of that data, then it should remain clear.

Luapulu: I mean, we could set this up real nice by using a BenchmarkGroup.

Luapulu: Something like:

```julia
suite = BenchmarkGroup()

suite["if-tree"]    = @benchmarkable foreach(raindrops1, 1:1000)
suite["lookup"]     = @benchmarkable foreach(raindrops2, 1:1000)
suite["switch"]     = @benchmarkable foreach(raindrops3, 1:1000)
suite["compact-v1"] = @benchmarkable foreach(raindrops4, 1:1000)
suite["compact-v2"] = @benchmarkable foreach(raindrops5, 1:1000)
suite["compact-v3"] = @benchmarkable foreach(raindrops6, 1:1000)
suite["extendable"] = @benchmarkable foreach(raindrops7, 1:1000)
suite["IOBuffer"]   = @benchmarkable foreach(raindrops8, 1:1000)

display(run(suite))
```

SaschaMann: I was thinking about something like this:

```
 1.000
 1.085
 1.120
 1.928
 1.592
 1.119
 2.421
 3.969
```

instead of the absolute values. But now that I actually calculated it, I'm not sure it adds any readability. Let's just leave it as it is :)

tracks/julia/exercises/raindrops/mentoring.md
```diff
@@ -56,15 +56,18 @@ julia> for i in 1:6
...
+3:   21.845 μs (914 allocations: 42.84 KiB)    # switch
+4:   37.608 μs (1589 allocations: 63.94 KiB)   # compact v1
+5:   31.062 μs (1256 allocations: 53.53 KiB)   # compact v2
+6:   21.833 μs (980 allocations: 44.91 KiB)    # compact v3 
+7:   47.231 μs (1589 allocations: 63.94 KiB)   # extendable
+8:   77.441 μs (4457 allocations: 233.03 KiB)  # IOBuffer
 ```
+
+Note, that your times may vary from these.
+
```
SaschaMann, line 70: These comments help a lot :)

Luapulu:
So, is everyone ok with how it is now?

For those that want to somehow sort the benchmarks or display them differently, can you post an example of the code to display the benchmarks? Most things I've tried are pretty complex and really don't represent what most people would ever use to compare a few different versions of a function.

tracks/julia/exercises/raindrops/mentoring.md
```diff
@@ -186,26 +206,56 @@ function raindrops4(number)
...
+Let us compare raindrops4, raindrops5 and raindrops6, the three versions of our compact solution.
+For overall benchmarks, see the aside on benchmarking raindrops above.
 
 ```julia
 using BenchmarkTools
 
-@btime raindrops4($(3))         # 0 allocations
-@btime raindrops4a($(3))        # 1 allocation
+# 4 hast to concatenate "" and "Pling", 5 and 6 don't.
+# 6 is slightly faster because it doesn't use the isempty function.
```
cmcaine, line 237:
```suggestion
# raindrops6 is slightly faster because it doesn't use the isempty function.
```

tracks/julia/exercises/raindrops/mentoring.md
```diff
@@ -186,26 +206,56 @@ function raindrops4(number)
...
+@btime raindrops6($(14))        # 0 allocations, much faster
+
+# 15 is divisible by 3 and 5, so solution 5 can still avoid the first concatenation of "" and "Pling"
+# It can't avoid the concatenation to "PlingPlang", which solution 6 can
+@btime raindrops4($(15))        # 2 allocations
+@btime raindrops5($(15))        # 1 allocations, faster
+@btime raindrops6($(15))        # 0 allocations, even faster
+
+# Since 35 has factors 5 and 7, solutions 4 and 5 both require two concatenations.
+# Solution 6 hass to concatenate only "Plang" and "Plong"
```
cmcaine, line 255:
```suggestion
# Solution 6 has to concatenate only "Plang" and "Plong"
```

tracks/julia/exercises/raindrops/mentoring.md
```diff
@@ -186,26 +206,56 @@ function raindrops4(number)
...
+    # the next line would have to contain another layer of branching
+    # At that point you might as well use the full if tree solution
+    (f7 = n % 7 == 0) && (s = f3 | f5 ? string(s, "Plong") : "Plong")
+    return f3 | f5 | f7 ? s : string(n)
+end
+```
+
+This solution retains the structure and compactness of the previous solutions, but achieves very
+good performance (similar to switch) because string concatenation is only required for numbers divisble by 3 and/or 5
+and 7.
```
cmcaine, line 225: The end of this sentence could be clearer. I'd also say that this solution is quite a lot harder to understand than the other compact one. Changing the tests to test and assign and using lots of ternary expressions means there's a lot going on.

cmcaine (commented):
I'm still opposed to this for the reasons above and below, but I'm happy for this to be merged over my objections if enough others prefer it.

tracks/julia/exercises/raindrops/mentoring.md
Luapulu: What about saying that it retains the structure, sacrifices some readability because of the added complexity but also improves performance to where this solution is as fast as the switch? We can also mention that it will depend on the situation if improving performance at the cost of readability is actually advantageous. Depending on the need for performance you may choose different variants.

cmcaine: I don't think it really does retain much of the structure. It has the same number of lines, but I don't think we should be encouraging assignments on the left hand side of an expression, it's already a little frowned on to be doing it on the right hand side of a short-circuiting op.

Here's the solution re-written to be a bit easier to read. It also makes clear that this is basically just the if/else tree solution:

```julia
function raindrops6(n)
    f3 = n % 3 == 0
    f5 = n % 5 == 0
    f7 = n % 7 == 0

    f3 && (s = "Pling")
    f5 && (s = f3 ? "PlingPlang" : "Plang")
    f7 && (s = f3 || f5 ? s * "Plong" : "Plong")

    # If we wanted to avoid all concatenations, we could use this instead, but
    # at that point, this solution becomes equivalent to the tree of if/elses
    # f7 && (s = f3 ? (f5 ? "PlingPlangPlong" : "PlingPlong")
    #               : (f5 ? "PlangPlong" : "Plong"))

    return f3 || f5 || f7 ? s : string(n)
end
```

I'm just not convinced that this solution offers much extra learning over the compact solution, and, if we do include it, I think it makes more sense to make it equivalent to the if/else tree to show that that solution can be written concisely, maybe replacing the "switch" solution (which is just a worse version of the lookup solution, IMO)

Again, if others prefer how this PR already is, then that's fine. This is not a veto :)

Luapulu: I like the code you posted there. Spacing it out and showing how you can make it more and more like the if tree (and thereby also improve performance) by adding comments is a more concise and clear way to say the same thing.

cmcaine: I guess I shouldn't say "equivalent", because the conditions are in a different order to the tree. It's just similar.

cmcaine: OTOH, if you replace just the bottom level of the if tree with ternary expressions you get this fairly short thing:

```jl
function raindrops(n)
    f3 = n % 3 == 0
    f5 = n % 5 == 0
    f7 = n % 7 == 0

    if f3
        if f5
            f7 ? "PlingPlangPlong" : "PlingPlang"
        else
            f7 ? "PlingPlong" : "Pling"
        end
    else
        if f5
            f7 ? "PlangPlong" : "Plong"
        else
            f7 ? "Plong" : string(n)
        end
    end
end
```

Or, golfed:

```jl
function raindrops(n)
    f3 = n % 3 == 0
    f5 = n % 5 == 0
    f7 = n % 7 == 0
    f3 ? f5 ? f7 ? "PlingPlangPlong" : "PlingPlang"
            : f7 ? "PlingPlong"      : "Pling"
       : f5 ? f7 ? "PlangPlong"      : "Plang"
            : f7 ? "Plong"           : string(n)
end
```

And that's just better in the number of comparisons done than raindrops6 (this one is always three comparisons, raindrops6 is 4 in the best case and 6 in the worst).

cmcaine:
Do you still want this merged? If you do, ping Sascha for a review and if they're happy it can go in :)

SaschaMann:
I agree with @cmcaine's concerns. This also feels almost overwhelmingly long. Perhaps there could be two sections? One for general examples, and one for the in-depth performance discussion for students who have indicated that they're interested in those?

Luapulu:
I like the idea of splitting things into separate sections.

In any case, I’d want to change quite a few things if this is supposed to be merged, so we can close this and if I get to it I’ll open a new PR in future.

PR #1866, opened 2020-12-19
Raindrops notes round 3

tomerarnon:
The names that made it into the notes are all `raindrops`, which has the unfortunate consequence of making the benchmark not runnable out of the box, and not easily associated with its version, which (I think) is a shame. This adds the numbers back in.


cmcaine (approved):
Thanks!

PR #1870, opened 2020-12-23
cpp/leap

zq-chen:
* Changing the argument type of the `is_leap_year ` function from `const int&` to `const int`, since passing the primitive integer type by value is preferred over passing-by-reference in C++.
* Adding a possible talking point of making the function `constexpr`.


SleeplessByte:
@exercism/cpp can someone give this a quick look?

siebenschlaefer (approved):
It's definitely an improvement.

vaeng (approved):

SleeplessByte (approved):
Thank you both <3

PR #1880, opened 2021-01-06
[CoC] Add missionising to language that is not welcome

SaschaMann:
The current CoC does not cover missionising or hateful religious speech such as wishing mean things upon non-believers.

Many people live in places where religion and religious speech is used as a tool of oppression. Exercism should be a safe place for those, just like it should be a safe place for religious people. That, imo, requires making it a largely religion-free place. Symbols such as wearing religious symbols in one's picture should obviously not be affected by this, this is merely about actively forcing religion upon others.

### DIFF
--- a/pages/code_of_conduct.md
+++ b/pages/code_of_conduct.md
@@ -32,7 +32,7 @@ Exercism should be a safe place for everybody regardless of
 - Physical appearance (including but not limited to body size)
 - Race
 - Age
-- Religion
+- Religion (or lack of religion)
 - Anything else you can think of.
 
 As someone who is part of this community, you agree that:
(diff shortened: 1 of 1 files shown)


pages/code_of_conduct.md
```diff
@@ -42,6 +42,7 @@ As someone who is part of this community, you agree that:
 - We respect people’s boundaries and identities.
 - We refrain from using language that can be considered offensive or oppressive (systemically or otherwise), eg. sexist, racist, homophobic, transphobic, ableist, classist, etc. - this includes (but is not limited to) various slurs.
```
SaschaMann, line 43:
```suggestion
- We refrain from using language that can be considered offensive or oppressive (systemically or otherwise), eg. sexist, racist, homophobic, transphobic, ableist, classist, missionizing, etc. - this includes (but is not limited to) various slurs.
```

PR #1887, opened 2021-01-15
Add note about `brew update`

ailyntang:
When typing in `brew update`, it looks like nothing is happening in terminal for awhile. 

People may think this means they have done something wrong.

My hope is this additional comment will help new coders with the installation.

### DIFF
--- a/walkthrough/index.html
+++ b/walkthrough/index.html
@@ -869,6 +869,8 @@ brew update
 brew install exercism
 ```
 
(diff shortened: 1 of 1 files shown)


kotp (approved):

ailyntang:
Thanks for approving @kotp . 

Do I need to do another step before someone is allowed to merge?

kotp:
Thank you for bumping it, it slipped through the cracks.

PR #1889, opened 2021-01-16
update a.json

AzT3Risk:
Adding myself as a mentor in R track


rpottsoh (changes requested):
Please adjust your indentations so that your bio content is aligned with the bio before yours.

rpottsoh (approved):

PR #1891, opened 2021-01-21
Fix link to documentation in C# mentoring notes for space-age

sanderploegsma:
Due to an extra set of parentheses, the link to expression-bodied methods was actually pointing to https://exercism.io/mentor/solutions/(https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/lambda-operator#expression-body-definition), which is of course not right.


ErikSchierboom (approved):

ErikSchierboom:
Thanks @sanderploegsma!

PR #1902, opened 2021-02-03
Format markdown according to new spec

iHiD:
This attempts to bring website-copy's markdown files in line with [this spec change](https://github.com/exercism/docs/pull/38).

PR #1907, opened 2021-02-17
julia: add phone-number mentoring notes

cmcaine:


tomerarnon (approved):
Nice, it would be good to have some notes for this one as well!

SaschaMann (commented):
@iHiD does the new spec (#1902) also apply to these notes?

iHiD:
Yes. But I'm happy to get this merged then fix them all seperately.

PR #1908, opened 2021-02-18
julia: update diff-of-sq mentoring notes

cmcaine:


tracks/julia/exercises/difference-of-squares/mentoring.md
```diff
@@ -52,6 +52,8 @@ You may see some solutions where the `difference` has been partially expanded an
 n111b111's solution:
 
 ````
+Using `evalpoly` makes this more efficient than most hand-written analytic solutions because `evalpoly` generates efficient code using Horner's method.
```
SaschaMann, line 55:
```suggestion
Using `evalpoly` makes this more efficient than most hand-written analytic solutions because `evalpoly` generates efficient code using [Horner's method](https://www.math10.com/en/algebra/horner.html).
```

I couldn't find a good explanation for it, but having one at all may be useful to people who aren't familiar with the method.

SaschaMann (dismissed):
Nice!

SaschaMann (approved):

PR #1916, opened 2021-02-28
Change comment format

neenjaw:
Change to `[comment]: # (this is a comment)`

### DIFF
--- a/analyzer-comments/elixir/two-fer/use_string_interpolation.md
+++ b/analyzer-comments/elixir/two-fer/use_string_interpolation.md
@@ -1,6 +1,4 @@
-<!--
-This warning is raised when the solution doesn't use string interpolation
--->
+[comment]: # (This warning is raised when the solution doesn't use string interpolation)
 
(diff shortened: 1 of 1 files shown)


neenjaw:
@iHiD

angelikatyborska:
https://stackoverflow.com/a/32190021/5726486 this says that there needs to be an empty line before that fake link. But I don't really understand why.

neenjaw:
I'm just going with what @iHiD had suggested we try.  He had an example here: https://johnmacfarlane.net/babelmark2/?text=above%0A%0A%5Bcomment%5D%3A+%23+(THIS+SHOULD+BE+REMOVED)%0Abelow

neenjaw:
And if you view the file with rich display, it is hidden from view in GitHub. So I'm hopeful it will work on the website

iHiD:
> says that there needs to be an empty line before that fake link.

If I understand correctly (which is far from certain) it's because if there's not it'll get treated as part of the same paragraph as the line above, rather than it's own thing. But this doesn't matter if it's the first line.

iHiD (approved):

neenjaw:
I don't have write access to merge this, thanks for approving though!

iHiD:
@neenjaw Weeeeeird!. You do now :)

PR #1919, opened 2021-03-03
Update Elixir word-count mentoring notes

bowmanmike:
I've seen some students use `Enum.frequencies_by` after splitting the string, which results in a really clean solution. Think we should add it to the guide!

### DIFF
--- a/tracks/elixir/exercises/word-count/mentoring.md
+++ b/tracks/elixir/exercises/word-count/mentoring.md
@@ -68,6 +68,17 @@ defmodule WordCount do
 end
 ```
 
+```elixir
+# Enum.frequencies_by/2 can do a lot of the legwork for you here
+defmodule WordCount do
+  def count(sentence) do
+    sentence
+    |> String.split(~r/[^[:alnum:]-]/u, trim: true)
+    |> Enum.frequencies_by(&String.downcase/1)
+  end
+end
+```
+
 ### Common suggestions
 
 #### German `öüä`


neenjaw:
There are some changes coming to this exercise with v3's release (cc@angelikatyborska). Can you make sure that the solution using frequencies_by still passes the exercise?

I think the example solution in the repo now uses frequencies by, so you could use that.

angelikatyborska:
None of the solutions listed in the mentoring notes pass the new tests. I don't think we need to worry about this now. Nobody has access to the updated exercises yet. Once everybody starts using them, mentors will notice that the mentoring notes are out of date and update them (I hope...). I'm sure it's not a new problem of what to do with mentoring notes when exercises change, I wonder if @iHiD has some insights 🤔. Should the people updating the exercise also always be responsible for checking mentoring notes, or is it fine to say YOLO and let mentors deal with it once they notice mentoring notes are out of date?

> I think the example solution in the repo now uses frequencies by, so you could use that.

Nope, it doesn't. Example solutions must pass tests in Elixir 1.6 and up so most of them don't use new features.

As for the new proposed solution, I think conceptually it's very similar to the previous one that uses `Enum.frequencies/1`, so I don't see a lot value in adding it, but I also don't see any harm in adding it because it's still a good solution.

angelikatyborska (approved):

neenjaw:
This repo isn't syncing to v2, so mentors will only see this in the context of v3.

iHiD:
> Should the people updating the exercise also always be responsible for checking mentoring notes, or is it fine to say YOLO and let mentors deal with it once they notice mentoring notes are out of date?

Both? 🙂 

There's to questions in there. The process and what to do now.
- Re the process, in normal times, I think if a maintainer changes an exercise then it's good practice to change the mentoring notes that go alongside it. In many ways, those notes belong to the exercise and are part of it. The only reason they're in website-copy rather than the exercise folder is to allow for mentors to be more autonomous in their changes. We're awful at having these processes documented (largely because of the adhoc YOLO nature of OSS) but maybe @ErikSchierboom, we could discuss where we could put info like this?
- In this situation, with so many exercises that will be tweaked, and maintainers burden is already high, I suspect the best thing would be to have a little "admonition" banner thing that in the nodes that says "Warning: This exercise has changed in the migration to v3. Please don't blindly rely on these notes. Please remove this banner once they are checked" or something. Maintainers can give us a list of the exercises that have changed and we can script-add that to them.

It's also worth noting that there's *loads* of commentary on mentor notes about how the exercise is "core" and fits into this bit of the track, and don't overload a student with XYZ. I'm not sure those things remain relevant at all in v3, so there's lots of content that probably needs pruning.

neenjaw:
@ihid, that makes sense. @angelikatyborska sorry, I forgot about the 1.6 compatibility requirement in the example.

So given that this repo is not syncing, and this solution (and the others) won't pass v3, what's the solve here for @bowmanmike 
 to get this merged? Just merge as is? Add the banner @ihid suggested? Fix all of the solutions for v3?

angelikatyborska:
I thing to be fair to @bowmanmike, we should judge the PR in the context of v2 (which probably means just merge!), and resolve the changed exercises issue outside of this PR.

neenjaw (approved):

bowmanmike:
Thanks everyone!

PR #1940, opened 2021-03-18
Update Haskell track Pangram exercise mentoring notes.

Javran:
Add few suggestions regarding `Set`.

Also remove a snippet that doesn't have the laziness property
described.

In particular: `Set.fromList` cannot work with list that contains `_|_` as
it's strict on keys.


angelikatyborska:
tagging @exercism/haskell because @exercism/website-copy-haskell has no members :)

SleeplessByte (approved):
Checked common solutions and this note seems to match

PR #1945, opened 2021-03-23
Mentor bio tradfursten

tradfursten:

PR #1946, opened 2021-03-23
Quote strings in labeler.yml

angelikatyborska:
`*` starts an alias in YAML, so we need to use quotes to turn them intro string literals. Hopefully fixes the failure seen in the CI run for https://github.com/exercism/website-copy/pull/1945


SaschaMann (approved):

SaschaMann:
Not sure if it works, let's try.

PR #1973, opened 2021-04-24
Julia: add mentoring notes for bob

cmcaine:


tomerarnon:
Glad you're adding notes to this problem! The solution I've been using is this one:
```julia
function bob(stimulus)
    all(isspace, stimulus) && return "Fine. Be that way!"
    
    upper = any(isuppercase,  stimulus) && !any(islowercase, stimulus)
    question = let
        indq = findlast(!isspace, stimulus)
        !isnothing(indq) && stimulus[indq] == '?'
    end

    if question && upper
        return "Calm down, I know what I'm doing!"
    elseif question
        return "Sure."
    elseif upper
        return "Whoa, chill out!"
    else
        return "Whatever."
    end
end
```
It's nearly identical to yours, the differences being the empty case being handled at the top, and the strange use of a let block in a function (which I point out with a note like "Notice that I assigned `question` as the returned value of a let-block which I admit is quite unusual, but I thought it would be nice to demonstrate the option!")

I actually didn't realize that `strip` didn't allocate, so I thought I was being quite clever using `findlast` to avoid the `strip`!

I like the two you have here and will approve though.

tomerarnon (dismissed):

cmcaine:
To be fair, views and substrings did used to allocate :) I think that might have changed in 1.3?

I think your solution is good, too. I think my use of inner functions is a bit silly when they will almost always all be executed, but I thought it was cute at the time and it's not awful :)

I prefer how the `if` comes out if you avoid nesting, too, but I thought I'd do something a bit different.

I think I'll add a short section about using a named tuple to collect all the strings (because some students like to collect the strings into one data structure then refer to them as constants). It's not very useful if you define all the logic in advance, but in more complicated functions that return strings it could be a useful pattern.

cmcaine:
I think this is ready to go :)

tracks/julia/exercises/bob/mentoring.md
```diff
@@ -62,7 +62,32 @@ end
...
-Some work can be repeated (`isspace` will be called by both `is_silence` and `rstrip`, and `is_question()` can be called twice), but I like how the work is clearly split into three simple predicates and that there are no intermediate variables.
+> Some work can be repeated (`isspace` will be called by both `is_silence` and `rstrip`, and `is_question()` can be called twice), but I like how the work is clearly split into three simple predicates and that there are no intermediate variables.
+>
+> This function avoids copying the string at any point, too, which is nice for performance.
 
-This function avoids copying the string at any point, too, which is nice for performance.
+** Collecting the strings in one place **
+
+Some students prefer to collect all of the possible return strings in one place and then refer to them by an ID.
+`Bob` is quite a simple exercise, so it doesn't necessarily help much, but in similar functions with more complicated control flow or where you want to support translations or similar, you might want to do this.
```
SaschaMann, line 72: > `Bob` is quite a simple exercise

I think Katrina would disagree: https://www.youtube.com/watch?v=GWEEPt8VvmU :D

SaschaMann (approved):

PR #1979, opened 2021-05-06
Alternate Reasonable Solution to python/robot-names

satorusxm:
Proposing an alternate Reasonable Solution to the robot-name exercise under the track/python. Please let me know if anything is wrong/inefficient about my proposed solution.

### DIFF
--- a/tracks/python/exercises/robot-name/mentoring.md
+++ b/tracks/python/exercises/robot-name/mentoring.md
@@ -20,12 +20,6 @@ class Robot(object):
     names = set()
 
     def __init__(self):
-        self.set_name()
-
-    def reset(self):
-        self.set_name()
-
-    def set_name(self):
         while True:
             self.name = ''.join(
                 random.choices(string.ascii_uppercase, 2) + random.choices(string.digits, 3)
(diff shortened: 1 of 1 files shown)


tracks/python/exercises/robot-name/mentoring.md
```diff
@@ -20,19 +20,16 @@ class Robot(object):
...
         while True:
             self.name = ''.join(
                 random.choices(string.ascii_uppercase, 2) + random.choices(string.digits, 3)
             )
             if self.name not in Robot.names:
                 Robot.names.add(self.name)
                 break
+
+    def reset(self):
+        self.__init__()
```
IsaacG, line 32: To quote from my Robot notes,

> In general, dunder-methods (double-underscore) should not be directly called when it can be helped. They are for internal use. Prefer instead to have `__init__` call -reset()`.

DanielNoord (changes requested):
I agree with @IsaacG here, I don't think this change should be made.

BethanyG:
Closing this.   
1.  It's an unnecessary/unwanted change.
2.  It's stale.

PR #1988, opened 2021-05-21
java/gigasecond: Update mentoring.md

richarda23:
A student had added try/catch blocks to their solution https://exercism.io/mentor/solutions/8987494a21964e9f93f86d68dac5ec8d#discussion-post-929451,  presumably to deal with these exceptions, but handled them incorrectly. Exception handling is a tricky concept for beginners so feel some advice here is useful.

### DIFF
--- a/tracks/java/exercises/gigasecond/mentoring.md
+++ b/tracks/java/exercises/gigasecond/mentoring.md
@@ -42,7 +42,10 @@ A student can choose to compute the result for `getDate()` in the constructor. I
 
 ### Exception handling
 
-`plusSeconds` can throw a `java.time.DateTimeException`. A student may choose to  catch this exception in the constructor, but this results in  `birthDateTime` being null, so the class won't have been instantiated properly. There's no sensible default value that be stored in  `birthDateTime`. It's better to let the exception be thrown, failing fast, and let calling code deal with it. 
+`plusSeconds` can throw a `java.time.DateTimeException`.
+A student may choose to catch this exception in the constructor, but this results in `birthDateTime` being null, so the class won't have been instantiated properly.
+There's no sensible default value that be stored in `birthDateTime`.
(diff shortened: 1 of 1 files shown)


ericbalawejder (approved):

tracks/java/exercises/gigasecond/mentoring.md
```diff
@@ -34,13 +34,16 @@ class Gigasecond {
...
-
 ## Talking points
 
 ### Eager computation
 
 A student can choose to compute the result for `getDate()` in the constructor. It is a really good optimization. In that case, it should be noted that it is possible because the `birthDate` is only provided in the constructor and is immutable.
 
+### Exception handling
+
+`plusSeconds` can throw a `java.time.DateTimeException`. A student may choose to  catch this exception in the constructor, but this results in  `birthDateTime` being null, so the class won't have been instantiated properly. There's no sensible default value that be stored in  `birthDateTime`. It's better to let the exception be thrown, failing fast, and let calling code deal with it.
```
kotp, line 45:
```suggestion
`plusSeconds` can throw a `java.time.DateTimeException`.
A student may choose to catch this exception in the constructor, but this results in `birthDateTime` being null, so the class won't have been instantiated properly.
There's no sensible default value that be stored in `birthDateTime`.
It's better to let the exception be thrown, failing fast, and let calling code deal with it. 
```

Single sentence per line, and removed some duplicate spaces. (I did this in the text area, so I might have missed some.)

kotp (changes requested):

tracks/java/exercises/gigasecond/mentoring.md
kotp: The last sentence could be written better, I think.

"It's better to let the exception be thrown, failing fast, letting the calling code deal with it."

SleeplessByte (approved):

PR #1989, opened 2021-05-22
[Ruby] Fix error in two-fer analyzer comment

angelikatyborska:
When I was testing the iterations page I noticed that I was given analyzer feedback with slightly weird  syntax coloring. I'm not a Ruby expert (anymore...), but I'm pretty sure the trailing `"` isn't supposed to be there.
Screen Shot 2021-05-22 at 10 41 50


angelikatyborska:
Found one more problem :) using the hint literally makes the tests fail because it should be "one for me", not "one for you".
Screen Shot 2021-05-22 at 10 50 18

SleeplessByte (approved):

SleeplessByte:
Thanks! ☺️

PR #2007, opened 2021-06-25
Escape % in elixir/pacman-rules

angelikatyborska:
I was told by Ruby wizards that this is the correct way to escape this magical character so that the website doesn't blow up :) 

Fixes https://github.com/exercism/website/issues/1156


iHiD:
Nice :)

PR #2029, opened 2021-07-18
julia: Correct secret-handshake mentoring notes

cmcaine:


tracks/julia/exercises/secret-handshake/mentoring.md
```diff
@@ -95,7 +95,7 @@ end
 Samyak's solution, with a nice, simple for loop.
 
 Declaring `actions` as a const outside of the function avoids allocating it each time.
-If the tests accepted tuples of strings instead of just vectors then we could have declared `actions` within the function as a tuple instead.
+Samyak could have declared `actions` within the function as a tuple instead, which would give the same or better performance and avoids defining a top-level constant.
```
SaschaMann, line 98:
```suggestion
One could declare `actions` within the function as a tuple instead, which would give the same or better performance and avoids defining a top-level constant.
```

This may sound a bit friendlier because it doesn't "blame" Samyak. Feel free to merge as it is if you disagree, though

SaschaMann (approved):

cmcaine:
Thanks for the review! Please approve and merge :)

SaschaMann (approved):

PR #2043, opened 2021-08-22
Description for Elixir analyzer check: `list-ops`

jiegillet:
@angelikatyborska 

It's basically the comment in the stub.


angelikatyborska (approved):
👍

PR #2050, opened 2021-08-31
[Python] Analyzer Comments Syntax & Formatting

BethanyG:
Grouping these together for expediency -- and because Python only has the one exercise for analysis right now.

Having looked at these on the website, there were a few syntax, highlight, and language errors that needed fixing.  Hopefully, these will be the last  for general Pylint and `two-fer` comments.  😄


iHiD (approved):
(Not read them)

PR #2053, opened 2021-09-02
javascript/matrix: suggest to defer value calculation to it's getter

salouri:
Many students try to do everything inside the constructor and that is not the best practice.

### DIFF
--- a/tracks/javascript/exercises/matrix/mentoring.md
+++ b/tracks/javascript/exercises/matrix/mentoring.md
@@ -70,7 +70,7 @@ input.split('\n')
     )
   })
 ```
- -If a student try to do everything inside the constructor suggest on them to defer calculations to getters as they are supposed to defer the cost of calculating the value until it is needed. If the value never needed, you never pay the cost.[[2](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)] 
+- If a student tries to implement everything inside a constructor, suggest on them to defer calculations to getters as they are supposed to defer the cost of calculating the result until it is needed. [[2](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)] 
 
 ## Talking points
 - Functions are first class citizens, so you can pass one to `Array#map` and it executes on each one
--- a/tracks/javascript/exercises/matrix/mentoring.md
+++ b/tracks/javascript/exercises/matrix/mentoring.md
@@ -70,6 +70,7 @@ input.split('\n')
     )
   })
 ```
+ -If a student try to do everything inside the constructor suggest on them to defer calculations to getters as they are supposed to defer the cost of calculating the value until it is needed. If the value never needed, you never pay the cost.[[2](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)] 
 
 ## Talking points
 - Functions are first class citizens, so you can pass one to `Array#map` and it executes on each one
--- a/tracks/javascript/exercises/matrix/mentoring.md
+++ b/tracks/javascript/exercises/matrix/mentoring.md
@@ -70,7 +70,7 @@ input.split('\n')
     )
   })
 ```
-- If a student tries to implement everything inside a constructor, suggest on them to defer calculations to getters as they are supposed to defer the cost of calculating the result until it is needed. [[2](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)] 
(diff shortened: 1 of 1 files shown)


tracks/javascript/exercises/matrix/mentoring.md
```diff
@@ -70,6 +70,7 @@ input.split('\n')
     )
   })
 ```
+ -If a student try to do everything inside the constructor suggest on them to defer calculations to getters as they are supposed to defer the cost of calculating the value until it is needed. If the value never needed, you never pay the cost.[[2](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)]
```
tejasbubane, line 73: I would suggest a few minor changes:

```suggestion
- If a student tries to implement everything inside a constructor, suggest on them to defer calculations to getters as they are supposed to defer the cost of calculating the result until it is needed. [[2](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)] 
```

salouri: Thanks

junedev:
@salouri I am sorry but I don't agree with your argument. One could just as well make an argument the other way around. When the code to calculate the rows/columns is in the getter, it will be executed over and over again every time the getter is called. So in a real world scenario with a bigger matrix where you want to use the rows and columns throughout your code, that would lead to bad performance.

So since it depends on the context whether or not to do the calculation in the getter or the constructor (and the context is not provided by the exercise), I would not advise in either direction in the mentor notes but rather say that both version are reasonable.

salouri (commented):
> @salouri I am sorry but I don't agree with your argument. 

It's not just my argument, it's also the Mozilla Developer Network's argument. (link was attached)

>One could just as well make an argument the other way around. When the code to calculate the rows/columns is in the getter, it will be executed over and over again every time the getter is called. So in a real world scenario with a bigger matrix where you want to use the rows and columns throughout your code, that would lead to bad performance.

when you define a function to calculate a value, then it's meant for this value to be calculated every single time you call this function. Otherwise, there was no reason for the function itself to begin with. That's how it should be done. HOWEVER, if you are re-calculating a **constant** value over and over;  inside a function (e.g. getter), without checking first if the recalculation is needed then this would be the mistake of the developer and not the best-practice fault...such mistake will definitely  lead to bad performance. If such value acts as an attribute of the object and therefore needs to be available at initialization time (e.g. applies to "arrows" in our simple example) then it should be calculated inside the constructor. 
 
> 
> So since it depends on the context whether or not to do the calculation in the getter or the constructor (and the context is not provided by the exercise), I would not advise in either direction in the mentor notes but rather say that both version are reasonable.

I would always advice towards best practices, and let the student know /think about the benefits of such practice (e.g. unnecessary initialization overhead) so the programmer can decide for himself/herself whenever those benefits are over weighed by any other factor.

junedev:
> when you define a function to calculate a value, then it's meant for this value to be calculated 

Yes, but the exercise/tests do not ask you to define a function to calculate rows. They just expect a property rows and columns to be defined on the matrix.

As you mentioned, the underlying question is whether Matrix is supposed to be changeable or not. There is no indication about that in the description of the exercise and there are no methods that would modify the matrix. Hence my point about the constructor version being a reasonable solution.

The current mentor notes clearly state the constructor version as "approvable" solution and say that the student may use the getters (Line 41-52). In any case, if you are convinced the getter version is the superior solution here, you should also change that part of the notes.

With just the current PR change, the mentor notes will become inconsistent.

tracks/javascript/exercises/matrix/mentoring.md
```diff
@@ -70,6 +70,7 @@ input.split('\n')
     )
   })
 ```
+- If a student tries to implement everything inside a constructor, suggest on them to defer calculations to getters as they are supposed to defer the the cost of calculating the value until the value is needed. If it is never needed, you never pay the cost. [[2](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)]
```
junedev, line 73: Consistency issue with line 41-52, see https://github.com/exercism/website-copy/pull/2053#issuecomment-917553347

salouri: I agree that those lines  are inconsistent. They are inconsistent with the **exercise's start state** which clearly gives the user getter skeletons to solve the exercise with! While in the mentor notes it proposes a different solution and THEN (lines 50-52) says the user MAY use getters to solve the problem..

junedev: @salouri I wasn't aware the start state (stub file) was changed like that. (I am sorry, I was absent from Exercism for a while.) You are totally right, the mentor notes should be consistent with the stub. So if you could make this PR a somewhat bigger update so that the notes match the stub again, that would be great!

SleeplessByte:
@salouri removed their fork so I cannot make changes to this PR to get it merged.

PR #2055, opened 2021-09-03
elixir: Update variable_name_snake_case.md

cmcaine:
I'm not totally sure that the wrong_name will be camelCase, but it seems likely?

analyser bit: https://github.com/exercism/elixir-analyzer/blob/main/lib/elixir_analyzer/exercise_test/common_checks/variable_names.ex

### DIFF
--- a/analyzer-comments/elixir/solution/function_name_snake_case.md
+++ b/analyzer-comments/elixir/solution/function_name_snake_case.md
@@ -7,11 +7,11 @@
 Function, macro, and guard names are always written in snake_case in Elixir.
 
 ```elixir
-# snake_case
+# this is snake_case
 
 %{expected}
 
-# not snake_case
+# this is not snake_case
 
 %{actual}
 ```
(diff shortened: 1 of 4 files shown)


angelikatyborska:
Oh. I think the comment is not explicit enough and we have interpreted it in two opposite ways.

My intention was:

```elixir
# this is snake_case
do_something

# this is not snake_cake
# doSomething
```

But another interpretation, which I guess you had because of the change you're suggesting, is:
```elixir
# use snake_case
do_something

# do not use camelCase
# doSomething
```

We should make this more explicit. Can you make an additional change? I don't really care which version we use.

cmcaine:
That makes sense, updated :)

angelikatyborska (dismissed):

angelikatyborska:
Wait, I forgot about the other files! If we're changing the comment in one of them because it's confusing, we should update all. They all have this kind of comment:
- analyzer-comments/elixir/solution/function_name_snake_case.md
- analyzer-comments/elixir/solution/module_attribute_name_snake_case.md
- analyzer-comments/elixir/solution/module_pascal_case.md

angelikatyborska (approved):
Perfect, thanks!

PR #2056, opened 2021-09-04
julia: nucleotide-count: Add broadcasting Dict constructor

SaschaMann:
There's probably a neater way to do `collect("ACGT")`?

### DIFF
--- a/tracks/julia/exercises/nucleotide-count/mentoring.md
+++ b/tracks/julia/exercises/nucleotide-count/mentoring.md
@@ -11,7 +11,7 @@ We shouldn't require the mutating-a-dict strategy because it's actually pretty s
 
 (Formatted so you can copy and paste easily)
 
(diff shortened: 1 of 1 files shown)


cmcaine:
I'm not aware of a neater way to do that.

While this does work, I'm not sure we should encourage it: it's not all that easy to read and is a bit inefficient because it allocates two arrays.

Are you sure you want it in the notes?

If we do have it then as a separate bullet point would be preferable to me.

SaschaMann:
I find it a bit cleaner/easier to read and didn't consider the performance here. Maybe not the best idea though.

cmcaine:
The performance almost certainly doesn't matter because it's not in the loop, but I've nudged people towards the generator solutions most of the time to encourage good habits.

If broadcasting is clearer to you though, it's probably clearer to many other people too. So perhaps we should put it in despite my quibbles.

PR #2058, opened 2021-09-04
java/bird-watcher: Add a simple mentoring guide

jmrunkle:


tracks/java/exercises/bird-watcher/mentoring.md
```diff
@@ -0,0 +1,108 @@
...
+
+    private boolean isBusyDay(int birds) {
+        return birds >= 5;
+    }
+}
+```
+
+## Common suggestions
+
+- Leverage common Java libraries like `Math.min`
```
angelikatyborska, line 101: Are you sure those two suggestions (streams and `Math.min`) make sense for this exercise? It was designed as a learning exercise for arrays and for loops and as such, and the mentor's role would be to ensure that for loops are used.

tracks/java/exercises/bird-watcher/mentoring.md
```diff
@@ -0,0 +1,108 @@
...
+        int busyDaysCount = 0;
+        for (int dailyBirdCount : birdsPerDay) {
+            if (dailyBirdCount >= 5) {
+                busyDaysCount++;
+            }
+        }
+        return busyDaysCount;
+    }
+}
+```
```
angelikatyborska, line 54: This is a bit different than the exemplar solution https://github.com/exercism/java/blob/main/exercises/concept/bird-watcher/.meta/src/reference/java/BirdWatcher.java

For example the ternary operators aren't required to learn about arrays so you shouldn't tell students to use them. They might not know about them yet.

angelikatyborska (commented):
I'm leaving my 2 cents as a maintainer of another track that doesn't know Java 😉

tracks/java/exercises/bird-watcher/mentoring.md
jmrunkle: Good point - I guess I was viewing the exercise more from the perspective of what I would recommend for canonical Java code. I will remove these since the exemplar solution is now part of the mentoring view on these concept exercises.

tracks/java/exercises/bird-watcher/mentoring.md
jmrunkle: As my other comment said, I will just remove the Reasonable Solutions altogether since the exemplar is now part of the mentoring view.

PR #2060, opened 2021-09-05
Grade School: update mentoring.md

bobahop:
Changed `def roster(self, num):` to `def roster(self):`


BethanyG (approved):
Many thanks for this!  🌟

PR #2063, opened 2021-09-06
javascript/armstrong-numbers: Add the Mathematical approach to get the digits

salouri:
It's also valid for students to use the mathematical approach to get the digits array out of a number.


SleeplessByte:
@salouri removed their fork so I cannot make changes to this PR to get it merged.

PR #2066, opened 2021-09-08
Add mentoring notes for common-lisp roman-numerals

verdammelt:
Students either do not know about the built-in feature to do this arabic->roman numeral conversion or believe they should not be using it for this exercise.

This mentoring note clarifies that in the opinion of the track the built-in feature is *the* correct solution.

It also links to a short summary of why this feature exists in the language which I wrote.

Fixes https://github.com/exercism/common-lisp/issues/487


neenjaw (approved):

BethanyG (approved):
I think I have permissions?  Let's see.  Approved.  🌟

PR #2068, opened 2021-09-10
feat: :memo: add mentor notes for javascript/annalyns-infiltration

AndrewLawendy:


SleeplessByte:
@junedev can you check this out? I think we are better to advice "returning the result of an expression" than to talk about truthyness (until/unless boolean exercise) but maybe you disagree.

AndrewLawendy:
> @junedev can you check this out? I think we are better to advice "returning the result of an expression" than to talk about truthyness (until/unless boolean exercise) but maybe you disagree.

I'd like maybe to share a code snippet from one of my students as a context

```javascript
 if (knightIsAwake == true || archerIsAwake == true || prisonerIsAwake == true){
    return true;
  }else{
    return false;
  }
```

junedev:
@AndrewLawendy Sorry for the wait. We are currently still discussing how to handle mentor notes for concept exercises in general. The issue is that in the near future concept exercises will have quite thorough analyzers that can pick up exactly those points you mentioned on your notes and give automated feedback. Once the analyzer is in place, the student would already have seen this feedback and there would be no point for the mentor to write the same again. 

I am sorry comment on the site saying "send a PR via GitHub" was misleading here.

The best way forward regarding your valuable input would be if you can check whether it is already reflected in the ideas for the analyzer. We collect those in the design.md file of the exercise: https://github.com/exercism/javascript/blob/main/exercises/concept/annalyns-infiltration/.meta/design.md#analyzer If not, a PR there would be great. Feel free to also include your examples there.

Regarding you suggestion of thruthy and falsy as things to talk about: Those will be introduced later in the concept `type-conversion` so the student should not be pointed towards those already. Otherwise it defeats the point of having a syllabus that introduces things step by step. This exercise is deliberately designed in a way that you don't need to know anything about thruthy and falsy since only booleans are involved.

More generally, all the interesting pointers that are related to that concept should already be part of the `about` version of the concept that the student can see after unlocking the concept. Again, not necessary to repeat those in the mentor notes.

AndrewLawendy:
> @AndrewLawendy Sorry for the wait. We are currently still discussing how to handle mentor notes for concept exercises in general. The issue is that in the near future concept exercises will have quite thorough analyzers that can pick up exactly those points you mentioned on your notes and give automated feedback. Once the analyzer is in place, the student would already have seen this feedback and there would be no point for the mentor to write the same again.
> 
> I am sorry comment on the site saying "send a PR via GitHub" was misleading here.
> 
> The best way forward regarding your valuable input would be if you can check whether it is already reflected in the ideas for the analyzer. We collect those in the design.md file of the exercise: https://github.com/exercism/javascript/blob/main/exercises/concept/annalyns-infiltration/.meta/design.md#analyzer If not, a PR there would be great. Feel free to also include your examples there.
> 
> Regarding you suggestion of thruthy and falsy as things to talk about: Those will be introduced later in the concept `type-conversion` so the student should not be pointed towards those already. Otherwise it defeats the point of having a syllabus that introduces things step by step. This exercise is deliberately designed in a way that you don't need to know anything about thruthy and falsy since only booleans are involved.
> 
> More generally, all the interesting pointers that are related to that concept should already be part of the `about` version of the concept that the student can see after unlocking the concept. Again, not necessary to repeat those in the mentor notes.

Wow the Analyzer seems pretty interesting
Sure I understand
Love your way of doing guys :slightly_smiling_face:

PR #2069, opened 2021-09-10
feat: :memo: add mentor notes for javascript/triangle

AndrewLawendy:


SleeplessByte (commented):
Only 3 would be considered the magic number. What would you respond as mentor when a student uses `this.sides[3]`?

AndrewLawendy:
> Only 3 would be considered the magic number. What would you respond as mentor when a student uses `this.sides[3]`?

Why only 3 would be considered the magic number?
Most students sort the triangles so I think will be more readable this way `const [shortestLengthSide, middleLengthSide, longestLengthSide] = this.sides`

Maybe it's more crucial in the case of making sure no side is under 0 but I think it would make a good approach for them to learn anyway
```javascript
const [shortestLengthSide] = this.sides;
return shortestLengthSide !== 0;
```

AndrewLawendy:
> > Only 3 would be considered the magic number. What would you respond as mentor when a student uses `this.sides[3]`?
> 
> Why only 3 would be considered the magic number?
> Most students sort the triangles so I think will be more readable this way `const [shortestLengthSide, middleLengthSide, longestLengthSide] = this.sides`
> 
> Maybe it's more crucial in the case of making sure no side is under 0 but I think it would make a good approach for them to learn anyway
> 
> ```js
> const [shortestLengthSide] = this.sides;
> return shortestLengthSide !== 0;
> ```

@SleeplessByte any updates?

SleeplessByte (approved):

SleeplessByte:
> Why only 3 would be considered the magic number?

Because 0, 1, and 2 are often disregarded, see https://en.wikipedia.org/wiki/Magic_number_(programming)#Accepted_uses

PR #2072, opened 2021-09-14
Add `use_map_update` comment to Elixir high-score

Br1ght0ne:
Complements exercism/elixir-analyzer#154.

### DIFF
--- a/analyzer-comments/elixir/high-score/use_map_update.md
+++ b/analyzer-comments/elixir/high-score/use_map_update.md
@@ -1,11 +1,11 @@
 # use `Map.update/4`
 
 Use [`Map.update`] instead of [`Map.put`] + [`Map.get`] in `update_score` to
-update `scores` in a single call. For the updating function, use an
-[anonymous function] or the [`&` operator shorthand].
+update `scores` in a single call. For the updating function, define an
+anonymous function using [`fn`] or the [`&` operator shorthand].
 
 [`Map.update`]: https://hexdocs.pm/elixir/Map.html#update/4
 [`Map.put`]: https://hexdocs.pm/elixir/Map.html#get/3
 [`Map.get`]: https://hexdocs.pm/elixir/Map.html#put/3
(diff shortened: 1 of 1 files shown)


analyzer-comments/elixir/high-score/use_map_update.md
```diff
@@ -0,0 +1,11 @@
+# use `Map.update/4`
+
+Use [`Map.update`] instead of [`Map.put`] + [`Map.get`] in `update_score` to
+update `scores` in a single call. For the updating function, use an
+[anonymous function] or the [`&` operator shorthand].
+
+[`Map.update`]: https://hexdocs.pm/elixir/Map.html#update/4
+[`Map.put`]: https://hexdocs.pm/elixir/Map.html#get/3
+[`Map.get`]: https://hexdocs.pm/elixir/Map.html#put/3
+[anonymous function]: https://hexdocs.pm/elixir/Kernel.SpecialForms.html#fn/1
```
angelikatyborska, line 10: "anonymous function or ..." sounds like the second option is not an anonymous function, but that's not true. Both `fn` and `&` define anonymous functions. Maybe we can phrase it like this instead?
```suggestion
update `scores` in a single call. For the updating function, define an
anonymous function using [`fn`] or the [`&` operator shorthand].

[`Map.update`]: https://hexdocs.pm/elixir/Map.html#update/4
[`Map.put`]: https://hexdocs.pm/elixir/Map.html#get/3
[`Map.get`]: https://hexdocs.pm/elixir/Map.html#put/3
[`fn`]: https://hexdocs.pm/elixir/Kernel.SpecialForms.html#fn/1
```

Btw. I didn't know that you can skip the second `[]` when creating links like this! I would always write something like `[Map.update][map-udpate]`. Your version is convenient 🙂

Br1ght0ne: Thanks! There's something in Markdown for everyone ✨

angelikatyborska (approved):

PR #2075, opened 2021-09-18
Elixir Analyzer Comment: last line assignment

jiegillet:
Relates to [this PR](https://github.com/exercism/elixir-analyzer/pull/153).

### DIFF
--- a/analyzer-comments/elixir/solution/last_line_assignment.md
+++ b/analyzer-comments/elixir/solution/last_line_assignment.md
@@ -1,0 +1,5 @@
+# last line assignment
+
+You have assigned a variable on the last line of a function definition. This is unnecessary since there is no way to use the variable.
+
+Please note that an assignment on the last line may sometimes trigger specific analyzer comments. If you see other comments, start with removing the variable assignment.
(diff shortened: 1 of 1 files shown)


analyzer-comments/elixir/solution/last_line_assignment.md
```diff
@@ -0,0 +1,5 @@
+# last line assignment
+
+You have assigned a variable on the last line of a function definition. This is unnecessary since there is no way to use the variable.
+
+Please note that an assignment on the last line may sometimes trigger specific analyzer comments. If you see other comments, start with removing the variable assignment.
```
angelikatyborska, line 5: I'm missing some words signaling that those other comments should not have been triggered. Maybe like this?
```suggestion
Please note that an assignment on the last line may sometimes falsely trigger other analyzer comments. If you see other comments that don't make sense, start with removing the variable assignment.
```

jiegillet: Thank you, I was running in circles with this comment, your suggestion is great.

angelikatyborska (approved):

PR #2087, opened 2021-09-29
Add new analyzer comment for Elixir: compiler warnings

jiegillet:
Is the "naked" triple backticks OK? 
I'm not sure if it will show the colors right...

### DIFF
--- a/analyzer-comments/elixir/solution/compiler_warnings.md
+++ b/analyzer-comments/elixir/solution/compiler_warnings.md
@@ -1,0 +1,10 @@
+# compiler warnings
+
+[comment]: # (requires 1 param:)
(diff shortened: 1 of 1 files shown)


angelikatyborska:
That's a great question... I don't think it will. I have no idea if ANSI colors are even supported in analyzer output, but if they are, then probably only outside of code blocks. Maybe in the end we need to drop the colors after all?

angelikatyborska (approved):

jiegillet:
Yeah, I guess we can drop the colors... 🤷
Or try it out like this to and see what happens first?

angelikatyborska:
Let's drop them, it' usually just yellow or red for the words "warning" or something like that

PR #2092, opened 2021-10-01
Add analyzer comment on usage of `[h | t]` form

Br1ght0ne:


analyzer-comments/elixir/solution/list_prepend_head.md
```diff
@@ -0,0 +1,4 @@
+# use `[item | list]` to prepend items to lists
+
+Using `[item] ++ list` instead of `[item | list]` is longer and confusing (prepending to list that looks like list concatenation).
```
angelikatyborska, line 3: I totally agree it's "confusing" 😁 but we could use a slightly more objective adjective, what do you think?

```suggestion
Using `[item] ++ list` instead of `[item | list]` is longer and less intention-revealing (prepending to list that looks like list concatenation).
```

jiegillet (approved):
Nice!

PR #2094, opened 2021-10-03
Add analyzer comments for Elixir Captains Log

joseemds:


analyzer-comments/elixir/captains-log/use_rand_uniform.md
```diff
@@ -0,0 +1,3 @@
+# Use `rand.uniform` instead of `Enum.random`
+
+Besides learning about random, this exercise aims to teach you about Erlang libraries. Use `:rand.uniform` instead of `Enum.random` in the `format_stardate` function.
```
angelikatyborska, line 3: In this particular check, I feel like there is a better explanation on why to use `:rand.uniform` than just "to learn Erlang libraries". I think this will become obvious if we try to come up with a solution that uses `Enum.random` and still passes all the tests. `Enum.random` is great for choosing a random integer from a range, but you cannot have a range of floats. There is this unit test for this exercise:

https://github.com/exercism/elixir/blob/main/exercises/concept/captains-log/test/captains_log_test.exs#L86-L95

To make this test pass, one would need to choose a very big random integer and then divide it by something, to get a float with a decimal part. Or even worse, somebody might attempt to define a list of possible floats to choose from...

I would suggest something like this:
```suggestion
Use `:rand.uniform` instead of `Enum.random` in the `format_stardate` function. `:rand.uniform` allows you to generate a random float directly.
```

angelikatyborska (commented):
I left one suggestion. Otherwise it looks good, but before final approval I would need to see the actual implementation of the checks to be sure that the comments match the checks!

Are you also working on the PR for https://github.com/exercism/elixir-analyzer/ ?

joseemds:
Hi @angelikatyborska, thanks for your feedback! I commited the suggested changes and opened the PR with the work i have already done in the analyzer

jiegillet (commented):
I have one more comment. I don't know if people would see several warnings at one, but two of them start with "The purpose of this exercise is..." and proceed to mention different purposes :)
Maybe "One purpose of this exercise" would be more appropriate.

joseemds:
Hi @jiegillet thanks for your feedback, i agree with you, updated!

angelikatyborska (approved):

PR #2095, opened 2021-10-04
Add analyzer comments for Elixir common check: boolean functions

jiegillet:
[Sister PR here](https://github.com/exercism/elixir-analyzer/pull/180).

### DIFF
--- a/analyzer-comments/elixir/solution/def_with_is.md
+++ b/analyzer-comments/elixir/solution/def_with_is.md
@@ -18,4 +18,4 @@ Elixir has specific naming conventions which include how to name [functions that
 %{actual}
 ```
 
-Note that functions that return booleans and are valid in guards follow another convention and are named with a `is_` prefix (`defguard is_foo`). Both conventions must not mix.
+Note that guards, some built-in functions and macros that return booleans and are valid in guards follow another convention and are named with a `is_` prefix (`def is_integer`, `defguard is_foo`, `defmacro is_foo`). Both conventions must not mix.
--- a/analyzer-comments/elixir/solution/def_with_is.md
+++ b/analyzer-comments/elixir/solution/def_with_is.md
@@ -1,5 +1,9 @@
 # def with is_
 
+[comment]: # (requires 2 params:)
+[comment]: # (- expected: the corrected name)
+[comment]: # (- actual: the incorrect name that the student used)
+
 Elixir has specific naming conventions which include how to name [functions that return a boolean](https://hexdocs.pm/elixir/naming-conventions.html#trailing-question-mark-foo):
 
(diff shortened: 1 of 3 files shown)


analyzer-comments/elixir/solution/def_with_is.md
```diff
@@ -0,0 +1,21 @@
...
+# this is the expected name of functions that return a boolean
+
+%{expected}
+
+# this is the current name
+
+%{actual}
+```
+
+Note that functions that return booleans and are valid in guards follow another convention and are named with a `is_` prefix (`defguard is_foo`). Both conventions must not mix.
```
angelikatyborska, line 21: > Note that functions ...

Aren't we implementing a check that checks that no function is named `is_foo`?

I know that the docs say:

> However, functions that return booleans and are valid in guards follow another convention, described next.

Are we making a mistake in this analyzer check and it is in fact possible to have _functions_ (not macros) that can be used in guards? I couldn't get that to work. Or are the docs wrong?

jiegillet: Mmh I assumed reading that they meant "function" in the general sense, not in the `def` sense.

```elixir
iex(56)> defmodule A do
...(56)> def is_integer?(x), do: is_integer(x)
...(56)> def foo(x) when is_integer?(x), do: ok
...(56)> end
** (CompileError) iex:58: cannot find or invoke local is_integer?/1 inside guard. Only macros can be invoked in a guard and they must be defined before their invocation. Called as: is_integer?(x)
```
Seems like only macros (and guards) can be guards. It might be because the functions are not know at that step of the compilation, but macros and guards are. Maybe.

jiegillet: `Kernel` does it... 
https://github.com/elixir-lang/elixir/blob/2dcc7b6f50d37f60c25353373f081af8b46fac98/lib/elixir/lib/kernel.ex#L576-L579

jiegillet: https://keathley.io/blog/elixir-guard-clauses.html

> There’s no trivial way to verify functional purity in Erlang. Instead the creators decided to limit guard clauses to internal functions. That way they could ensure that these functions were both fast and pure.
> 
> This is why we get errors when we try to use our own functions in guard clauses. Erlang has no way of knowing if that function is safe. Instead it stops you from using the function at all.

If this is true, then our checks make sense, but we may need to rephrase some of the comments.

angelikatyborska: That explanation makes me happy!

angelikatyborska (approved):

PR #2098, opened 2021-10-06
Add analyzer comments for elixir doc_spec_order common check

joseemds:


analyzer-comments/elixir/solution/doc_spec_order.md
```diff
@@ -0,0 +1,16 @@
+# doc spec order
+
+[comment]: # (requires 2 params:)
+[comment]: # (- expected: How the doc and spec order should be)
+[comment]: # (- actual: The incorrect order that the student used)
+
+
+There is not a correct order between the `@doc` and `@spec` modules attributes. But the elixir community convention is to use `@doc` first and `@spec` next to the function
```
jiegillet, line 8:
```suggestion
Developers can choose the order of the `@doc` and `@spec` modules attributes, but the Elixir community convention is to use `@doc` first and `@spec` next to the function.
```

jiegillet (commented):
Nice!
I suggested a small change because the notion of "correct" seems too loaded to me.

jiegillet (commented):
I would still write an example below.

```elixir
@doc """
This function counts the number elements in a list
"""
@spec count(list(any())) :: integer()
def count(list), do: length(list)
```

joseemds:
Done!

jiegillet (approved):
Looks good!

analyzer-comments/elixir/solution/function_annotation_order.md
```diff
@@ -0,0 +1,12 @@
...
+
+Developers can choose the order of the `@doc` and `@spec` modules attributes, but the Elixir community convention is to use `@doc` first and `@spec` next to the function.
+
+Example:
+```elixir
+    @doc """
+    This function counts the number elements in a list
+    """
+    @spec count(list(any())) :: integer()
+    def count(list), do: length(list)
```
jiegillet, line 11: Any reason why this is indented?

joseemds: i thought that it would be better to read, but i think its not necessary, i will remove it

jiegillet (approved):
Perfect, thank you!

PR #2118, opened 2021-11-02
[Elixir] Update analyzer comment about private helper functions

angelikatyborska:
Not all exercises have stubs with function definitions. We need a different argument.

### DIFF
--- a/analyzer-comments/elixir/solution/private_helper_functions.md
+++ b/analyzer-comments/elixir/solution/private_helper_functions.md
@@ -18,4 +18,4 @@ In every Exercism exercise, the public interface is defined by the test suite. I
 %{actual}
 ```
 
-Note that when writing a library, it is often necessary to make functions public for use among modules. To hide such functions from the documentation and from library users, add `@doc false` before each function definition.
+Note that when writing a library, it is often necessary to make functions public for use among modules. To hide such functions from the documentation, add `@doc false` before the function definitions or `@moduledoc` false to hide the whole module.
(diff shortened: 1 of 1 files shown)


jiegillet (commented):
Yes, this is a better explanation.

After [reading this](https://hexdocs.pm/elixir/1.12/writing-documentation.html#hiding-internal-modules-and-functions) more carefully, I think we should also change the last sentence to something like:

> Note that when writing a library, it is often necessary to make functions public for use among modules. To hide such functions from the documentation, add `@doc false` before the function definitions or `@moduledoc false` to hide the whole module.

I had no idea about the last part in the link: 
> Functions starting with underscore [for example, `__add__/2`] are automatically treated as hidden

But I don't think we need to include that.

angelikatyborska:
> or @moduledoc false to hide the whole module.

Good point, I added that.

> But I don't think we need to include that.

I agree, it's a very obscure feature. It's good to know about because it appears in the standard library, but in my 4 years of writing Elixir code, I never had to use it myself.

analyzer-comments/elixir/solution/private_helper_functions.md
```diff
@@ -18,4 +18,4 @@ In this exercise, the public interface has already been provided by the stub fil
 %{actual}
 ```
 
-Note that when writing a library, it is often necessary to make functions public for use among modules. To hide such functions from the documentation and from library users, add `@doc false` before each function definition.
+Note that when writing a library, it is often necessary to make functions public for use among modules. To hide such functions from the documentation, add `@doc false` before the function definitions or `@moduledoc` false to hide the whole module.
```
jiegillet, line 21:
```suggestion
Note that when writing a library, it is often necessary to make functions public for use among modules. To hide such functions from the documentation, add `@doc false` before the function definitions or `@moduledoc false` to hide the whole module.
```

jiegillet (approved):
The backquote mark jumped off :)

jiegillet (approved):
I have to approve again? That's interesting...

PR #2121, opened 2021-11-07
[Elixir] Update `high-score` Analyzer comment

jiegillet:
As discussed [here](https://github.com/exercism/elixir-analyzer/issues/216).

Closes https://github.com/exercism/elixir-analyzer/issues/216

### DIFF
--- a/analyzer-comments/elixir/high-score/use_module_attribute.md
+++ b/analyzer-comments/elixir/high-score/use_module_attribute.md
@@ -1,3 +1,3 @@
 # use module attribute
 
-Use a [module attribute](https://elixir-lang.org/getting-started/module-attributes.html#as-constants) as a constant for the initial score.
+Use a [module attribute](https://elixir-lang.org/getting-started/module-attributes.html#as-constants) in `reset_score/2` as a constant for the initial score.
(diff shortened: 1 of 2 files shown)


analyzer-comments/elixir/high-score/use_module_attribute.md
```diff
@@ -1,3 +1,3 @@
 # use module attribute
 
-Use a [module attribute](https://elixir-lang.org/getting-started/module-attributes.html#as-constants) as a constant for the initial score.
+Use a [module attribute](https://elixir-lang.org/getting-started/module-attributes.html#as-constants) in `reset_score/2` as a constant for the initial score.
```
angelikatyborska, line 3: A person that didn't use the module attribute at all will also get this comment, so I think for them this comment would be more clear if it mentions both functions that need it. I know they will get the second comment about the default argument too, but that one focuses on the default argument part more.
```suggestion
Use a [module attribute](https://elixir-lang.org/getting-started/module-attributes.html#as-constants) in `add_player/3` and `reset_score/2` as a constant for the initial score.
```

angelikatyborska (approved):

analyzer-comments/elixir/high-score/use_module_attribute.md
jiegillet: I wrote it this way first, but then I realized that a person that didn't use any module attribute would also get the `high_score_use_default_argument_with_module_attribute` comment which mentions `add_player/3` specifically. This way, they would get two separate comments for two functions.

The checks for this exercise are quite convoluted.

jiegillet: Oh I'm sorry I read your comment too fast, you are aware. Then sure, we can add it.

angelikatyborska (approved):

PR #2123, opened 2021-11-08
Ruby exercises markdown styles

kotp:
**Do not squash**

This includes markdown style for single sentence per line and reference
links for the following mentor notes:

* acronym
* accumulate
* clock
* difference-of-squares
* grains
* hamming
* high-scores
* isogram
* leap
* luhn
* matrix
* pangram
* raindrops
* resistor-color-duo
* scrabble-score
* series
* tournament
* two-fer
* word-count

### DIFF
--- a/tracks/ruby/exercises/accumulate/mentoring.md
+++ b/tracks/ruby/exercises/accumulate/mentoring.md
@@ -1,12 +1,11 @@
 # Mentoring
 
-This exercise is asking you to re-implement `Array#map`. The goal is to get a
-feel for what's happening under the hood.
+This exercise is asking you to re-implement `Array#map`.
+The goal is to get a feel for what's happening under the hood.
 
 ## Reasonable Solutions
 
-There are 4 general classes of solutions that fall in a matrix
-`(functional | imperative) X ( high-level | low-level)`.
+There are 4 general classes of solutions that fall in a matrix `(functional | imperative) X ( high-level | low-level)`.
 
 ```ruby
 # High-level functional
@@ -62,6 +61,7 @@ end
 ```
 
 ### Talking Points
+
 - Block syntax (`yield` vs `block.call`)
 - All `Enumerable` methods can be implemented in terms of `each` or `reduce`
 - High vs low-level approaches
@@ -72,6 +72,8 @@ end
 Note that the student should _not_ use `Array#map` here as that defeats the
 purpose of this exercise. From the instructions:
 
-> Keep your hands off that collect/map/fmap/whatchamacallit functionality
-> provided by your standard library! Solve this one yourself using other basic
-> tools instead.
+> Keep your hands off that collect/map/fmap/whatchamacallit functionality provided by your standard library!
+> Solve this one yourself using other basic tools instead.
+
+## Changelog
+
--- a/tracks/ruby/exercises/acronym/mentoring.md
+++ b/tracks/ruby/exercises/acronym/mentoring.md
@@ -11,35 +11,45 @@ module Acronym
   end
 end
 ```
+
 with variations for the regex (see below).
 
 ## Common suggestions
-* Most first submissions split the phrase into words with `String#split`:
+
+- Most first submissions split the phrase into words with `String#split`:
+
   ```ruby
   phrase.split(/[ -]/)
   ```
+
   followed by operations to find the first letters.
   Variation: first `gsub` or `tr` the non-space delimiters, then `split`.
 
-In all these cases: it's recommended to suggest the `String#scan` method and, because the exercise is not about Regular Expressions, give them also the RegEx (`/\b[a-zA-Z]/`). (At this point in the track, it's more about experiencing the strengths of Ruby's built-in methods than making them find them. There's ample opportunity for research later on in the track.)
+In all these cases: it is recommended to suggest the `String#scan` method and, because the exercise is not about Regular Expressions, give them also the RegEx (`/\b[a-zA-Z]/`).
+(At this point in the track, it is more about experiencing the strengths of Ruby's built-in methods than making them find them.
+There's ample opportunity for research later on in the track.)
 
 ## Talking points
-* `Class vs Module`: If the solution uses Class instead of Module: that's totally acceptable at this stage. No reason to discuss.
-* Because of the regex needed for `split`, `scan` is in this exercise (marginally) faster.
-* It can be hard to catch all the possible delimiters.
-* `upcasing` the whole string is (10% ?) slower than `upcasing` the acronym only.
-* Store the regex in a constant or a variable, mostly so it can be named.
-* `join("")`: The `("")` after `join` can be omitted, because it's the default for `join`.
 
+- `Class vs Module`: If the solution uses Class instead of Module: that's totally acceptable at this stage.
+  No reason to discuss.
+- Because of the regex needed for `split`, `scan` is in this exercise (marginally) faster.
+- It can be hard to catch all the possible delimiters.
+- `upcasing` the whole string is (10% ?) slower than `upcasing` the acronym only.
+- Store the regex in a constant or a variable, mostly so it can be named.
+- `join("")`: The `("")` after `join` can be omitted, because it is the default for `join`.
 
 ## Mentoring strategy
-* When they start with `split`, with or without `tr` or `gsub`, and `map` to `upcase`: start with pointing them to `scan`, name what it accomplished afterwards:
-> 1) (first comment)
-  Using split works, but it forces you to focus on the delimiters (space, -). Ruby has `String#scan`. With `scan` and a regex (Regular Expression), you can catch the parts you need. For instance, `scan` with the regex `/\b[a-zA-Z]/` will catch 'word boundaries' (space, - and more) followed by a word character.
-I use "www.rubular.com" whenever I want to check a regex.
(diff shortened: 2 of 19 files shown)


tracks/ruby/exercises/accumulate/mentoring.md
```diff
@@ -1,32 +1,35 @@
...
 ```ruby
 # High-level functional
 
 class Array
+
   def accumulate
     reduce([]) { |arr, item| arr.append(yield item) }
   end
+
 end
```
iHiD, line 19: Could we maybe revert the adding whitespace lines? I don't think it adds anything for these simple examples, but means you can fit less info on the screen when reading the notes.

iHiD:
Thanks for this! Looks good in general. I'll try and find time to check it properly. Ping me again in a couple of days if I've not responded :)

tracks/ruby/exercises/accumulate/mentoring.md
kotp: Except that the last method is not tightly coupled to the class definition nor the end of the class definition?  So "white space as meaning"?

I could convert all methods that are single line methods to:

```ruby
def accumulate= reduce([]) { |memo, item| meno.append(yield item) }
```

… if we are worried about viewing real estate rather than "space around the things".

But yeah, up to you.  I think it communicates the right thing.

tracks/ruby/exercises/grains/mentoring.md
```diff
@@ -19,26 +20,37 @@ module Grains
...
+* With a non-maths solution, on `total` -> `inject` or `reduce` are sub-optimal, suggest `sum` (Ruby 2.4+)
+* If the solution uses iteration on the `square` method instead of using exponentiation, consider challenging the student to write out the sequence of grains and ask them if they see a pattern that may point to a simpler solution (1, 2, 4, 8, ...).
+* If the solution uses `64` in multiple places, suggest extracting the number to a constant.
+  Refactoring Gugu has an blog artible about [replacing magic numbers with symbolic constants][replace-magic-number-with-symbolic-constant] can be a good resource to share.
 
-- If the solution uses `64` in multiple places, suggest extracting the number to a constant. [This link](https://refactoring.guru/replace-magic-number-with-symbolic-constant) can be a good resource to share.
+## Talking points
 
+* Naming things is so relevant for this exercise.
+  What does the magic number 64 means?
```
kotp, line 47:
```suggestion
  What does the magic number 64 mean?
```

tracks/ruby/exercises/accumulate/mentoring.md
iHiD: > Except that the last method is not tightly coupled to the class definition nor the end of the class definition

I mean, it's tightly coupled in the sense it's part of it. And surely by the same logic the code inside the method isn't tightly coupled either, so there should be whitespace added within the method.

I don't see how adding the whitespace makes things more readable for a mentor. The counter-example you give clearly reduces the ability to see what's going on, so that makes it less readable. But others may have different perspectives. Let's see if anyone else weighs in.

kotp: Right.  Yeah, the white space in the methods was considered, but I went for the larger scope of that, treating the methods as units.  I would have personally added some white space in the method (and rearranged for organization as well) but it was too much.  So I left it to the smallest change (I considered) reasonable.

kotp:
🐪 for @iHiD  (if it were two bumps it would not have been a dromedary camel!)

iHiD:
Thanks - @ErikSchierboom Could you check this one pls?

tracks/ruby/exercises/accumulate/mentoring.md
```diff
@@ -72,6 +72,8 @@ end
 Note that the student should _not_ use `Array#map` here as that defeats the
 purpose of this exercise. From the instructions:
 
-> Keep your hands off that collect/map/fmap/whatchamacallit functionality
-> provided by your standard library! Solve this one yourself using other basic
-> tools instead.
+> Keep your hands off that collect/map/fmap/whatchamacallit functionality provided by your standard library! Solve this one yourself using other basic tools instead.
```
ErikSchierboom, line 75: Should this be:

```suggestion
> Keep your hands off that collect/map/fmap/whatchamacallit functionality provided by your standard library!
> Solve this one yourself using other basic tools instead.
```

?

tracks/ruby/exercises/acronym/mentoring.md
```diff
@@ -11,41 +11,55 @@ module Acronym
...
+
   ```ruby
   phrase.split(/[ -]/)
   ```
+
   followed by operations to find the first letters.
   Variation: first `gsub` or `tr` the non-space delimiters, then `split`.
 
-In all these cases: it's recommended to suggest the `String#scan` method and, because the exercise is not about Regular Expressions, give them also the RegEx (`/\b[a-zA-Z]/`). (At this point in the track, it's more about experiencing the strengths of Ruby's built-in methods than making them find them. There's ample opportunity for research later on in the track.)
+In all these cases: it is recommended to suggest the `String#scan` method and, because the exercise is not about Regular Expressions, give them also the RegEx (`/\b[a-zA-Z]/`).
```
ErikSchierboom, line 28:
```suggestion
In all these cases: it is recommended to suggest the `String#scan` method and, because the exercise is not about Regular Expressions, also give them the RegEx (`/\b[a-zA-Z]/`).
```

tracks/ruby/exercises/grains/mentoring.md
```diff
@@ -19,26 +19,36 @@ module Grains
   def self.total
     square(SQUARES.max) * 2 - 1
   end
+
```
ErikSchierboom, line 22: Personally, I'm not a fan of these additional empty lines (it won't block approval)

tracks/ruby/exercises/grains/mentoring.md
```diff
@@ -19,26 +19,36 @@ module Grains
...
 end
 
 class BoardPositionError < ArgumentError
+
```
ErikSchierboom, line 26: Personally, I'm not a fan of these additional empty lines (it won't block approval)

tracks/ruby/exercises/hamming/mentoring.md
```diff
@@ -1,48 +1,54 @@
...
 1. Iterating through both strings simultaneously via `Enumerable#zip`;
 2. Iterating explicit indices using a `Range`, `Integer#upto`, or `Integer#times`;
-3. A hybrid approach with `Enumerable#each_with_index`, iterating over one of the strings while also
-   tracking the index to find the equivalent character in the other string. Uses `Enumerable#each_with_index`.
-   (see notes below on `chars` vs `each_char`)
+3. A hybrid approach with `Enumerable#each_with_index`, iterating over one of the strings while also tracking the index to find the equivalent character in the other string.
+   Uses `Enumerable#each_with_index`.
+   (See notes below on `chars` vs `each_char`.)
+
+Counting approach on the other hand really only has one good solution:
```
ErikSchierboom, line 21:
```suggestion
A counting approach on the other hand really only has one good solution:
```

tracks/ruby/exercises/hamming/mentoring.md
```diff
@@ -85,23 +92,19 @@ the _Common Suggestions_ and _Talking Points_ sections below.
...
-constructs available that will automatically manage these for us.
-
-- I try to focus on one big thing in the first round of feedback and this is
-almost always it. I leave style and minor improvement suggestions for a quick
-final round at the end.
-
-- No matter how important Naming Things is, Hamming is not the best place to discuss the naming of the parameters,
-because there is no solution that seems to satisfy everyone, while a lot of people have strong opinions
-on it. The following few core exercises offer plenty opportunity to discuss naming.
+  manually managed intermediate counter and index variables.
```
ErikSchierboom, line 95: Should this be part of the previous line?

tracks/ruby/exercises/isogram/mentoring.md
```diff
@@ -10,7 +10,17 @@ end
...
+
+- Comparing lengths (`letters.uniq.length == letters.length`) is **marginally** quicker than comparing letters (`letters.uniq == letters`) but its more code.
+  Which is better?
+
+## Changelog
+
+- 2021 Nov - Markdown adjustment to style guide (sentences and links)
+
+[String#scan]: https://ruby-doc.org/core-2.2.0/String.html#method-i-scan
+    []
```
ErikSchierboom, line 26: What does this line do?

ErikSchierboom (approved):
Great work! Some minor comments

tracks/ruby/exercises/acronym/mentoring.md
ErikSchierboom: Sure, but I also saw some `it's` changed to `it is`, so 🤷

kotp: I considered that as well, but this is potentially a bit more complex as well.

Because: "suggest the banana and, maybe, ice cream" compared to "suggest the banana and, maybe, also ice cream", the also is really redundant here, though we have more words between the comma and the continuation.  I have to consider if the second is being suggested alternatively or additionally.  The also can be placed better, or may even be removed.

ErikSchierboom (approved):

PR #2124, opened 2021-11-09
Remove personal development environment ignores

kotp:
The `.gitignore` for the project should have things that this project should ignore, yet may create.

The "global" gitignore is for things for developers environment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,0 @@
-# ignore Mac OS temp file
-.DS_Store


angelikatyborska:
I agree 100% but I recall @SleeplessByte recently voiced the opposite opinion. If he doesn't have anything against this PR here, I would approve.

SleeplessByte:
In general I could agree with this, but we really never want that folder submitted right? So why should it not be in the project gitignore?

angelikatyborska:
Right, but having a gitignore with a single environment-specific file opens up the door for everyone to open PRs that add more environment-specific patterns, for their Jetbrains files, VSC files, vim swap files etc. etc., adding a bit more work for us to review all of that. In the end, it's basically just about giving a good example.

angelikatyborska (approved):

kotp:
Thanks for the review, and yes @angelikatyborska has it exactly right.

PR #2126, opened 2021-11-09
Add analyser comment for elixir.solution.unless_with_else

jlirochon:


jiegillet (commented):
Awesome, very clear. We should also add an actionable suggestion to use `if` and `else` instead.

analyzer-comments/elixir/solution/unless_with_else.md
```diff
@@ -0,0 +1,7 @@
+ # unless having else block
+ 
+ Unless conditions should avoid having an `else` block.
+ 
+ The reason for this is not a technical but a human one. The `else` in this
+ case will be executed when the condition is met, which is the opposite of
+ what the wording seems to apply.
```
angelikatyborska, line 7: I see a few small grammar problems here (also formatting - no need for the leading single space on every one)

- It's not the condition that needs to avoid "else", but the programmer.
- The verb "apply" doesn't fit in this context.

Here's my suggestion for a small rewording, already with @jiegillet suggestion applied (to mention `if`).

```suggestion
# unless having else block

Avoid using `unless` expressions with an `else` block. Use an `if` expression instead.

The reason for this is not a technical but a human one. The `else` in this
case will be executed when the condition is met, which is the opposite of
what the wording suggests.
```

jlirochon:
My comment was actually copied as is from credo, but your version is better.

jiegillet (approved):
Yes, better all around. Thank you all!

I think credo wrote "apply" instead of "imply". Maybe it's worth giving back with a PR on their side :)

PR #2132, opened 2021-11-16
Add mentor notes Kotlin/Anagram

Zanerast:


tracks/kotlin/exercises/anagram/mentoring.md
```diff
@@ -0,0 +1,19 @@
...
+    }
+
+    private fun String.order(): List<Char>{
+        return this.map { it.lowercaseChar() }.sorted()
+    }
+}
+```
+
+## Common suggestions
+* Function `sorted()` can be used to sort according to the characters natural sort order
```
SleeplessByte, line 19:
````suggestion

```kotlin
class Anagram(var anagram: String){
    fun match(list: List<String>): Set<String>{
        return list.filter {
            it.lowercase() != anagram.lowercase() && it.order() == anagram.order()
        }.toSet()
    }

    private fun String.order(): List<Char>{
        return this.map { it.lowercaseChar() }.sorted()
    }
}
```

## Common suggestions

* Function `sorted()` can be used to sort according to the characters natural sort order
````

SleeplessByte (dismissed):

SleeplessByte (approved):
Approving because something is better than nothing

PR #2135, opened 2021-11-18
java/isogram mentoring notes

cdpaiva:


tracks/java/exercises/isogram/mentoring.md
```diff
@@ -0,0 +1,40 @@
...
+    return true;
+  }
+}
+```
+
+## Talking points
+
+- Solutions that use the Collections Framework should opt for a Set interface to check for duplicates instead of Maps and Lists. Sets don’t allow duplicates, so they are best fit for this check.
+- The `add()` method in the Set interface returns a boolean value indicating if the value is already present or not. This can be used to stop code execution earlier, instead of building the entire Set and later on checking its size against the original String.
+- For brute force solutions that check every character against ever other character in the string (n² time complexity), you may point the strategies above to improve efficiency.
```
kotp, line 39:
```suggestion
- For brute force solutions that check every character against every other character in the string (n² time complexity), you may point the strategies above to improve efficiency. 
```

tracks/java/exercises/isogram/mentoring.md
```diff
@@ -0,0 +1,40 @@
...
+  }
+}
+```
+
+## Talking points
+
+- Solutions that use the Collections Framework should opt for a Set interface to check for duplicates instead of Maps and Lists. Sets don’t allow duplicates, so they are best fit for this check.
+- The `add()` method in the Set interface returns a boolean value indicating if the value is already present or not. This can be used to stop code execution earlier, instead of building the entire Set and later on checking its size against the original String.
+- For brute force solutions that check every character against ever other character in the string (n² time complexity), you may point the strategies above to improve efficiency. 
+- One common algorithm to check for duplicates is to use an array as a hand-made counter. Each character points to an index of the array and that element is incremented according to the characters frequency. This is  fine for the tests, but it tends to overlook any non-letter characters. It is worth to note that Strings in Java are not limited to ASCII chars and the code might break for other characters.
```
kotp, line 40:
```suggestion
- One common algorithm to check for duplicates is to use an array as a hand-made counter. Each character points to an index of the array and that element is incremented according to the characters frequency. This is fine for the tests, but it tends to overlook any non-letter characters. It is worth to note that Strings in Java are not limited to ASCII characters and the code might break for other characters.
```

kotp (commented):
Spacing, typo, and consistent use of _character_.

tracks/java/exercises/isogram/mentoring.md
```diff
@@ -0,0 +1,40 @@
...
+        }
+    }
+    return true;
+  }
+}
+```
+
+## Talking points
+
+- Solutions that use the Collections Framework should opt for a Set interface to check for duplicates instead of Maps and Lists. Sets don’t allow duplicates, so they are best fit for this check.
```
kotp, line 37:
```suggestion
- Solutions that use the Collections Framework should opt for a Set interface to check for duplicates instead of Maps and Lists. Sets don’t allow duplicates, so they are the best fit for this check.
```

cdpaiva:
Thank you very much for the revision @kotp!

tracks/java/exercises/isogram/mentoring.md
```diff
@@ -0,0 +1,40 @@
...
+  }
+}
+```
+
+## Talking points
+
+- Solutions that use the Collections Framework should opt for a Set interface to check for duplicates instead of Maps and Lists. Sets don’t allow duplicates, so they are the best fit for this check.
+- The `add()` method in the Set interface returns a boolean value indicating if the value is already present or not. This can be used to stop code execution earlier, instead of building the entire Set and later on checking its size against the original String.
+- For brute force solutions that check every character against every other character in the string (n² time complexity), you may point the strategies above to improve efficiency. 
+- One common algorithm to check for duplicates is to use an array as a hand-made counter. Each character points to an index of the array and that element is incremented according to the characters frequency. This is fine for the tests, but it tends to overlook any non-letter characters. It is worth to note that Strings in Java are not limited to ASCII characters and the code might break for other characters.
```
kotp, line 40: The test code might break or the student solution might break?  How might the code break?  Also, letters are a specific kind of character, so in an Isogram solution, I would expect that only letters are collected (though non-letters can be given).  So I might say that it should "overlook any non-letter characters".

tracks/java/exercises/isogram/mentoring.md
```diff
@@ -0,0 +1,40 @@
...
+    return true;
+  }
+}
+```
+
+## Talking points
+
+- Solutions that use the Collections Framework should opt for a Set interface to check for duplicates instead of Maps and Lists. Sets don’t allow duplicates, so they are the best fit for this check.
+- The `add()` method in the Set interface returns a boolean value indicating if the value is already present or not. This can be used to stop code execution earlier, instead of building the entire Set and later on checking its size against the original String.
+- For brute force solutions that check every character against every other character in the string (n² time complexity), you may point the strategies above to improve efficiency.
```
kotp, line 39:
```suggestion
- For brute force solutions that check every character against every other character in the string (n² time complexity), you may point to the strategies above to improve efficiency. 
```

tracks/java/exercises/isogram/mentoring.md
```diff
@@ -0,0 +1,40 @@
...
+    }
+    return true;
+  }
+}
+```
+
+## Talking points
+
+- Solutions that use the Collections Framework should opt for a Set interface to check for duplicates instead of Maps and Lists. Sets don’t allow duplicates, so they are the best fit for this check.
+- The `add()` method in the Set interface returns a boolean value indicating if the value is already present or not. This can be used to stop code execution earlier, instead of building the entire Set and later on checking its size against the original String.
```
kotp, line 38:
```suggestion
- The `add()` method in the Set interface returns a Boolean value indicating if the value is already present or not. This can be used to stop code execution earlier, instead of building the entire Set and later on checking its size against the original String.
```

tracks/java/exercises/isogram/mentoring.md
```diff
@@ -0,0 +1,40 @@
...
+        }
+    }
+    return true;
+  }
+}
+```
+
+## Talking points
+
+- Solutions that use the Collections Framework should opt for a Set interface to check for duplicates instead of Maps and Lists. Sets don’t allow duplicates, so they are the best fit for this check.
```
kotp, line 37:
```suggestion
- Solutions that use the Collections Framework should opt for a Set interface to check for duplicates instead of Maps and Lists. Sets do not allow duplicates, so they are the best fit for this check.
```

tracks/java/exercises/isogram/mentoring.md
cdpaiva: Since the return is a primitive boolean, I think we should keep the word in lowercase. `Boolean` would imply a wrapper object.

tracks/java/exercises/isogram/mentoring.md
cdpaiva: There are special characters used as letters that are not limited to the ASCII range, which is usually the choice for the array size. If any of these letters is used, the student solution will break.
Since this type of solution is very rare in the java track, I will not add it to the mentoring notes.

tracks/java/exercises/isogram/mentoring.md
kotp: Then should it be wrapped in a code block marking?  `boolean` because it is a primitive Boolean, as opposed to a wrapper named `Boolean` that is treated as a Boolean?

tracks/java/exercises/isogram/mentoring.md
kotp: The comment was only to identify which code might break, as it only says "the code might break" and does not indicate the test code or the student solution code.  I feel like it might be helpful in the note to specify which code is being at risk. If it is that the test code might break, then we should probably fix that.

cdpaiva: Sure, I see your point. I was referring to the student's solution. I ended up removing this paragraph from these notes, since I felt it would be rarely used.

tracks/java/exercises/isogram/mentoring.md
cdpaiva: Exactly, it is a primitive. There is a clear distinction in Java between `boolean` and `Boolean`. I'll add the code markings, thanks for the suggestion.

kotp (approved):
Looks good to me, though could use single sentence per line.  If not taken care of here, I might do so after this is in.

kotp:
Since I suggested that I might make the changes once this is in, if not taken care of here, I have created #2170 to address it (and more).

ericbalawejder:
Ok, thank you. It was sitting around for 3 months so I wanted to move it along.

PR #2141, opened 2021-12-03
Use reference links for Python mentoring notes

IsaacG:


IsaacG:
@kotp This "mass" ref link cleanup?

Also, I can't add assignees so ... @exercism/python folks?

kotp (approved):

PR #2150, opened 2022-01-22
Add mentoring/python/perfect-number

IsaacG:
@BethanyG  for approval.


BethanyG:
Given that this exercise also requires that the `ValueError` message thrown matches the one in the tests, we may want to note that for mentors, in case they encounter confusion or frustration from students who don't know why their thrown error might still be failing.  See https://github.com/exercism/python/blob/main/exercises/practice/perfect-numbers/perfect_numbers_test.py for the error message we are checking for.  

Updated the code example to reflect that as well.

BethanyG (approved):

PR #2151, opened 2022-01-22
Add mentoring/python/pangram

IsaacG:


IsaacG:
@BethanyG  for approval

BethanyG:
@IsaacG  -- I will push an update to this that omits the typehints.  I know we've let some slip in the past -- but we should avoid them even in mentor notes since we don't want to give mentors nor students the impression that we are using `mypy` or _**requiring**_ that they use typehints.  Longer term, we will have to figure out what we might want to do - but for now we shouldn't use them in any documentation or stubs.

IsaacG:
Type hints dropped.

BethanyG (approved):

Issue #2155, opened 2022-02-05
[sublist] Correct spelling mistake in mentoring notes


Timo4B:
"procudes" should be "produces"

PR #2156, opened 2022-02-05
[sublist] Correct spelling mistake in mentoring notes

Timo4B:
"procudes" changed to "produces" according to #2155


BethanyG (approved):
Thank you for this.  😄

PR #2162, opened 2022-02-22
julia/mentoring: Super tiny formatting updates

icweaver:
Hi all, just following up on the slack discussion about our mentoring notes here. My takeaway from @cmcaine it is that these notes are meant to be easy to share in mentoring sessions by preserving the original markdown formatting. With that in mind, I tried to keep the changes as conservative as possible to preserve this intention, but happy to make any additional changes if I've misunderstood!

Thanks for putting together such a neat resource for all us, it is really helpful!


cmcaine (approved):
Thanks for these changes!

kotp (approved):

PR #2163, opened 2022-02-22
elixir/analyzer: add'l context for German Sysadmin warning about String functions

hunkyjimpjorps:
Just saw a mentoring request that was confused about why their solution that used `~c` charlist sigils was rejected by the analyzer (since it's a essentially just a macro for String.to_charlist); this note should give some guidance to anyone else who runs into the same problem

### DIFF
--- a/analyzer-comments/elixir/german-sysadmin/no_string.md
+++ b/analyzer-comments/elixir/german-sysadmin/no_string.md
@@ -1,3 +1,3 @@
 # no string
 
-The purpose of this exercise is to teach charlists. Solve it without using strings - do not use `to_string`, `to_charlist`, the `String` module, the charlist sigil `~c` or the bitstring special form `<<>>`.
+The purpose of this exercise is to teach charlists. Solve it without using strings - do not use `to_string`, `to_charlist`, the `String` module, the charlist sigil `~c` (it uses `to_charlist` under the hood), or the bitstring special form `<<>>`.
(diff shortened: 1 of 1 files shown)


analyzer-comments/elixir/german-sysadmin/no_string.md
```diff
@@ -1,3 +1,3 @@
 # no string
 
-The purpose of this exercise is to teach charlists. Solve it without using strings - do not use `to_string`, `to_charlist`, the `String` module, or the bitstring special form `<<>>`.
+The purpose of this exercise is to teach charlists. Solve it without using strings - do not use `to_string`, `to_charlist`, the `String` module, the charlist sigil `~c` or the bitstring special form `<<>>`.
```
angelikatyborska, line 3: > since it's a essentially just a macro for String.to_charlist

Huh, I had no idea! It looks like that's true: https://github.com/elixir-lang/elixir/blob/v1.13.3/lib/elixir/lib/kernel.ex#L5694-L5696

It might not be obvious because at a first glance it looks like `~c(ae)` would work the same as `'ae'`. Maybe we can add a clarification (and also my beloved [oxford comma](https://en.wikipedia.org/wiki/Serial_comma) back 😇)
```suggestion
The purpose of this exercise is to teach charlists. Solve it without using strings - do not use `to_string`, `to_charlist`, the `String` module, the charlist sigil `~c` (it uses `to_charlist` under the hood), or the bitstring special form `<<>>`.
```

hunkyjimpjorps: Yeah, I'm cool with that.

angelikatyborska (approved):
Perfect, thank you 🙏

PR #2164, opened 2022-02-24
java/two-fer: update mentoring notes

cdpaiva:
The third reasonable solution import is incorrect. Changed to the proper import.


angelikatyborska (approved):
I'm approving and merging based on @ericbalawejder approval

PR #2166, opened 2022-03-04
Add an additional suggestion to Python/kindergarten-garden

DanielNoord:
I found this in one of the submitted solutions and I think it is good to make mentors aware of this. Considering that most mentors probably use linters etc. they might not immediately spot this mistake as they are not used to having to look for this error (at least that's what happened when I first took a look at the submitted solution).


BethanyG (approved):
@DanielNoord   Thank you for submitting this.

PR #2177, opened 2022-03-20
Updated work experience information

chivalry:


IsaacG:
I could be mistaken but I believe these files are no longer used for anything.

angelikatyborska:
@IsaacG You are correct. Those files are no longer used.

@chivalry To update your bio that shows up on the website, go to https://exercism.org/settings

PR #2180, opened 2022-03-26
Add Elm Analyzer comments for common rules: `NoUnused` and `Simplify`

jiegillet:
@ceddlyburge @mpizenberg 
Sister PR [to this one](https://github.com/exercism/elm-analyzer/pull/29).

### DIFF
--- a/analyzer-comments/elm/common/no_unused/custom_type_constructor_args.md
+++ b/analyzer-comments/elm/common/no_unused/custom_type_constructor_args.md
@@ -1,6 +1,7 @@
 # custom type constructor arguments
 
 [comment]: # (This error is raised when a type constructor argument is unused)
+[comment]: # (Phantom types do not trigger this message)
 [comment]: # (requires 1 param:)
 [comment]: # (- definition: code extract with unused value)
 
--- a/analyzer-comments/elm/common/no_unused/custom_type_constructor_args.md
+++ b/analyzer-comments/elm/common/no_unused/custom_type_constructor_args.md
@@ -1,0 +1,13 @@
+# custom type constructor arguments
+
+[comment]: # (This error is raised when a type constructor argument is unused)
+[comment]: # (requires 1 param:)
+[comment]: # (- definition: code extract with unused value)
+
+The type constructor argument indicated below is not used anywhere.
+You could remove it or use it somewhere.
+
+```
+%{definition}
+```
+
--- a/analyzer-comments/elm/common/no_unused/parameters.md
+++ b/analyzer-comments/elm/common/no_unused/parameters.md
@@ -1,6 +1,7 @@
 # parameters 
 
 [comment]: # (This error is raised when a parameter is unused)
+[comment]: # (Some parameters, like `_ as x` cannot be used)
 [comment]: # (requires 1 param:)
 [comment]: # (- definition: code extract with unused value)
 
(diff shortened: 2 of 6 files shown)


analyzer-comments/elm/common/no_unused/custom_type_constructor_args.md
```diff
@@ -0,0 +1,13 @@
+# custom type constructor arguments
+
+[comment]: # (This error is raised when a type constructor argument is unused)
+[comment]: # (requires 1 param:)
+[comment]: # (- definition: code extract with unused value)
+
+The type constructor argument indicated below is not used anywhere.
```
ceddlyburge, line 7: Does this rule fire for phantom types? I assume not but would be good to be sure!

analyzer-comments/elm/common/no_unused/parameters.md
```diff
@@ -0,0 +1,13 @@
+# parameters 
+
+[comment]: # (This error is raised when a parameter is unused)
+[comment]: # (requires 1 param:)
+[comment]: # (- definition: code extract with unused value)
+
+The parameter indicated below is not used anywhere.
+You could remove it or use it somewhere if it can be used.
```
ceddlyburge, line 8: The copy here is different to the previous examples ...

analyzer-comments/elm/common/simplify.md
```diff
@@ -0,0 +1,12 @@
+# parameters 
+
+[comment]: # (This error is raised when a parameter is unused)
```
ceddlyburge, line 3: Is this comment correct?

analyzer-comments/elm/common/no_unused/custom_type_constructor_args.md
jiegillet: I double checked and it does not fire. It fires only for constructor arguments.

analyzer-comments/elm/common/no_unused/parameters.md
jiegillet: Yes, that's on purpose.

For cases like 
```elm
foo =
    let
        ( left, right ) =
            tupleValue
    in
    left
```
Then `right` could be used somewhere, but here
```elm
foo =
    let
        ( _, _ ) =
            bar
    in
    1
```
the pattern `(_, _)` cannot actually be used anywhere.

That's why I say "or use it somewhere if it can be used" in this case. For other rules, like `NoUnused.CustomTypeConstructorArgs`, it's different, whatever argument it catches could always potentially be used.

analyzer-comments/elm/common/simplify.md
jiegillet: Oh, not at all! Thank you

analyzer-comments/elm/common/no_unused/custom_type_constructor_args.md
ceddlyburge: Cool, could you mention that somewhere, in the comment or something?

analyzer-comments/elm/common/no_unused/parameters.md
ceddlyburge: Cool, could you mention that somewhere, in the comment or something?

ceddlyburge (approved):

angelikatyborska (approved):

PR #2191, opened 2022-05-17
[Clojure]: reverse-string: Mentor note update

mbjarland:
In my mind `(apply str (reverse s))` expresses the intent more clearly.

I think using what is arguably an implementation detail of list (i.e. that it appends at the start) is fine but not as explicit and obvious as `(apply str (reverse s))`.

### DIFF
--- a/tracks/clojure/exercises/reverse-string/mentoring.md
+++ b/tracks/clojure/exercises/reverse-string/mentoring.md
@@ -10,6 +10,13 @@ This exercise is about re-implementing a common library function.
 (ns reverse-string)
 
 (defn reverse-string [s]
+  (apply str (reverse s)))
+```
+
+```clojure
+(ns reverse-string)
+
+(defn reverse-string [s]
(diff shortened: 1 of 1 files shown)


tracks/clojure/exercises/reverse-string/mentoring.md
```diff
@@ -9,6 +9,13 @@ This exercise is about re-implementing a common library function.
 ```clojure
 (ns reverse-string)
 
+(defn reverse-string [s]
+  (apply str (reverse s)))
+```
+
```
kotp, line 15: What you wrote in the PR description might be helpful to add to the mentoring notes, as to why one might prefer this over another solution.  And what comments might be made while mentoring this exercise.

SleeplessByte, line 15:
```suggestion

One may prefer this because the above expresses the intent more clearly.
```

SleeplessByte (approved):

PR #2206, opened 2022-07-01
added mentor notes for grains

eklatzer:
Hey, as discussed in https://github.com/exercism/go/issues/2250 I added the part about the integer overflow to the mentoring notes of the exercise grains for the Golang track. I am not very satisfied with the result, but I am opening the PR so we can discuss the mentoring notes for this exercise. I do not know if the headline `Talking points` fits for the integer overflow topic. @junedev What are your recommendations for the mentoring notes for this exercise.

### DIFF
--- a/tracks/go/exercises/grains/mentoring.md
+++ b/tracks/go/exercises/grains/mentoring.md
@@ -1,0 +1,73 @@
+# Grains
+
+_Grains_
+
+## Reasonable Solutions
+
+This can be solved by calculating `2^n` (e.g. with `math.Pow`) or by using the shift operator `<<`. This solution uses the shift operator:
+
+```
+package grains
+
+import (
+	"fmt"
+)
+
+// Square returns the number of grains on a square on a chess board where the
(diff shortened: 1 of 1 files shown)


junedev (commented):
I would recommend to put the explanation about the overflow under "Common suggestions" instead of "Talking Points". Talking points is meant for "if the solution is good, you can mention this as further reading" which does not really apply to your content.

When you phrase your content as "Please watch out, there can be wrong implementations here that pass all the tests. ..." that would make it fit the "Common Suggestions" category even if it might not be all that common.

junedev (approved):

junedev:
@eklatzer Before merging, I made some adjustments:
* One sentence per line.
* Added the language name to the code blocks so we get the correct syntax highlighting.
* Removed the exalysis reference as the tool is not up-to-date anymore afaik and does not support anything "grains" specific.
* Moved a sentence from the end to the beginning to "set the stage" better.

PR #2207, opened 2022-07-02
Create mentoring.md

dsletten:


SleeplessByte:
@exercism/java please check!

sanderploegsma:
This is good feedback, but is actually a bug in the exercise itself. I checked the [reference implementation](https://github.com/exercism/java/blob/main/exercises/practice/rational-numbers/.meta/src/reference/java/Rational.java#L105) and it turns out it uses a completely different implementation compared to the exercise instructions.

@dsletten would you be willing to submit a PR to https://github.com/exercism/java to address this? I think the reference implementation should be updated to use

```java
Math.pow(Math.pow(base, numerator), 1.0/denominator);
```

And the tolerance in the tests should be lowered to the point where this implementation passes the tests.

PR #2214, opened 2022-08-15
[csharp/anagram] A reasonable solution for the anagram

michalporeba:
Adding a single reasonable solution for the anagram problem in the csharp track.

### DIFF
--- a/tracks/csharp/exercises/anagram/mentoring.md
+++ b/tracks/csharp/exercises/anagram/mentoring.md
@@ -1,23 +1,28 @@
 ### Reasonable solutions: 
 
+
+```csharp
 public class Anagram
 {
-  private readonly IEnumerable<char> orderedLetters;
-
-  public Anagram(string word)
-  {
-    orderedLetters = word.ToLower().OrderBy(character => character);
-  }
-  
-  public string[] FindAnagrams(string[] words)
-    => words.Where(IsAnagram).ToArray(); 
-  
-  private bool IsAnagram(string candidate)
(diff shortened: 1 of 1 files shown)


tracks/csharp/exercises/anagram/mentoring.md
```diff
@@ -0,0 +1,23 @@
...
+  
+  private bool IsAnagram(string candidate)
+    => Enumerable.SequenceEqual(candidate.ToLower().OrderBy(c => c), orderedLetters);
+}
+  
+### Common suggestions: Good suggestions specific to this exercise. Good lessons that emerge from it.
+  
+-- consider the cost of generating strings for comparison
+  
+### Talking points: Questions to challenge more advance learners with.
```
SleeplessByte, line 23:
````suggestion

```csharp
public class Anagram
{
  private readonly IEnumerable<char> orderedLetters;

  public Anagram(string word)
  {
    orderedLetters = word.ToLower().OrderBy(character => character);
  }
  
  public string[] FindAnagrams(string[] words)
    => words.Where(IsAnagram).ToArray(); 
  
  private bool IsAnagram(string candidate)
    => Enumerable.SequenceEqual(candidate.ToLower().OrderBy(c => c), orderedLetters);
}
```

### Common suggestions
  
- consider the cost of generating strings for comparison
  
````

SleeplessByte (commented):
@exercism/csharp okay with the reasonable solution?

ErikSchierboom:
@SleeplessByte I would suggest to add a separate method for `.ToLower().OrderBy(character => character)`, which is called in two places.

SleeplessByte (approved):

PR #2218, opened 2022-09-02
Elixir captains-log: Separate "use" and "do not use" comments

angelikatyborska:
Analyzer PR: https://github.com/exercism/elixir-analyzer/pull/328


jiegillet (approved):
Perfect!
I like how detailed we can make the messages and keep them accurate.

PR #2225, opened 2022-09-20
Create mentoring.md

jclstefan:
added mentoring notes for difference of squares

### DIFF
--- a/tracks/dart/exercises/difference-of-squares/mentoring.md
+++ b/tracks/dart/exercises/difference-of-squares/mentoring.md
@@ -1,0 +1,38 @@
+# Mentoring
+
+## Reasonable solutions
+
+We need to have two three methods:
+
+- the first one is sqareOfSum that will return the sum of all numbers and squared afterwards
+- the second method is the sumOfSquares which returns the sum of sqared numbers
+- the last method - differenceOfSquares - returns the difference of both of the methods above
+
+This exercise allows students to be tidy in coding their solution. Putting everything in methods that can be used in other methods is welcomed practice.
+
+```dart
+class DifferenceOfSquares {
+  int squareOfSum(int n) => pow(List<int>.generate(n, (i) => i + 1).reduce((a,b) => a+b), 2);
+  int sumOfSquares(int n) => List<int>.generate(n, (i) => pow(i+1, 2)).reduce((a,b) => a+b);
+  int differenceOfSquares(int n) => squareOfSum(n) - sumOfSquares(n);
+}
+```
+
+
+## Common suggestions
+
+- they need to create three methods
+- use [generate-method][reference-generate-method] to generate a List of incremented numbers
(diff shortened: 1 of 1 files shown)


siebenschlaefer:
Do we really recommend the "manual" approach where we numbers from `1` to `n` are generated and summed up?   
There's a direct and much more efficient solution, and I think that's what the last paragraph in the instructions is all about ("*You are not expected to discover an efficient solution to this yourself from first principles; research is allowed, indeed, encouraged. Finding the best algorithm for the problem is a key skill in software engineering.*")

kotp:
The question of "How do other tracks solve this exercise" is easy to answer, as you can look in each tracks repository and find their solution in each of the files.

For each track that has the exercise you will find the solution in:

    <language>/exercises/practice/difference-of-squares/.meta/example.*

kotp:
```dart
import "dart:math" show pow;

class DifferenceOfSquares {
  int squareOfSum(int input) => pow(_sum(input), 2).toInt();

  int sumOfSquares(int input) => _range(input).map((i) => pow(i, 2).toInt()).reduce((r, i) => r + i);

  int differenceOfSquares(int input) => squareOfSum(input) - sumOfSquares(input);

  num _sum(int input) => input * (input + 1) ~/ 2;

  List<int> _range(int length) => new List<int>.generate(length, (i) => i + 1);
}
```

This is how the `example.dart` looks currently, and it iterates, rather than using the formula that can be used.

siebenschlaefer:
In a discussion two or three years ago I was told that the example solution from the `.meta` directory is not necessarily the best (easiest to read, most idiomatic, most efficient) solution, it's more like proof that the exercise can be solved, that it's possible to pass the tests.  

I just wanted to comment that an efficient implementation exists and to draw attention to the [last paragraph](https://github.com/exercism/problem-specifications/blob/main/exercises/difference-of-squares/description.md) of the instructions, IMHO they clearly hint to such an efficient implementation.

But I have no skin in this game, I'm neither a Dart mentor nor do I know what Dart programmers would consider idiomatic. I'm fine with whatever you guys decide for these Dart mentoring notes.

kotp:
> In a discussion two or three years ago I was told that the example solution from the `.meta` directory is not necessarily the best (easiest to read, most idiomatic, most efficient) solution, it's more like proof that the exercise can be solved, that it's possible to pass the tests.

This is true for this exercise, if this is a practice exercise and not a concept exercise.  I think that concept exercises should have an exemplar solution.  I think, also, that there is the difference of that specific file being named, for practice, `example.ext`<sup>1</sup>, and for concept exercises, `exemplar.ext`<sup>1</sup>.

---

**1**: replacing, of course, the `ext` for the proper extension.

SleeplessByte:
@smanookian do you feel like making changes to the notes based on the discussion above?

PR #2248, opened 2023-01-29
javascript resistor-color-duo: fix student given empty comments

angelikatyborska:
This PR is a response to this bug report from the forum: https://forum.exercism.org/t/analyzer-claims-to-have-recommendation-but-reveals-nothing/3085

A student submitted a solution that triggers the comment:
```
 "externalTemplate": "javascript.resistor-color-duo.prefer_number_over_parse"
```

But didn't see any text on the platform. I am no expert in the JS track, but I think this has something to do with the old experimental analyzer meant for mentors and its comments being in "wrong" directory from the perspective of the current analyzers.

I don't know if this is the right fix attempt (maybe some of the analyzer checks should be removed instead?), but this PR moves all the resistor-color-duo comments from the nested `./commentary` directory to the right place, removing some phrasing that was clearly meant for mentors (see commit history).


SleeplessByte:
I'll check this out

PR #2250, opened 2023-02-25
Add elixir analyzer comment for no-rescue

angelikatyborska:
Analyzer PR: https://github.com/exercism/elixir-analyzer/pull/369

I couldn't find any good article about this to link to.

### DIFF
--- a/analyzer-comments/elixir/solution/no_rescue.md
+++ b/analyzer-comments/elixir/solution/no_rescue.md
@@ -1,0 +1,25 @@
+# no rescue
+
+Do not use exceptions for flow control. Try to solve this exercise without using `rescue`.
+
+If you're using a standard library function whose name ends with `!` (raises errors), find an equivalent function.
+
+Note: If you're seeing this automated comment on your solution, the other automated comments might not be accurate. Fix this problem first!
+
+```elixir
+# preferred: for example `case`
+
+case Date.new(1999, 13, 1) do
+  {:ok, date} -> date
+  {:error, _} -> nil
+end
+
+# discouraged: rescuing exceptions
+
+try do
+  Date.new!(1999, 13, 1)
+rescue
+  ArgumentError ->
+    nil
+end
+```
--- a/analyzer-comments/elixir/solution/no_rescue.md
+++ b/analyzer-comments/elixir/solution/no_rescue.md
@@ -1,10 +1,12 @@
 # no rescue
 
-Do not use exceptions for flow control. Try to solve this exercise without using `rescue`.
+Do not use exceptions for flow control.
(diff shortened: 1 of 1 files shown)


analyzer-comments/elixir/solution/no_rescue.md
```diff
@@ -0,0 +1,25 @@
+# no rescue
+
+Do not use exceptions for flow control. Try to solve this exercise without using `rescue`.
```
jiegillet, line 3:
```suggestion
Do not use exceptions for flow control.
Try to solve this exercise without using `rescue`.
```

analyzer-comments/elixir/solution/no_rescue.md
```diff
@@ -0,0 +1,25 @@
...
+
+If you're using a standard library function whose name ends with `!` (raises errors), find an equivalent function.
```
jiegillet, line 5:
```suggestion
If you are using a standard library function with a name ending with a `!` (raises errors), look for an equivalent function without the `!` (often return `{:ok, result}` or `{:error, message}`).
```

analyzer-comments/elixir/solution/no_rescue.md
```diff
@@ -0,0 +1,25 @@
...
+
+Note: If you're seeing this automated comment on your solution, the other automated comments might not be accurate. Fix this problem first!
```
jiegillet, line 7:
```suggestion
Note: if you are seeing this automated comment on your solution, other automated comments might not be accurate.
Fix this problem first!
```

analyzer-comments/elixir/solution/no_rescue.md
```diff
@@ -0,0 +1,25 @@
...
+
+```elixir
+# preferred: for example `case`
+
+case Date.new(1999, 13, 1) do
+  {:ok, date} -> date
+  {:error, _} -> nil
+end
+
+# discouraged: rescuing exceptions
```
jiegillet, line 17: Should we also mention that something like this would raise
```elixir
{:ok, date} = Date.new(1999, 13, 1)
```
Or does it feel like a different topic?

jiegillet (commented):
Nice, I left some wording suggestions.

analyzer-comments/elixir/solution/no_rescue.md
angelikatyborska: It does feel like a different topic. This comment is only triggered by usages of `rescue`, it doesn't really matter how the error is raised... the goal is to prevent `rescuing`.

jiegillet (approved):
Wonderful, thank you!

PR #2251, opened 2023-03-09
Create mentoring.md for Square Root objective in Python.

ConstLuck:
I noticed this was still blank, thought this could be a good start for mentoring notes. Please feel free to come up with other suggestions to add before this gets merged.

### DIFF
--- a/tracks/python/exercises/square-root/mentoring.md
+++ b/tracks/python/exercises/square-root/mentoring.md
@@ -1,1 +1,26 @@
-Consider guiding towards a solution in which none of the existing Python library functions are used, but rather one in which students create the logic/algorithm theirselves. Try and suggest different approaches and perhaps introduce them to benchmarking methods, and see for themselves the impact of the different approaches. 
+# Mentoring
+Consider guiding towards a solution in which none of the existing Python library functions are used, but rather one in which students create the logic/algorithm themselves.
+
+## Problem and challenges
+
+## Reasonable solution
+
+This solution would be more expected of new Pythonista's, be that new to programming or coming from a different
+language. Also, this is solution is encouraged so that the students come up with their own logic on how to tackle
+this without the use of a library.
+
+// TODO
+
+## Compact solution
+
+One of the easier ways for this exercise to be resolved is utilizing the `math` library's `sqrt()` method. However, this will only be valid for usage with a number greater than or equal to 0.
+
+```python
+def square_root(number: int) -> float:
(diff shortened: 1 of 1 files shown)


IsaacG:
You may want to look at other mentoring files and reuse structure from another as a template. As a first pass, it should be possible to fill in more than this. 

Are there specific standard libraries students might be using here? Is there a reason they shouldn't be used here?

Benchmarking isn't specific to this exercise in any way so it's not relevant to the mentoring notes for this exercise; instead that's more of a general mentoring choice/philosophy. 

I don't think there's value in "just putting anything here as a first step" over leaving this empty; if we're adding mentoring notes, they should be filled out.

ConstLuck:
The commit 'flesh out the mentoring notes more' was also done by me, I still had my old github account setup on my other device. Giving out this info to avoid confusion.

tracks/python/exercises/square-root/mentoring.md
```diff
@@ -0,0 +1,26 @@
+# Mentoring
+Consider guiding towards a solution in which none of the existing Python library functions are used, but rather one in which students create the logic/algorithm themselves.
```
IsaacG, line 2:
```suggestion
# Mentoring

Consider guiding towards a solution in which none of the existing Python library functions are used, but rather one in which students create the logic/algorithm themselves.
```

Is there a particular library/function that should be avoided? Why should it be avoided? This is rather vague and I'm not sure how mentors are supposed to use it.

tracks/python/exercises/square-root/mentoring.md
```diff
@@ -0,0 +1,26 @@
...
+
+```python
+def square_root(number: int) -> float:
+  return math.sqrt(number)
+```
+
+## Talking points
+
+### Exploration
+Perhaps we can guide the learner towards exploring other solutions and the differences between them. Where should or could they look for more information about a library. If they submitted a solution without using `math.sqrt()`, show themthe compact solution and it's constraints.
```
IsaacG, line 26: This seems like a placeholder. If mentors should be demonstrating how to derive the square root using other methods, show or link to other methods. Having a "mentors should show how to solve this" without details about *how* to solve it ... this doesn't add any value.

IsaacG (changes requested):

ConstLuck:
Closed branch as I realized that my suggestions didn't really add value, it might as well be left untouched until a better mentoring guide is written.

PR #2270, opened 2023-09-28
Add homersimpsons (myself) to PHP automator

homersimpsons:
I did build https://github.com/exercism/php-representer

I also do mentoring on the JS track, but I guess it has more attention.

### DIFF
--- a/automators.json
+++ b/automators.json
@@ -20,5 +20,6 @@
   },
   { "username": "habere-et-dispertire", "tracks": ["raku"] },
   { "username": "ee7", "tracks": ["nim", "zig"] },
-  { "username": "jonmcalder", "tracks": ["r"] }
+  { "username": "jonmcalder", "tracks": ["r"] },
+  { "username": "homersimpsons", "tracks": ["php"] }
 ]


IsaacG (approved):
On the one hand, you did build it. On the other, you've only mentored 25 students. I would highly recommend going through a few more mentoring sessions prior to adding any automated feedback. Having experience with different students and approaches can be helpful when it comes to writing one-off feedback.

homersimpsons:
Thank you for the approval @IsaacG 

> On the other, you've only mentored 25 students

I did mentor 39 solutions but I first I was closing the mentoring sessions when the student send me a "thank you message" and those didn't count. I also closed some without answers after months to cleanup my workspave (it looks like since yesterday those get now closed with timeout).

> I would highly recommend going through a few more mentoring sessions prior to adding any automated feedback

I will be careful with the tool. At first the idea is more to check how the representer behaves, in fact I have a local setup to download the first 500 submissions then run them through the representer and open only the different ones to see if there is any change I can do to the representer, but having this directly integrated in Exercism feel way better.

> Having experience with different students and approaches can be helpful when it comes to writing one-off feedback.

I agree

PR #2273, opened 2023-10-02
Remove suggestions from Elixir's Lasagna

kahgoh:
The function naming, regular expression and String suggestions aren't applicable to the exercise anyway.

Related post on forums: https://forum.exercism.org/t/wrong-mentoring-notes-in-elixirs-lasagna/7546


iHiD:
Thank you!

PR #2282, opened 2023-10-26
Elixir square root: mention not to use **

angelikatyborska:
Related to https://github.com/exercism/elixir-analyzer/issues/405


jiegillet (approved):
One might argue that `such as` implies there are others, but ok :)

PR #2283, opened 2023-10-28
Add Elm Analyzer comment for practice exercise `zebra-puzzle`

jiegillet:
[Sister PR over here](https://github.com/exercism/elm-analyzer/pull/60).

@ceddlyburge could you check this?


ceddlyburge (approved):
It is tempting! But you wouldn't learn anything doing it that way so also pointless :)

jiegillet:
> It is tempting! But you wouldn't learn anything doing it that way so also pointless :)

Haha, technically you do get Exercism points! ... Which is also pointless :p

@exercism/reviewers could I have an approval please? 💛

PR #2288, opened 2023-10-31
Elm Analyzer: tweak `top-scorers` comment

jiegillet:
@ceddlyburge could you take a look?

You originally wrote
> Maybe to something like this:
'aggregateScorers is most easily written using updateGoalCountForPlayer and List.foldl (list.foldr also works, but it a bit les ergonomic).'

But `foldl` and `foldr` have the same signature, so there is no change in ergonomy. There might be some efficiency difference, but it's not obvious and I don't want to get into that in an exercise about dictionaries.


ceddlyburge (approved):

jiegillet:
@exercism/reviewers an I get an approval please? 💛

ErikSchierboom (approved):

kotp (approved):

PR #2308, opened 2024-01-06
Update mentoring.md

RomaLetodiani:


IsaacG:
What makes this solution better than the prior one -- for the purpose of mentors?

There are many ways to solve this exercise, and mentors can browse the community solutions if they want to look at other solutions. Does the prior solution not pass the tests? Is there something wrong with it?

RomaLetodiani:
Just wanted to contribute, sry to disturb

IsaacG:
Contributions are great! But the change should improve something specific and be objectively better. Having everyone update the solution to their solution just to change it doesn't add much value and adds works for maintainers.

RomaLetodiani:
Thx for the feedback i'll close the PR now

PR #2311, opened 2024-01-10
add mentoring notes for go/pig-latin

bernot-dev:
Add some mentoring notes for the Pig Latin exercise.

### DIFF
--- a/tracks/go/exercises/pig-latin/mentoring.md
+++ b/tracks/go/exercises/pig-latin/mentoring.md
@@ -1,0 +1,72 @@
+# Pig Latin
+
+### Reasonable solutions:
+
+There is a (somewhat) reasonable solution using regular expressions, however most students are likely to use a combination of `strings` funtions or string slicing to solve this problem.
+
+Solutions vary considerably, but most reasonable solutions should demonstrate these three insights:
+
+1. Decoupled logic to break down a sentence into words and reform the Pig Latin sentence.
+
+```
+func Sentence(sentence string) string {
+	words := strings.Fields(sentence)
+	pigWords := make([]string, 0, len(words))
+	for _, w := range words {
+		pigWords = append(pigWords, pigWord(w))
+	}
+	return strings.Join(pigWords, " ")
+}
+```
+
+2. Identify that some number of consonants (according to the rules) will be added moved to the end of the word and "ay" will be added to the end of the word.
+
+```
+func pigWord(word string) string {
+	consonants := consonantCount(word)
+	return fmt.Sprintf("%s%say", word[consonants:], word[:consonants])
+}
+```
+
+3. Identify how many letters count as consonants without overly convoluted logic.
+
+```
+func consonantCount(word string) int {
+	if strings.HasPrefix(word, "xr") || strings.HasPrefix(word, "yt") {
+		return 0
+	}
+	for i, r := range word {
+		switch r {
+		case 'a', 'e', 'i', 'o':
+			return i
+		case 'u':
+			if i > 0 && word[i-1] == 'q' {
+				// When a 'qu' is together, the u is also moved to the end of the word
+				return i + 1
+			}
+			return i
+		case 'y':
+			if i == 0 {
+				// Leading y is not considered a vowel
+				continue
+			}
+			return i
+		}
+	}
+	return 0
+}
+```
+
+It is possible to combine parts 2 and 3, but this often leads to less readable code that repeats parts of the logic to form the pig latin word.
+
+### Common suggestions:
+* Avoid complex, deeply nested logic.
+* Are all of the conditions treated as special cases in code actually special cases? For instance, does a leading vowel need to be treated differently than a first vowel in the middle of the word?
+* Prefer switch/case to long if/else chains.
+* Add explanatory comments when the intent of the code might not be clear to the reader.
+* `gofmt` ensures consistent formatting, and is standard across most of the Go community.
+
+### Talking points
+* `rune` vs `byte` and why iterating over a string with `range` returns runes: [Strings, bytes, runes and characters in Go](https://blog.golang.org/strings)
+* The test cases do not include uppercase letters. How would the student's solution have to change to handle uppercase letters?
+* `fmt.Sprintf` and `strings.Builder` are more common than using the `+` operator to concatenate strings.
--- a/tracks/go/exercises/pig-latin/mentoring.md
+++ b/tracks/go/exercises/pig-latin/mentoring.md
@@ -61,7 +61,7 @@ It is possible to combine parts 2 and 3, but this often leads to less readable c
 
 ### Common suggestions:
 * Avoid complex, deeply nested logic.
-* Are all of the conditions treated as special cases in code actually special cases? For instance, does a leading vowel need to be treated differently than a first vowel in the middle of the word?
+* Reduce the number of special cases by generalizing logic.  For instance, the first vowel in the word can potentially be handled the same, regardless of whether it is at the start of the word or not.
 * Prefer switch/case to long if/else chains.
 * Add explanatory comments when the intent of the code might not be clear to the reader.
 * `gofmt` ensures consistent formatting, and is standard across most of the Go community.
--- a/tracks/go/exercises/pig-latin/mentoring.md
+++ b/tracks/go/exercises/pig-latin/mentoring.md
@@ -28,7 +28,7 @@ func pigWord(word string) string {
 }
 ```
 
-3. Identify how many letters count as consonants without overly convoluted logic.
+3. Identify how many letters count as consonants.
 
 ```
 func consonantCount(word string) int {


tracks/go/exercises/pig-latin/mentoring.md
```diff
@@ -0,0 +1,72 @@
+# Pig Latin
+
+### Reasonable solutions:
```
IsaacG, line 3: This should probably be `##` and not `###` based on a spot check of other notes

tracks/go/exercises/pig-latin/mentoring.md
```diff
@@ -0,0 +1,72 @@
...
+
+Solutions vary considerably, but most reasonable solutions should demonstrate these three insights:
+
+1. Decoupled logic to break down a sentence into words and reform the Pig Latin sentence.
+
+```
+func Sentence(sentence string) string {
+	words := strings.Fields(sentence)
+	pigWords := make([]string, 0, len(words))
+	for _, w := range words {
```
IsaacG, line 15: Are single-letter variables idiomatic Go code?

tracks/go/exercises/pig-latin/mentoring.md
```diff
@@ -0,0 +1,72 @@
...
+2. Identify that some number of consonants (according to the rules) will be added moved to the end of the word and "ay" will be added to the end of the word.
+
+```
+func pigWord(word string) string {
+	consonants := consonantCount(word)
+	return fmt.Sprintf("%s%say", word[consonants:], word[:consonants])
+}
+```
+
+3. Identify how many letters count as consonants without overly convoluted logic.
```
IsaacG, line 31:
```suggestion
3. Identify how many letters count as consonants.
```

The subjective convolution measure probably isn't actually relevant.

tracks/go/exercises/pig-latin/mentoring.md
```diff
@@ -0,0 +1,72 @@
...
+
+```
+func consonantCount(word string) int {
+	if strings.HasPrefix(word, "xr") || strings.HasPrefix(word, "yt") {
+		return 0
+	}
+	for i, r := range word {
```
IsaacG, line 38: More readable var names might be good here.

tracks/go/exercises/pig-latin/mentoring.md
```diff
@@ -0,0 +1,72 @@
...
+				continue
+			}
+			return i
+		}
+	}
+	return 0
+}
+```
+
+It is possible to combine parts 2 and 3, but this often leads to less readable code that repeats parts of the logic to form the pig latin word.
```
IsaacG, line 60: "often" sounds subjective. Why not leave it up to the mentors to notice if the code is readable or not and to call out repeated code?

```suggestion
It is possible to combine parts 2 and 3, though sometimes doing so is hard.
```

tracks/go/exercises/pig-latin/mentoring.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Common suggestions:
+* Avoid complex, deeply nested logic.
+* Are all of the conditions treated as special cases in code actually special cases? For instance, does a leading vowel need to be treated differently than a first vowel in the middle of the word?
```
IsaacG, line 64:
```suggestion
* Reduce the number of special cases by generalizing logic.  For instance, the first vowel in the word can potentially be handled the same, regardless of whether it is at the start of the word or not.
```

tracks/go/exercises/pig-latin/mentoring.md
```diff
@@ -0,0 +1,72 @@
...
+* Prefer switch/case to long if/else chains.
```
IsaacG, line 65: Is this personal preference or idiomatic Go?

tracks/go/exercises/pig-latin/mentoring.md
```diff
@@ -0,0 +1,72 @@
...
+* Add explanatory comments when the intent of the code might not be clear to the reader.
+* `gofmt` ensures consistent formatting, and is standard across most of the Go community.
```
IsaacG, line 67: These two are general comments and not specific to this exercise.

thekaranchauhan (approved):
This solution best practices, featuring clear and decoupled logic. It also effectively handles various cases, such as 'qu' combinations and leading 'y.'

tracks/go/exercises/pig-latin/mentoring.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Talking points
+* `rune` vs `byte` and why iterating over a string with `range` returns runes: [Strings, bytes, runes and characters in Go](https://blog.golang.org/strings)
+* The test cases do not include uppercase letters. How would the student's solution have to change to handle uppercase letters?
+* `fmt.Sprintf` and `strings.Builder` are more common than using the `+` operator to concatenate strings.
```
IsaacG, line 72:
```suggestion
* `strings.Builder` is the preferred tool for building large strings. `fmt.Sprint` can be useful in place of string concatenation for shorter strings.
```

tracks/go/exercises/pig-latin/mentoring.md
```diff
@@ -0,0 +1,72 @@
+# Pig Latin
+
+### Reasonable solutions:
+
+There is a (somewhat) reasonable solution using regular expressions, however most students are likely to use a combination of `strings` funtions or string slicing to solve this problem.
```
IsaacG, line 5: This section has a bunch of components but not a complete solution. You may want to present a complete solution here. If you want to explain it, you can use functions and comments and delve into details in following sections.

tracks/go/exercises/pig-latin/mentoring.md
thekaranchauhan: What my seniors says is, it is idiomatic for short lived variables and loops (in small snippets as well).   However, for variables with broader scopes and long lifetimes, descriptive names are appreciated.

iHiD:
@bernot-dev Thank you for this! :)

tracks/go/exercises/pig-latin/mentoring.md
bernot-dev: This is normative, and has associated linter rules. For instance:

- [GoCritic ifElseChain rule](https://go-critic.com/overview.html#ifelsechain)
- [StaticCheck QF1003 - Convert if/else-if chain to tagged switch](https://staticcheck.dev/docs/checks#QF1003)

tracks/go/exercises/pig-latin/mentoring.md
bernot-dev: Yeah, the other notes in the Go track seem to disagree with [the README for the repo](https://github.com/exercism/website-copy/blob/main/README.md?plain=1#L24), which suggests `###` for headings. I wasn't sure which to follow.

tracks/go/exercises/pig-latin/mentoring.md
bernot-dev: [The greater the distance between a name's declaration and its uses, the longer the name should be](https://go.dev/talks/2014/names.slide#4). In this case, it is scoped to just a couple lines, so I consider this idiomatic.

tracks/go/exercises/pig-latin/mentoring.md
bernot-dev: `i` is a [well-accepted name for an index variable](https://go.dev/talks/2014/names.slide#6).

`r` is short for `rune`, and rune is a keyword. I'm open to suggestions here, but I wouldn't question this in code review, myself.

tracks/go/exercises/pig-latin/mentoring.md
bernot-dev: The note about adding comments is more likely to come up in this exercise than many others, because the logic is actually somewhat complex here.

Formatting is general, but comes up frequently for me. There are similar notes on other exercises on this track. I'm not sure what the right call is here.

tracks/go/exercises/pig-latin/mentoring.md
bernot-dev: I'm not too familiar with using suggested changes in GitHub, but can you modify the suggest to clarify "in place of string concatenation with the `+` operator"?

I feel like that's ambiguous, otherwise.

tracks/go/exercises/pig-latin/mentoring.md
bernot-dev: The three parts/functions together form a complete solution. I can try to reorganize if you think that would help it flow better.

tracks/go/exercises/pig-latin/mentoring.md
bernot-dev: Most of the community solutions I saw that count consonants inline end up with more deeply logic and many returns at various levels of organization. It's not wrong and there may be marginal performance benefits, but it does reduce [Clarity](https://google.github.io/styleguide/go/guide.html#clarity), which is the most important style goal of code, according to Google's Style Guide.

IsaacG (approved):

PR #2319, opened 2024-01-27
[typescript/resistor-color] Create mentoring.md

mendoncamaria:
Created mentoring notes for resistor-color exercise in typescript


IsaacG:
This looks more like you're explaining how to solve this exercise to a student. Are there discussion points for a mentor here?

If the only thing this file provides is a solution, the mentors can already look at the example solution, the community solutions, or, usually, their own solution. 

The file format doesn't line up with other mentor notes. You may want to look at other mentor notes and make sure the file has a similar structure and similar type of content.

SleeplessByte (commented):
I agree with @IsaacG. These are not mentoring notes unfortunately.

PR #2322, opened 2024-02-02
[python/yacht] Create mentoring.md

F4s0lix:
more like sample program but I think it will help some people

### DIFF
--- a/tracks/python/exercises/yacht/yacht/mentoring.md
+++ b/tracks/python/exercises/yacht/yacht/mentoring.md
@@ -1,0 +1,70 @@
+# Mentoring
+## score categories
+They should be stored in variables at the start of a file and used in if statement like below.
+```python
+ONES = 1
+TWOS = 2
+...
+def score(dice, category):
+  if category == ONES:
+    ...
+```
+## categories ONES to SIXES
+Good solution is 
+```python
+return sum(value for value in dive if value == X)
+```
+where X is number corresponding to category.
+But better solution with less code repetition is
+```python
+if category in (ONES, ..., SIXES):
+    return dice.count(category) * category
+```
+but you must make proper score categories
+## FULL_HOUSE
+Ideal solution is with using collections and Counter
+```python
+dice_values = collections.Counter(dice).values()
+# 'dict' with items quantity
+if set(dice_values) == {2, 3}: 
+    # set is sorted, unique elements
+    # three of one number and two of another 
+    return sum(dice)
+```
+## LITTLE_STRAIGHT and BIG_STRAIGHT
+```python
+set(dice) == {1, 2, 3, 4, 5} # little
+set(dice) == {2, 3, 4, 5, 6} # big
+# set will return sorted, unique values
+```
+## CHOICE
+```python
+sum(dice) # it's sum of the dice
+```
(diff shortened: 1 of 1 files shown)


IsaacG:
This should be targeted at mentors, not students.

Mentors typically already solved the exercise, know how to read code and know how to browse the community solutions. Can you add discussion notes, common gotchas, talking points, etc?

F4s0lix:
I rewrited everything using [example from README](https://github.com/exercism/website-copy/blob/main/tracks/ruby/exercises/isogram/mentoring.md) and Mentore notes section from README.
If you have any other suggestions I will try to implement it

SleeplessByte (commented):
I think the file is in the wrong place.

@exercism/python want to check the contents?

BethanyG:
@SleeplessByte  -- you are correct.  The path should be `tracks/python/exercises/yacht/mentoring.md`.  There is an extra level in the path that needs to be removed.

This is also quite sparse for mentoring notes.  It's missing a setup of the problem and is very light on suggestions and talking points.

 It also recommends a 'reasonable solution' that .....**_isn't_**.   We should never encourage students to assign a name to a `lambda` in Python.  It screws with stack traces, debugging, and the help system.   At the very least, there should be talking points around the "controversy", and details as to why this is a bad/not bad idea, what alternatives there are, and what circumstances (_if any at all_) might lead to (acceptably) doing it.

Additionally, there are other ways to solve this problem idiomatically that are not covered by these notes.  While we don't need to be exhaustive with solutions, we do need to provide some reasonable variation and strategy for mentors to look at and suggest.  I would recommend looking at/for community solutions that do not use `lambdas`, but do use strategies beyond the `if/elif/else` block.

As it stands, I don't think we should accept this PR.

@F4s0lix -- please take a look at this [acronym mentoring notes](https://github.com/exercism/website-copy/blob/main/tracks/python/exercises/acronym/mentoring.md) doc and  alternatively [this](https://github.com/exercism/website-copy/blob/main/tracks/python/exercises/allergies/mentoring.md) doc as examples of what we're looking for.  You can find additional mentoring note variations for Python [here](https://github.com/exercism/website-copy/tree/main/tracks/python/exercises).

As mentioned above, these notes are geared toward mentors who have previously solved the problem, and are looking for tips, hints, and strategies for discussing the problem with students in a productive and helpful manner.  

The focus is on walking a mentor through:
1.   The learning purpose of the exercise
2.  The main (idiomatic) strategies for solving the exercise, or the main parts of the exercise.
3.   The discussion points or prompts one might give the student to 'nudge' them in an idiomatic solution direction.
4.   Other tools/techniques/considerations/extensions to the problem the mentor (and student) might think about.


Some things you might want to consider as part of these notes:

- The original intent of this exercise was to mimic **`Enums`**.  It was written before Enums were available in core Python, but the constants in it can serve the same purpose.
-  What other tools in Python could be used to shorten the `if/elif/else` block? (Hint: structural pattern matching)
-  What other _constructs_ could be used _instead_ of `if/elif/else` blocks?  (_the [example solution](https://github.com/exercism/python/blob/main/exercises/practice/yacht/.meta/example.py) uses an interesting one_)

PR #2325, opened 2024-02-15
Add java analyzer comments for ``Need For Speed``

manumafe98:
Related: [#134](https://github.com/exercism/java-analyzer/pull/134)

### DIFF
--- a/analyzer-comments/java/need-for-speed/avoid_loops.md
+++ b/analyzer-comments/java/need-for-speed/avoid_loops.md
@@ -1,0 +1,3 @@
+# avoid loops
+
+Consider converting the `loop` statement and returning a single expression using the `<=` operator while reusing methods already implemented in the class.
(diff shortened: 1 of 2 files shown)


analyzer-comments/java/need-for-speed/avoid_loops.md
```diff
@@ -0,0 +1,3 @@
+# avoid loops
+
+Consider converting the `loop` statement and returning a single expression using the `<=` operator while reusing methods already implemented in the class.
```
sanderploegsma, line 3: I'm not sure that it makes sense to wrap the word "loop" in backticks, as it is not a keyword in Java.

sanderploegsma, line 3: This is quite a long sentence with is a bit hard to read. I'd break it up a bit, something like

```suggestion
Instead of using a loop, consider returning a single expression using the `<=` operator.
You can re-use the other methods already implemented in the class.
```

manumafe98: 👍

sanderploegsma (approved):

SleeplessByte (approved):

PR #2332, opened 2024-03-19
Updating use_substring comment for ``log-levels``

manumafe98:
The goal is to make the comment more clear for students

Related: [#151](https://github.com/exercism/java-analyzer/pull/151)

### DIFF
--- a/analyzer-comments/java/log-levels/use_substring_method.md
+++ b/analyzer-comments/java/log-levels/use_substring_method.md
@@ -1,5 +1,5 @@
 # use substring method
 
-Consider using the [substring][substring-method] method to solve this exercise.
+Consider using the [substring][substring-method] method in `%<inMethod>s` to solve this exercise.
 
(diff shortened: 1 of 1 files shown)


IsaacG (approved):

sanderploegsma:
🤔 I'm not sure this should have been merged without the corresponding changes to the analyzer. Right now the copy contains a parameter but the analyzer doesn't provide the parameter in its output, so my guess is that the copy renders incorrectly for students...

IsaacG:
Should we push a revert PR?

sanderploegsma:
Yes I think that is best. And going forward we should try to avoid making breaking changes to existing comments.

Instead, we can consider creating a new version `comment_name_v2.md` next to the original one, so that we can safely update the analyzer asynchronously.

sanderploegsma:
See https://github.com/exercism/website-copy/pull/2333

PR #2337, opened 2024-03-29
Adding analyzer comments for Salary Calculator

manumafe98:
Related [#157](https://github.com/exercism/java-analyzer/pull/157)

The goal is to add comments for the new salary calculator analyzer
I also deleted an unnecessary black space on a wizards and warriors comment

@sanderploegsma

### DIFF
--- a/analyzer-comments/java/salary-calculator/use_ternary_operators.md
+++ b/analyzer-comments/java/salary-calculator/use_ternary_operators.md
@@ -1,3 +1,3 @@
 # use ternary operators
 
-Consider using `ternary operators` in `%<inMethod>s` to solve this exercise.
+As the goal of this exercise is to learn about ternary operators consider using them in `%<inMethod>s` to solve this exercise.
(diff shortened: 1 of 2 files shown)


manumafe98:
Just in case, do not merge this one until further notice, because the analyzer was merged by mistake and we are reverting it

analyzer-comments/java/salary-calculator/use_ternary_operators.md
```diff
@@ -0,0 +1,3 @@
+# use ternary operators
+
+Consider using `ternary operators` in `%<inMethod>s` to solve this exercise.
```
sanderploegsma, line 3: 1. Maybe the copy can be a little bit more verbose and explain where and why a ternary operator should be used. As for where, presumably this comment is triggered when the student uses an `if`/`else` statement, right? Perhaps include that in the copy. As for why: maybe we can point out that the goal of this exercise is to learn about ternary operators and so solutions should use them instead of something else.
3. I would not put the phrase "ternary operators" in backticks, as that is usually only done when referring to Java code or keywords.

sanderploegsma (changes requested):

analyzer-comments/java/salary-calculator/use_ternary_operators.md
manumafe98: Done!

sanderploegsma (approved):

manumafe98:
It can be merged now! when reviewed! thanks

IsaacG (approved):

PR #2338, opened 2024-04-04
[python/square-root] Add mentoring doc

antoniouaa:
Hello, my attempt at mentoring notes for [square-root](https://exercism.org/tracks/python/exercises/square-root) for python.
I see there's already been a past attempt at introducing this mentoring document , in #2251.

Document includes sections suggesting and explaining 2 different solutions, main solution using Heron's method from the wiki pages linked in the description of the problem. 
I understand that's what the problem statement is pushing people to investigate and implement, instead of using cheap hacks like `** 0.5` and `pow`.

Comments welcome!

### DIFF
--- a/tracks/python/exercises/square-root/mentoring.md
+++ b/tracks/python/exercises/square-root/mentoring.md
@@ -21,10 +21,56 @@ def square_root(number):
     return int(revised) # our answer should be an integer
 ```
 
-### Talking points
+## Talking points
 
 The formula to revise estimates is small, the overall algorithm is pretty simple.
+
+### The algorithm
+
+The idea of the algorithm is to iteratively compute a better, revised estimate for the square root, using the previous estimate.
+
+For example, a single iteration in calculating the square root of 50 looks like this:
+
+```py
+x_0 = 50 / 2                # 25, not a very good first guess
+x_1 = (x_0 + 50 / x_0) / 2  # 13.5, not a very good revised estimate either
+```
+
+The power of Heron's method comes from how quickly it converges to a good approximation of a square root. In just 5 iterations, we get an estimation accurate to 6 decimal places!
+
+```py
+x_0 = 50 / 2                    # 25
+x_1 = (25 + 50/25) / 2          # 13.5
+x_2 = (13.5 + 50/13.5) / 2      # 8.60
+x_3 = (8.60 + 50/8.60) / 2      # 7.21
+x_4 = (7.21 + 50/7.21) / 2      # 7.0724
+x_5 = (7.0724 + 50/7.0724) / 2  # 7.0710679
+...
+x_n = sqrt(50) =~ 7.0710678118654755
+```
 
 People may possibly struggle with figuring out how many iterations to perform to get a decent estimate. Technically, you could get away with hardcoding the iteration count to a number as small as 10 and that would get you good accuracy.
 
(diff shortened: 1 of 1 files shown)


tracks/python/exercises/square-root/mentoring.md
```diff
@@ -0,0 +1,76 @@
+# Mentoring
+
+## Problem and Challenges
+
+The problem asks us to find the square roots of numbers without using builtin functions or operators, like `**`, `pow`, `math.pow`, `math.sqrt` or `sum`.
```
IsaacG, line 5: Why would `sum` not be allowed?

tracks/python/exercises/square-root/mentoring.md
```diff
@@ -0,0 +1,76 @@
...
+The idea of the algorithm is to iteratively compute a better, revised estimate for the square root, using the previous estimate.
+
+For example, a single iteration in calculating the square root of 50 looks like this:
+
+```py
+x_0 = 50 / 2                # 25, not a very good first guess
+x_1 = (x_0 + 50 / x_0) / 2  # 13.5, not a very good revised estimate either
+```
+
+The power of Heron's method comes from how quickly it converges to a good approximation of a square root. In just 5 iterations, we get an estimation accurate to 6 decimal places!
```
IsaacG, line 39: Exercism markdown style is one sentence per line.

tracks/python/exercises/square-root/mentoring.md
antoniouaa: I'm not sure exactly why, but it's mentioned in the exercise instructions as an example builtin to avoid.
> Python offers a wealth of mathematical functions in the form of the [math module](https://docs.python.org/3/library/math.html) and built-ins such as [pow()](https://docs.python.org/3/library/functions.html#pow) and [sum()](https://docs.python.org/3/library/functions.html#sum). However, we'd like you to consider the challenge of solving this exercise without those built-ins or modules.

I can see at least 2 commmunity solutions where this function is used, inappropriately I might say:
1. https://exercism.org/tracks/python/exercises/square-root/solutions/PetaGb
2. https://exercism.org/tracks/python/exercises/square-root/solutions/soroushmz

antoniouaa: Mentoring notes can't really force a student to change their solution and I don't think the first couple lines suggest it absolutely _shouldn't_ be done using `**`. Students could even hardcode values to fool the tests.

I imagine that people who asked for mentorship did so for 2 reasons:
1. Their solution is too simple and are wondering if there's more to it,
2. They tried implementing one of the methods in the wiki articles and want to improve on that implementation.

The spirit of the exercise is not to use a builtin operator, but to investigate other methods of doing the calculation without it.

Heron's method happens to be the simplest, its the first method linked in the [wiki page](https://en.wikipedia.org/wiki/Methods_of_computing_square_roots#Heron's_method), there's no intermediate step between `**` or `math.pow` and that.

It's Heron's method that's the intermediary step, Newton's method and binary search (included as an alternative solution) follow the same path of shrinking the lookup space.

I could add the simpler solution of:
```py
def square_root(number):
    for num in range(number+1):
        if num * num == number:
            return num
```
But this isn't really related to Heron's method, IMO they're distinct methods.

As for a `sum()` solution, I can't think of an appropriate one. The two `sum()` community solutions linked above sum just one number from the range. The snippet linked just above is arguably an improvement on it.

PR #2346, opened 2024-06-07
Update mentoring.md

thomasantony12:
changed a grammatical error


IsaacG:
The grammar is correct as-is.

PR #2349, opened 2024-06-08
Update mentoring.md

thomasantony12:
Removed a word and replaced it with the correct word


IsaacG:
The original text is correct. Please double check changes before creating PRs.

PR #2363, opened 2024-12-18
Create mentoring.md

theomelo:


tracks/ruby/mentoring.md
IsaacG, line 1: This is rather sparse. You may want to look at other mentoring docs and follow the style of an existing doc.

PR #2372, opened 2025-04-18
Fix typo

phuc1nguyen:
Hi, I spotted a typo while reading the Typescript mentoring guide.
- Should be `default value` instead of `default type`
- Use better wording IMHO

### DIFF
--- a/tracks/typescript/exercises/two-fer/mentoring.md
+++ b/tracks/typescript/exercises/two-fer/mentoring.md
@@ -151,9 +151,9 @@ exercises.
   accidental mistakes. If you expect the function to return a string, mark it as
   such.
 - Even if you have a default value applied to your parameter, mark its type even
-  though it currently can be inferred. This lets you later remove the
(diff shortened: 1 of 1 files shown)


IsaacG:
Given that these notes are only intended for mentors, we generally don't encourage changing the wording from whomever wrote these notes. If something is incorrect (eg a spelling error or technical error), it might be worth fixing. But we don't especially want to spend time improving the wording (and reviewing) on mentor notes.

phuc1nguyen:
Hi, thank you for the response. I've updated to fix the typo only.

IsaacG:
+cc @exercism/typescript maintainers

kotp (approved):

kotp:
The reason for the approval is that in untyped languages, this wording could be confusing.  In typed languages, it also may be the wrong thing to communicate.

I also noticed that there is no discussion linked to on the forums.

SleeplessByte (approved):
This allows you to later remove the `default value` is indeed correct.

PR #2373, opened 2025-04-25
Create mentoring.md for collatz-conjecture for cpp

raphaelracine:
Proposition of guidance for Collatz Conjecture exercice.

### DIFF
--- a/tracks/cpp/exercises/collatz-conjecture/mentoring.md
+++ b/tracks/cpp/exercises/collatz-conjecture/mentoring.md
@@ -1,0 +1,34 @@
+Hello,
+
+I sugget to use a ```while``` loop and a ternary expression (more concise than an if-else) for this exercice.
+
+Example of solution :
+
+```cpp
+
+#include "collatz_conjecture.h"
+#include <stdexcept>
+
+namespace collatz_conjecture {
+
+    int steps(int number) {
+
+        if (number <= 0) {
+            throw std::domain_error("Only positive numbers are allowed!");
+        }
+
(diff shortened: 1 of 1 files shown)


IsaacG:
You may want to review other mentor notes. These notes should follow a similar format. Mentor notes are not a place to share a solution. It's a place to discuss points that a mentor may want to cover in a discussion.

raphaelracine:
Hello @IsaacG,

Ah ok, sorry I didn't understand well. Thanks for feedback.

raphaelracine:
Hello @IsaacG , I was actually offering this as guidance for mentors, not as notes.

I've seen quite a few exercises where the Guidance tab during mentoring presents a solution that allows the mentor to better guide a student.

I may have the wrong directory for the file :-)

IsaacG:
Including a solution as part of the mentoring notes is fine. But the solution should be incidental and not the primary guidance. Assume mentors know how to solve the exercise already and have their own solution. Either they used a while loop and a ternary or they didn't. If they did, case your notes aren't adding anything. If they used something else, is your guidance to not solve it another way? Why? What are the pros and cons? How would they use that to mentor students?

You may want to review other mentoring notes and see what sort of content is there beyond just a sample solution. Every exercise already has a sample solution ([see C++ Collatz Conjecture example](https://github.com/exercism/cpp/blob/main/exercises/practice/collatz-conjecture/.meta/example.cpp)) as well as all the community solutions. The mentor notes should be something more than "here's yet another solution to look at".

PR #2378, opened 2025-06-03
Add mentoring notes for python/guidos-gorgeous-lasagna

IsaacG:


BethanyG (dismissed):
Nice work!  Thanks for writing these up.  

One niggle:  There is a 5th "task", which is adding [docstrings to the functions](https://github.com/exercism/python/blob/main/exercises/concept/guidos-gorgeous-lasagna/.docs/instructions.md#5-update-the-recipe-with-notes), which we might want to add a sentence on, with ref to PEP257 (_and if you wanted to, sample feedback_).

But its a niggle, so I am marking this approved, and you can merge or add the comments as you see fit.  😄

BethanyG (approved):
Awesome!  Thanks for adding docstrings in.  Good to merge from my side.  😄

PR #2405, opened 2026-06-14
Improve: fix nested code block and expand Common Suggestions in Rust bob mentoring

PaezCande:
- Fixed malformed nested code block in the match suggestion
- Expanded Common Suggestions with three new idiomatic Rust tips:
  - Prefer zero-allocation char checks over .to_uppercase()
  - Prefer .chars().any() over .contains() for character property checks
  - Prefer match on tuple over nested if/else

### DIFF
--- a/tracks/rust/exercises/bob/mentoring.md
+++ b/tracks/rust/exercises/bob/mentoring.md
@@ -95,13 +95,14 @@ See about finding a way to check if it's alphabetic without using regex.
 ```
 
 If they don't use a `match` to compare the conditions:
-````md
+
 We can `match` on just two conditions: `is_question` and `is_yelling`. Like this:
-```rust
+
+````rust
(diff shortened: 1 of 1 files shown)


tracks/rust/exercises/bob/mentoring.md
```diff
@@ -95,13 +95,14 @@ See about finding a way to check if it's alphabetic without using regex.
 ```
 
 If they don't use a `match` to compare the conditions:
-````md
```
IsaacG, line 98: This is intentional and formats correctly. This is not malformed.

PaezCande: Thanks for the feedback and for fixing the code block! The three new Common Suggestions are still included. Let me know if there's anything else to adjust.

PR #2407, opened 2026-07-18
Improve Elixir function annotation order comment

### LINKED ISSUE #470: bug: elixier-analyzer is telling me to put @doc before @spec but I don't use @doc
I've just completed the Lucas Numbers exercise, but I'm getting automatic feedback with the following:

> Developers can choose the order of the `@doc` and `@spec` modules attributes, but the Elixir community convention is to use `@doc` first and `@spec` next to the function.

However, I'm not even using `@doc` in my answer at all!

```ex
defmodule LucasNumbers do
  @moduledoc """
  Lucas numbers are an infinite sequence of numbers which build progressively
  which hold a strong correlation to the golden ratio (φ or ϕ)

  E.g.: 2, 1, 3, 4, 7, 11, 18, 29, ...
  """
  def generate(count) when not is_integer(count) or count < 1 do
    raise ArgumentError, "count must be specified as an integer >= 1"
  end

  def generate(1), do: [2]
  def generate(2), do: [2, 1]

  @spec generate(pos_integer()) :: [pos_integer()]
  def generate(count) do
    Stream.iterate({2, 1}, fn {x, y} -> {y, x + y} end)
    |> Enum.take(count)
    |> Enum.map(fn {x, _} -> x end)
  end
end
```

I also tried removing the `@moduledoc` to see if that would resolve it, but I still get the feedback. 🤔

angelikatyborska:
Resolves https://github.com/exercism/elixir-analyzer/issues/470

### DIFF
--- a/analyzer-comments/elixir/solution/function_annotation_order.md
+++ b/analyzer-comments/elixir/solution/function_annotation_order.md
@@ -1,12 +1,14 @@
 # Function Annotation Order
 
-Developers can choose the order of the `@doc` and `@spec` modules attributes, but the Elixir community convention is to use `@doc` first and `@spec` next to the function.
+Conventionally, `@doc` function annotations should come before `@spec` function annotations, and `@spec` function annotations should come directly before the function. If the function has multiple clauses, function annotations should go before the first clause.
 
 Example:
+
 ```elixir
 @doc """
 This function counts the number elements in a list
 """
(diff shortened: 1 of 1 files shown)


SethFalco (approved):

IsaacG (approved):

kotp (approved):

jiegillet (approved):
Nice!

PR #2408, opened 2026-09-10
fix: mention String.formatted() in prefer_string_concatenation comment

Thakurji890:
The feedback message only referenced String.format (static method) but the analyzer also flags String.formatted() (instance method, Java 15+). Update the message to mention both methods to avoid confusion.


kotp (commented):
This used to be an order of magnitude slower, but not sure if "extremely" slower is now language that is warranted.

Since Java 17's improvement with `String.format` being 3 times faster than it was, it is likely a reasonable solution, until you are really identifying this as the bottleneck of the program.

Should we remove the `extremely` or at least quantify what that means?

Thakurji890:
Done, let me know if any changes need to be done before merging.

IsaacG:
+cc @exercism/java

kahgoh (approved):
Looks good to me.

kotp (approved):

